PR #1196, opened 2015-03-07
add option to show current project in menubar (osx)

olvidalo:


tanel:
Very nice! Can you please also run "make fmt" to format the sources and "make lint" to check the code with cpplint?

olvidalo:
there you go! everything fine now as far as i can see

tanel:
Thank you!

PR #1272, opened 2015-04-05
Add support for tray icon (linux)

buma:
It should work everywhere Xembed is supported (KDE, gnome2). There
seems to be some problems in Gnome 3 and Unity support in [Qt 5.3](http://stackoverflow.com/questions/24373967/qt-5-3-qsystemtrayicon-not-working-correctlylinux).
It probably also works in MacOS. Icon is by default grayscale. It gets
normal color when time is tracked. Clicking on it opens a menu where
you can stop/continue tracking and log out.

It is a long time since I last programmed in C++ and Qt and I hope that everything is correct.
Tests, fmt and linter passes, uitest throws segfault. But this also happens with last master commit. I haven't figured out why yet. Manual testing of icon and app seems to work OK.

Currently only 16x16 icon is grayscale.

QIcon seems to creaty grayscale pixmap by itself when disabled mode is requested but I haven't figured out how to use this as tray icon.

I think that menu on tray icon has same functionality like it was before in previous versions.


buma:
@IndrekV: was there a problem with this commit and how can I fix it?

IndrekV:
I reverted the commit because it wasn't working properly on ubuntu. Looking around in the internet I found that the tray icon implementation in linux doesn't have a good solution in qt.

Here's a screenshot what I saw when testing:

linux_tray

buma:
Could be a compile time or configuration option to enable it? Since tray icon isn't supported only in Gnome 3 and unity because they don't support Xembed. All other window managers and KDE work.

IndrekV:
Ok, I created a separate issue for this. #1303

Issue #1881, opened 2016-04-20
Getting error running make lint and make test on fresh install


eatskolnikov:
I'm working from my mac following the instructions in the readme and everything is ok. If i run `make` or `make run` it works perfectly, but when I run `make lint` and `make test` I get the following results:

![screen shot 2016-04-20 at 5 21 34 pm](https://cloud.githubusercontent.com/assets/1693000/14690884/5a259fe4-071c-11e6-9e95-e8b9af91f179.png)

![screen shot 2016-04-20 at 5 19 56 pm](https://cloud.githubusercontent.com/assets/1693000/14690846/30363cca-071c-11e6-9c8c-3ffaa397b499.png)

IndrekV:
Error on `lint` means that lint has failed and we should manually fix the formatting of the files. It actually shows what exactly is not ok and in what line of what file.

About the `make test` error. It seems we have forgotten to update tests after some updates.

So it's not you it's the code.

eatskolnikov:
I can help to fix the formatting and even try to fix a few tests then. That way I can get familiar with the codebase.

IndrekV:
I've fixed the tests. You can try to fix the formatting errors.

PR #1883, opened 2016-04-22
linting

eatskolnikov:
As mentioned on issue #1881

screen shot 2016-04-21 at 7 34 52 pm

I liked how the formatter is made with go  👍


IndrekV:
Thank you for your contribution and yes we are fans of Go here in Toggl. A lot of our backend is running on Go. Here is an article from long time ago when we decided to give Go a chance. https://blog.toggl.com/2012/09/moving-to-go/

PR #1951, opened 2016-07-21
Pomodoro break

unmultimedio:
Guys, this is my **very first** input into open-source projects, and I'm very excited I finally managed to develop a new tiny functionality in one of my favourite products!

The Pomodoro Timer was incomplete, and breaks were measured manually (or mentally)

Added the "Pomodoro Break" Option, with a very similar functionality that the original _"Pomodoro"_ has.

If activated both options (Pomodoro and Pomodoro Break), and set the due times for each task, e.g. 25min for Pomodoros, and 5min for breaks (as methodology suggest), once initiated a task, and meet the 25 minutes, automatically another task named "Pomodoro Break" starts, and finishes (and more importantly, alerts) after 5 minutes, so we get a hint to get back to work.

Currently only developed the option for Library and OSX Client.

toggl_desktop_and_preferences_and_toggl-desktop_ _make___users_julian_development_toggl-desktop_ _toggldesktop_ _make_run_ _208x58


unmultimedio:
Any update on this? 😺

aronax1:
I second this option to be released also on windows. Thank you for your work!

IndrekV:
I will look into this next week.

IndrekV:
Thank you for your contribution. You did a good job. I added the functionality and settings to other platforms and fixed some issues. The main issue I found was that pressing `Continue` in Pomodoro Break should start the entry that was running before the break.

Very good job @unmultimedio . I'm impressed you figured these things out so fast.

I'll try to push the update to beta channel at the end of this week or if something comes up in the beginning of next.

Thank you come again.

nschurmann:
Dude i love this!!, i want to add something to this. Basically that the pomodoro timer won't stop when the time is reached, because you still may be away from the computer and i want to track that! (sometimes i take longer breaks, i want to measure that too). The downside is that i don't know c++ :P, will try to figure it out.

Issue #2273, opened 2017-09-28
Don't send 'Content-Encoding' and 'Content-Length' on GET


shedokan:
Currently all requests without a form have the `Content-Encoding` and `Content-Length` headers attached, including GET.
See https://github.com/toggl/toggldesktop/blob/master/src/https_client.cc#L398

I believe this has caused the https://github.com/toggl/toggldesktop/issues/2272 issue, and just against HTTP specs.

PR #2275, opened 2017-09-28
Do not send content-length on GET request

refiito:
should fix https://github.com/toggl/toggldesktop/issues/2273


IndrekV:
Seems ok with live api. With staging api I'm still seeing `data not valid error`. Is the setup somehow different in staging.

screen shot 2017-09-29 at 12 52 20

PR #2278, opened 2017-10-04
Add Fedora dep readline-devel

jayvdb:
`libreadline-dev` is a Debian based distro package name.
Move it into the appropriate section, and add the Fedora
equivalent.


IndrekV:
Thank you for your contribution, Sir.

Issue #2279, opened 2017-10-04
Linux build should use native libraries where possible


jayvdb:
Especially for openssl, using custom libraries is a bad idea, and also makes compilation take a very long time unnecessarily, and will result in often security problems due to an old version of openssl.

At the moment, [this](https://github.com/toggl/toggldesktop/commit/601fd720b07cf40cd116e) is the latest change to the bundled openssl lib, and it is v1.0.2e , which is horrifically old and the number of [openssl security vunerabilities](https://www.openssl.org/news/vulnerabilities.html) since then is very large.  Some will be irrelevant for a variety of reasons, but I'll bet that some are a real issue.

I've confirmed that the .deb and .tgz being distributed via the web UI includes its own v1.0.2e.  Also confirmed the macOS installer doesnt include its own ssl library.  Didnt try Windows ;-)

On Fedora & Debian, poco is available as packages.  wrt published [poco security vunerabilities](https://www.cvedetails.com/vulnerability-list/vendor_id-13266/product_id-27528/Pocoproject-Poco-C-Libraries.html), it seems there are none in the v1.6.1 found in `third_party/poco`.  However Fedora is using 1.7.9, and Debian stable is using 1.7.6 .

The following patch is a quick hack to disable the `third_party/poco` & `third_party/openssl` and use the packaged libraries instead.

```diff
diff --git a/Makefile b/Makefile
index 8144a4ded..589375b5c 100644
--- a/Makefile
+++ b/Makefile
@@ -5,8 +5,8 @@ uname=$(shell uname)
 architecture=$(shell uname -m)
 timestamp=$(shell date "+%Y-%m-%d-%H-%M-%S")
 
-pocodir=third_party/poco
-openssldir=third_party/openssl
+pocodir=/third_party/poco
+openssldir=/third_party/openssl
 jsoncppdir=third_party/jsoncpp/dist
 
 GTEST_ROOT=third_party/googletest-read-only
@@ -77,7 +77,6 @@ endif
 
 ifeq ($(osname), linux)
 cflags=-g -Wall -Wextra -Wno-deprecated -Wno-unused-parameter -static \
-	-I$(openssldir)/include \
 	-I$(GTEST_ROOT)/include \
 	-I$(GTEST_ROOT) \
 	-I$(pocodir)/Foundation/include \
@@ -244,18 +243,7 @@ endif
 
 ifeq ($(osname), linux)
 lib:
[...]

PR #2283, opened 2017-10-05
Build with system openssl and poco

jayvdb:
Allow openssldir and pocodir to be undefined,
to build using the system provided libraries.

Related to https://github.com/toggl/toggldesktop/issues/2279


IndrekV:
Thank you for your contribution. We should definitely also update the third party libraries.

PR #2376, opened 2018-01-31
Reword visibility preference for clarity

Leland:
Preference settings are typically verbs ("Do this") but the always-shown visibility setting wasn't. Currently, it reads `On top of other windows` which is rather confusing for the end user.

This minor change moves it to `Keep Toggl on top of other windows` which should hopefully be a bit more clear.


IndrekV:
I agree, this is much better. Thank you for your contribution, Sir!

PR #2418, opened 2018-03-27
Link to Github repo on MacOS build

Jackymancs4:
Hello,
just a little modification to make the UI similar to the Windows one.
Copied from ```src/ui/osx/TogglDesktop/test2/et.lproj/Credits.rtf```.


IndrekV:
The text is a hyperlink but not very readable
screen shot 2018-03-27 at 14 24 16

Jackymancs4:
Hello, sorry for the late response.
I don't think the text is unreadable (at least on my system), but I definetely agree that it looks pretty bad.

I tried but I am not able to fix this. I'm pretty sure that it's directly related to how the SO natively renders rtf files, so there is nothing I can do about it.
Sorry, feel free to close the PR.

Issue #1280, opened 2015-04-07
Window changes to default height after closing edit view


IndrekV:
Window re-snaps to full height after editing a task Steps to repeat: 
(1) shrink window height/width
(2) click on a task to get to detailed view 
(3) click done

window snaps back to default height

PR #2560, opened 2018-09-04
Prevent TimeEntryEdit view from changing size when "Add new project" is clicked

### LINKED ISSUE #2382: Edit view changes size when opening "Add project" form
Edit view is made much wider. May be because of some long text client,tag etc in one of the dropdowns.

May be connected to #1280

MartinBriza:
Fixes #2382


IndrekV:
closes #1280, closes #2382

PR #2562, opened 2018-09-04
Move the cursor to the end of Time string in Edit View (windows)

### LINKED ISSUE #2097: Cursor Placement Issue On Windows Desktop
"When I need to correct the start time of an entry most of the time only the last digit is affected. Thus, I would like to put the cursor to the right of that last digit. - No text entry field has the cursor showing! - When I click into the blank space at the right of the text field that shows the start time, the cursor appears at the far left. - If I click again, the cursor stays where it is. - I have to move the mouse to within two pixels to the right of the last digit of the time stamp before clicking will move the cursor to the right of the last digit." User included a gif viewable here 410030.

MartinBriza:
Fixes #2097

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/controls/ExtendedTextBox.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/controls/ExtendedTextBox.cs
@@ -9,9 +9,12 @@ namespace TogglDesktop
 
         public bool SelectAllOnKeyboardFocus { get; set; }
 
+        public bool MoveToEndOnFocus { get; set; }
+
         public ExtendedTextBox()
         {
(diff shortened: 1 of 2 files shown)


IndrekV:
Sadly it does not fix this issue. It works only on the very first time you focus on the textfield. After that the cursor is not set to the end if you click on the right side of the field. It works when you click a few pixels right from the text. 

It works properly with the description field in timer (the "What are you doing?" field) I think the issue might be connected to the size of the textfield or something like this.

IndrekV:
It should work so that if you click anywhere inside the textfield right from the text it puts the cursor to the end of the inserted text. It is a default logic I think. Why it doesn't work so in the edit form fields is strange.

PR #2564, opened 2018-09-05
Clear ExtendedTextBox history when set programatically (windows)

### LINKED ISSUE #2409: Resolve Ctrl+Z changing entries on Windows app
As described in /case/657584:

Using the windows desktop client, on Windows 10:

This seems to affect the text / time entry fields, I have noticed it on both the description, the start time, and end time.

When one of these fields is focused, pressing ctrl-z resets the value to some previously entered value. Even if not for the current entry.

To reproduce:
* Start an entry "test" wait 10 seconds and stop.
* Click the entry to edit its details.
* Fix the name to be "my test" and the end time so that the duration becomes 15 seconds. For example, "6:00:10 PM" becomes "6:00:15 PM"
* Create an entry "test2" wait 10 seconds and stop.
* Click the entry to edit its details.
* Focus the name and press ctrl-z
* Focus the end time and press ctrl-z

### Actual behaviour
The name "test2" will become "my test" and the end time will become "6:00:15 PM"

### Expected behaviour
The name and time should remain unchanged, unless there is some history for this item specifically. The history should be item-centric not control centric.

MartinBriza:
Fixes #2409

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/views/EditView.xaml.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/views/EditView.xaml.cs
@@ -180,7 +180,10 @@ namespace TogglDesktop
         {
             if (evenIfFocused || !textBox.IsKeyboardFocused)
             {
+                textBox.Clear();
                 textBox.SetText(project, task);
+                textBox.IsUndoEnabled = false;
+                textBox.IsUndoEnabled = true;
             }
         }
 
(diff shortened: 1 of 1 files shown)


MartinBriza:
I'm not sure if this is the expected behavior though, please check.

IndrekV:
Sadly. It's not working as expected.

*Expected:*
The undo resets only when:
- the entry displayed is changed
- Edit view is closed

*Current:*
if I edit description and switch to project field and then switch back to description field the description field undo list has already been erased.

PR #2608, opened 2018-11-06
Fix for ui not showing up in manjaro gnome distro

nabuhodonozo:
There has been a bug, that none of all 3 aur packages working on my os. I have been struggling with it for a month, until finally solving the issue myself. I'd like to share solution with other ppl, I hope it will help someone.

### DIFF
--- a/src/ui/linux/TogglDesktop.sh
+++ b/src/ui/linux/TogglDesktop.sh
@@ -16,7 +16,8 @@ export QTWEBENGINEPROCESS_PATH
 # Xubuntu, i3 and Cinnamon tray icon fix
 XDG=$XDG_CURRENT_DESKTOP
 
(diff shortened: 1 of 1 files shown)


nabuhodonozo:
It kinda might cause issues on other environments or distros. But is the only solution for mine. I just wanted to let you know that the issue exist, and how to dirty fix it. Maby you come up with better solution, when you know what is causing the problem

MartinBriza:
Hmm, I don't think abusing `QT_QPA_PLATFORMTHEME` is a good idea, at all. I'm going to look into this next week on Manjaro with GNOME to see what is the real problem you're trying to solve with this patch and I'm planning to remove the `if` clause you are modifying in this PR from the script altogether.

MartinBriza:
I have just downloaded and tried and tested our official Linux release (the tarball one) and everything works fine. Also an experimental build (tag 330) works fine for me. I see you've mentioned AUR - that's not officially supported by us and if our tarball works, there's most likely something wrong with the AUR package - please consult this issue with the maintainter as there's nothing I can do about this from my side, sorry.

PR #2566, opened 2018-09-06
New custom autocomplete for linux

IndrekV:
Main features are:
- Different type of items in the list
- Colors in the list for project items
- Filtering
- Selecting with keyboard

PR #2626, opened 2018-11-12
Fix autocomplete behavior on Linux

MartinBriza:
Please check the code. From my point of view, it's better to keep the consistency by filling the list each time it is requested.

The code is not the greatest at this moment, from my point of view. I hence propose rewriting autocomplete to QItemView instead of QItemWidget (along with the changes to the underlying data model), especially for the ability to utilize QAbstractItemModel and QSortFilterProxyModel. 

It is easier to control how the items are filtered with that approach and also how many items get to be displayed.


IndrekV:
Connected to abandoned autocomoplete implementation #2566

Issue #2369, opened 2018-01-26
Strip spaces from description when only there are only spaces


IndrekV:
If time entry description is only spaces all spaces should be stripped so that `(no description)` is shown in list.

PR #2634, opened 2018-11-19
Trim all leading & trailing space in Description

NghiaTranUIT:
## ❓ What's this?
Remove leading and trailing whitespace in Description 

## 💾 How is it done?
- Override the setter of Description of `TimeEntry` in library and trim all leading & trailing whitespace.
- Trim whitespace in macOS UI.

## 🤯 Changelogs 
- [x] Introduce `trim_whitespace` in `toggl_api_private.h`
- [x] Override the setter of Description of `TimeEntry` in library and trim all leading & trailing whitespace.
- [x] Override the setter/getter of Description property in `TimeEntryViewItem` and `AutocompleteItem`
## 👫 Relationships
Closes toggl/toggldesktop#2369

## 🔎 Review hints
- Try to enter "          " in Description Label in `TimeEntryEditViewController` and `TimerEditViewController` => If the text is become "" and show `(no description)` => OK 💯

### DIFF
--- a/src/time_entry.cc
+++ b/src/time_entry.cc
@@ -23,6 +23,7 @@
 #include "Poco/Logger.h"
 #include "Poco/NumberParser.h"
 #include "Poco/Timestamp.h"
+#include "toggl_api_private.h"
 
 namespace toggl {
 
@@ -197,8 +198,9 @@ void TimeEntry::SetStop(const Poco::UInt64 value) {
 }
 
 void TimeEntry::SetDescription(const std::string value) {
-    if (description_ != value) {
-        description_ = value;
+    const std::string trimValue = trim_whitespace(value);
+    if (description_ != trimValue) {
+        description_ = trimValue;
         SetDirty();
     }
 }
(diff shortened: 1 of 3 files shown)


NghiaTranUIT:
Agree that, however the current macUI logic isn't completely corresponding with the value on `TimeEntry`. 

If we just fix in the library, the bug is still remaining in few **edge scenarios**, because the TextFields are still maintain an old state. Most of the time it works, otherwise it don't.

IndrekV:
If you can describe edge cases you referred to then let me know so I can test. I didn't figure out how this is possible.

The app logic is:
- Changes go to the library (like time entry description change)
- Library saves the changes into the model and updates local database
- This triggers UI update and new data is sent to UI (at this point the trimmed time entry description will go to UI)

It shouldn't be possible that the `setDescription` is called on either `TimeEntryView` or `AutocompleteView` with data that isn't coming from the library and therefore is not already trimmed.

NghiaTranUIT:
I'm unable to replicate the edge cases I mention before anymore.

I will roll-back the second commit 👍

NghiaTranUIT:
Done - Pls continue your review 🥇

Issue #2063, opened 2016-12-05
No cursor shown in email field on login screen on macOS


fonorobert:
As seen on screenshot, on macOS when signing in there is no visible cursor in the email field.
<img width="454" alt="screen shot 2016-12-05 at 12 47 48" src="https://cloud.githubusercontent.com/assets/6963710/20883963/0f618a1e-bae9-11e6-8c2c-7650cfdbd3a0.png">

Signin works fine however, this is just an aesthetic issue.

PR #2635, opened 2018-11-20
Reset Cursor's color in Email/Pass TextField

NghiaTranUIT:
## ❓ What's this?
Fix the bug when the Cursor's color is vanished completely.

The problem is that we're trying to override the Cursor's color to WhiteColor in `NSCustomTimerComboBox`, `NSTextFieldClickable` and `AutoCompleteInput`.
Ex: https://github.com/toggl/toggldesktop/blob/b2f297423762b9df2e66a67ab67e7f4137bb2e88/src/ui/osx/TogglDesktop/test2/NSTextFieldClickable.m#L34

So, it results in the remaining of `White Color` of the Email/Password's TextField Cursor.

## 💾 How is it done?
- Reset the Cursor's color with `[NSTextField textColor]` whenever the `LoginViewController` did appear.
=> By selecting the `textColor`. It automatically adopts with White/Dark theme in OS Mojave. As a result, the cursor is always visible.

## 🤯 Changelogs 
- Introduce the extension of NSTextField, and implement `[NSTextField resetCursorColor]`
- Reset cursor color of Email/Pass TextField when the controller appears.

## 👫 Relationships
Closes toggl/toggldesktop#2063

## 🔎 Review hints
- Login and open the Entry List, then selecting the Description TextField, then logging out and check the color of cursor of Email field.
=> If it visible => Correct behavior.
- Repeat on White / Dark theme.

### DIFF
--- a/src/ui/osx/TogglDesktop/test2/NSTextField+Ext.m
+++ b/src/ui/osx/TogglDesktop/test2/NSTextField+Ext.m
@@ -1,0 +1,22 @@
+//
+//  NSTextField+Ext.m
+//  TogglDesktop
+//
+//  Created by Nghia Tran on 11/20/18.
+//  Copyright © 2018 Alari. All rights reserved.
+//
+
+#import "NSTextField+Ext.h"
+
+@implementation NSTextField(Extension)
+
+-(void)resetCursorColor
+{
+    NSTextView *textField = (NSTextView *)[self currentEditor];
+    if ([textField respondsToSelector:@selector(setInsertionPointColor:)])
+    {
+        [textField setInsertionPointColor:[self textColor]];
+    }
+}
+
+@end
--- a/src/ui/osx/TogglDesktop/test2/NSTextField+Ext.m
+++ b/src/ui/osx/TogglDesktop/test2/NSTextField+Ext.m
@@ -12,10 +12,10 @@
 
 -(void)resetCursorColor
 {
-    NSTextView *textField = (NSTextView *)[self currentEditor];
-    if ([textField respondsToSelector:@selector(setInsertionPointColor:)])
+    NSTextView *editor = (NSTextView *)[self currentEditor];
+    if ([editor respondsToSelector:@selector(setInsertionPointColor:)])
     {
-        [textField setInsertionPointColor:[self textColor]];
+        [editor setInsertionPointColor:[self textColor]];
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/ui/osx/TogglDesktop/test2/NSTextField+Ext.m
```diff
@@ -0,0 +1,22 @@
...
+//  Copyright © 2018 Alari. All rights reserved.
+//
+
+#import "NSTextField+Ext.h"
+
+@implementation NSTextField(Extension)
+
+-(void)resetCursorColor
+{
+    NSTextView *textField = (NSTextView *)[self currentEditor];
```
IndrekV: I'd use a better name for this variable. `editor` or `textView` or something like that. `textField` seems out of place here.

src/ui/osx/TogglDesktop/test2/NSTextField+Ext.m
```diff
@@ -0,0 +1,22 @@
...
+    if ([textField respondsToSelector:@selector(setInsertionPointColor:)])
+    {
+        [textField setInsertionPointColor:[self textColor]];
```
IndrekV: We can't use system defined text color here as we do not currently support macOs Dark mode and it will we white on white if the user has Dark mode enabled.

You can use hex color with `[ConvertHexColor hexCodeToNSColor:color]];` method.

src/ui/osx/TogglDesktop/test2/NSTextField+Ext.m
NghiaTranUIT: Yup, it's my mistake 👍

src/ui/osx/TogglDesktop/test2/NSTextField+Ext.m
NghiaTranUIT: If the app runs on OS Mojave, and using System Text Color, it means that the app is automatically adopted the Dark Mode. (See the screenshots). The `textColor` will change depend on the system Mode.
screen shot 2018-11-20 at 16 24 20
screen shot 2018-11-20 at 16 24 24

So, by setting the [self textColor], it guarantees that the Cursor Color is always same with Text Color, means it's visible all time regardless of white/dark mode.

I tested it in both mode, and it worked as our exceptions. Or I'm missing something? 🤔

NghiaTranUIT: Although, if the app is built on Xcode 9.3, and run on OS Mojave, then the value of `[self textColor]` guarantees that the color of cursor color is same with the text color in Email. It means that the cursor is always visible.

As a result, it still corrects IMO

PR #2636, opened 2018-11-20
Enable/Disable Command+Q depend on the state of Setting's dock_icon

### LINKED ISSUE #1776: Menubar apps should not quit
The normal behaviour for menubar apps is to not quit until explicitly chosen from the right click menu. I keep accidentaly quitting toggl with command-Q.

NghiaTranUIT:
## ❓ What's this?
According to previous [discussion](https://github.com/toggl/toggldesktop/issues/1776#issuecomment-439827698) (#1776), we expect that:
1. `Cmd + Q` to quit the app if the Dock is visible 
2. Disable `Cmd + Q` if the Dock is hidden.

This PR is the elegant solution to tackle this problem 💁‍♂️

 ## 💾 How is it done?
- We explicitly enable / disable the `Command + Q` shortcut, which relies on the state of `Setting's dock_icon`.
- If it's ON, we set a shortcut (Cmd+Q) for `QuitMenuItem`
- Otherwise, we disable it. Consequently, the user must quit the app by selecting on Status Menu.

 ## 🤯 Changelogs
- [x] Introduce `AppDelegate(Shortcut)` extension to handle shortcut appropriately. In the future, if we would like to add more shortcut, it's potential place to do it.
- [x] Set OUTLET of `QuitItemMenu` to `AppDelegate`.
- [x] Enable / Disable the shortcut whenever the Setting is changed from `kDisplaySettings` notification.

 ## 👫 Relationships
Closes #1776 

 ## 🔎 Review hints
- Enable "Show Dock Icon" in Preferences, then trying to quit (Cmd+Q) => App quit => Correct 💯
- Disable "Show Dock Icon" => Unable to quit (Cmd+Q) the app until selecting on Status Menu => Correct 💯


IndrekV:
We made a change in direction with this https://github.com/toggl/toggldesktop/issues/1776#issuecomment-446161393

PR #2645, opened 2018-11-22
Collect analytics about app window sizes

### LINKED ISSUE #2644: Collect analytics about app window sizes
We want to know the most common app window size's so we should collect this information.

Data should be collected on app start.

The logic behind it:

- Create new collecting function into toggl_api.cc
- Inside that function the analytics function from Context is called.
   See this `toggl_set_wake` function in `toggl_api.cc` for reference
- Create function in analytics class and call it from Context class
  See this function for reference `Analytics::TrackOs`

NghiaTranUIT:
## ❓ What's this?
A feature to track the size of window when app starts.

 ## 💾 How is it done?
- Implement the tracking logic in `Analytics`, `Context` and `Rectangle`
- Tracking Window Size when the `MainWindowController` didLoad.

 ## 🤯 Changelogs 
- [x] Implement the `Analytics::TrackWindowSize` func with category = `stats` and action = `winsize-{w}x{h}`
- [x] Introduce Rectangle to handle `Rectangle::str()` appropriately.
- [x] Implement `Context::TrackWindowSize(` and execute if it's production env
- [x] Introduce `TrackingService` in macOS app
- [x] [macOS] Track the size of window when `MainWindowController` didLoad.
- [x] [Window] Track the size of window.

 ## 👫 Relationships
Closes #2644

 ## 🔎 Review hints
- Switch app to Production
- Build and run Toggl MacOS, then checking the record in the Google Analytic page when app opens
- or put a breakpoint in `Analytics::TrackWindowSize` to see if it executes when app opens

### DIFF
--- a/src/analytics.cc
+++ b/src/analytics.cc
@@ -47,9 +47,10 @@ void Analytics::TrackOs(const std::string client_id,
 }
 
 void Analytics::TrackWindowSize(const std::string client_id,
+                                const std::string os,
                                 const toggl::Rectangle rect) {
     std::stringstream ss;
-    ss << "winsize-"<< rect.str();
+    ss << os << "/mainsize-" << rect.str();
 
     Track(client_id, "stats", ss.str());
 }
--- a/src/analytics.cc
+++ b/src/analytics.cc
@@ -46,6 +46,14 @@ void Analytics::TrackOs(const std::string client_id,
     Track(client_id, "os", ss.str());
 }
 
+void Analytics::TrackWindowSize(const std::string client_id,
+                                const toggl::Rectangle rect) {
+    std::stringstream ss;
+    ss << "winsize-"<< rect.str();
+
+    Track(client_id, "stats", ss.str());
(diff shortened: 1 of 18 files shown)


src/analytics.cc
```diff
@@ -46,6 +46,14 @@ void Analytics::TrackOs(const std::string client_id,
     Track(client_id, "os", ss.str());
 }
 
+void Analytics::TrackWindowSize(const std::string client_id,
+                                const toggl::Rectangle rect) {
+    std::stringstream ss;
+    ss << "winsize-"<< rect.str();
```
IndrekV: We should change winsize to `mainsize` as we are going to add tracking for other windows of the app in the future.

Also we want to track platform so the final format should be `os/mainsize`

`os` can be `win`, `mac` or `linux`

NghiaTranUIT:
I've updated by following your suggestion.

All are good now @IndrekV 💯

PR #2647, opened 2018-11-23
Added popup notice about not supported OS X version (mac), closes #2502

### LINKED ISSUE #2502: Clarify os support for mac
We should support only 3 of the latest os releases to avoid issues with old operating systems not supporting our latest developments.

- [x] Update knowledgebase info
- [x] Detect os version and notify user

At first show that os version is not supported and will be blocked on 1st of Jan 2019

If the date has passed block the app and show overlay with info that OS is not supported and user should update operating system or use web app.

update only when a new release is supported (if not supported stay on an old version and notify the user that the app may become unusable and is not supported)

IndrekV:
### What it does?
Detects version of Mac OS X and when the version is older than 10.11 shows the user an alert on app startup and when updater wants to check for a new version. Also when the date is later than 1st of January 2019, it closes the app immediately after the alert is closed.

### How to test?
You probably need to change the version number that is used to detect `unsupportedOS` and then change the deadline date in `UnsupportedNotice` class

When you start the app or open up About window the app should show the popup with the info.

### References
Closes #2502

### DIFF
--- a/src/ui/osx/TogglDesktop/AboutWindowController.m
+++ b/src/ui/osx/TogglDesktop/AboutWindowController.m
@@ -12,6 +12,11 @@
 #import "DisplayCommand.h"
 #import "Utils.h"
 #import "Sparkle.h"
+#import "UnsupportedNotice.h"
+
+@interface AboutWindowController ()
+@property BOOL unsupportedOS;
+@end
 
 @implementation AboutWindowController
 
@@ -39,6 +44,9 @@ extern void *ctx;
 	self.updateChannelComboBox.stringValue = [NSString stringWithUTF8String:str];
 	free(str);
 
+	NSOperatingSystemVersion osVersion = [[NSProcessInfo processInfo] operatingSystemVersion];
+	self.unsupportedOS = (osVersion.majorVersion == 10 && osVersion.minorVersion < 15);
+
 	if ([self updateCheckEnabled])
 	{
 		self.updateChannelComboBox.hidden = NO;
@@ -55,6 +63,12 @@ extern void *ctx;
 
 - (BOOL)updateCheckEnabled
 {
+	if (self.unsupportedOS)
+	{
+		[[UnsupportedNotice sharedInstance] showNotice];
+		return NO;
+	}
+
 	NSDictionary *infoDict = [[NSBundle mainBundle] infoDictionary];
 
 	return [infoDict[@"KopsikCheckForUpdates"] boolValue];
--- a/src/ui/osx/TogglDesktop/AboutWindowController.m
+++ b/src/ui/osx/TogglDesktop/AboutWindowController.m
@@ -14,10 +14,6 @@
 #import "Sparkle.h"
 #import "UnsupportedNotice.h"
 
-@interface AboutWindowController ()
-@property BOOL unsupportedOS;
-@end
-
 @implementation AboutWindowController
 
 extern void *ctx;
@@ -44,9 +40,6 @@ extern void *ctx;
 	self.updateChannelComboBox.stringValue = [NSString stringWithUTF8String:str];
 	free(str);
 
-	NSOperatingSystemVersion osVersion = [[NSProcessInfo processInfo] operatingSystemVersion];
-	self.unsupportedOS = (osVersion.majorVersion == 10 && osVersion.minorVersion < 15);
-
 	if ([self updateCheckEnabled])
 	{
 		self.updateChannelComboBox.hidden = NO;
@@ -63,9 +56,8 @@ extern void *ctx;
 
 - (BOOL)updateCheckEnabled
 {
-	if (self.unsupportedOS)
+	if (![[UnsupportedNotice sharedInstance] validateOSVersion])
 	{
-		[[UnsupportedNotice sharedInstance] showNotice];
 		return NO;
 	}
 
--- a/src/ui/osx/TogglDesktop/AboutWindowController.m
+++ b/src/ui/osx/TogglDesktop/AboutWindowController.m
@@ -56,11 +56,6 @@ extern void *ctx;
 
 - (BOOL)updateCheckEnabled
 {
-	if (![[UnsupportedNotice sharedInstance] validateOSVersion])
-	{
-		return NO;
-	}
-
 	NSDictionary *infoDict = [[NSBundle mainBundle] infoDictionary];
 
 	return [infoDict[@"KopsikCheckForUpdates"] boolValue];
@@ -68,6 +63,10 @@ extern void *ctx;
 
 - (void)checkForUpdates
 {
+	if (![[UnsupportedNotice sharedInstance] validateOSVersion])
+	{
+		return;
+	}
 	if (![self updateCheckEnabled])
 	{
 		return;
--- a/src/ui/osx/TogglDesktop/test2/AppDelegate.m
+++ b/src/ui/osx/TogglDesktop/test2/AppDelegate.m
@@ -93,7 +93,6 @@
 void *ctx;
 BOOL manualMode = NO;
 BOOL onTop = NO;
-BOOL unsupportedOS = NO;
 
 - (void)applicationWillFinishLaunching:(NSNotification *)not
 {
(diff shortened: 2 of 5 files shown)


src/ui/osx/TogglDesktop/AboutWindowController.m
```diff
@@ -12,6 +12,11 @@
 #import "DisplayCommand.h"
 #import "Utils.h"
 #import "Sparkle.h"
+#import "UnsupportedNotice.h"
+
+@interface AboutWindowController ()
+@property BOOL unsupportedOS;
```
NghiaTranUIT: Ideally, `AboutWindowController` shouldn't have a knowledge about the OS version or unsupportedOS logic. I suggest that `UnsupportedNotice` should be in charge.

As a result, this logic should be moved to `UnsupportedNotice` as well.
https://github.com/toggl/toggldesktop/blob/c52da165d527d423f208cb63a65c1826923773fb/src/ui/osx/TogglDesktop/AboutWindowController.m#L48

NghiaTranUIT: On the other hand, it would be nice if we start acquiring new modern objc-c style.
It means that it should be
```obj-c
// UnsupportedNotice.m
@property (nonatomic, assign) BOOL unsupportedOS;

// UnsupportedNotice.h
@property (nonatomic, assign, readonly) BOOL unsupportedOS;
```
Make sure that the public interface of `unsupportedOS` is only `readable`. We don't want other guys changed it by accident.

src/ui/osx/TogglDesktop/AboutWindowController.m
```diff
@@ -39,6 +44,9 @@ - (void)windowDidLoad
 	self.updateChannelComboBox.stringValue = [NSString stringWithUTF8String:str];
 	free(str);
 
+	NSOperatingSystemVersion osVersion = [[NSProcessInfo processInfo] operatingSystemVersion];
+	self.unsupportedOS = (osVersion.majorVersion == 10 && osVersion.minorVersion < 15);
```
NghiaTranUIT: By following the above approach, this duplicated code can be eliminated.

src/ui/osx/TogglDesktop/AboutWindowController.m
```diff
@@ -55,6 +63,12 @@ - (void)windowDidLoad
 
 - (BOOL)updateCheckEnabled
 {
+	if (self.unsupportedOS)
```
NghiaTranUIT: Then, reusing the logic in `UnsupportedNotice`
It could be
```objective-c
if ([[UnsupportedNotice shared] unsupportedOS])  {
    // Your logic
}
```

src/ui/osx/TogglDesktop/test2/AppDelegate.m
```diff
@@ -92,6 +93,7 @@ @implementation AppDelegate
 void *ctx;
 BOOL manualMode = NO;
 BOOL onTop = NO;
+BOOL unsupportedOS = NO;
```
NghiaTranUIT: Pls remove this unnecessary state. Ideally, we shouldn't introduce more `State`, as more as `State`, more maintenance cost.

src/ui/osx/TogglDesktop/test2/AppDelegate.m
```diff
@@ -100,18 +102,10 @@ - (void)applicationWillFinishLaunching:(NSNotification *)not
...
-		char *str = toggl_get_update_channel(ctx);
-		NSAssert(str, @"Could not read update channel value");
-		NSString *channel = [NSString stringWithUTF8String:str];
-		free(str);
-		[Utils setUpdaterChannel:channel];
-	}
 	[NSApp setActivationPolicy:NSApplicationActivationPolicyRegular];
+
+	NSOperatingSystemVersion osVersion = [[NSProcessInfo processInfo] operatingSystemVersion];
+	unsupportedOS = (osVersion.majorVersion == 10 && osVersion.minorVersion < 15);
```
NghiaTranUIT: This logic is appeared twice in our codebase. We should respect DRY. We could use the `unsupportedOS` in `UnsupportedNotice`

src/ui/osx/TogglDesktop/test2/UnsupportedNotice.m
```diff
@@ -0,0 +1,52 @@
...
+	NSDate *today = [NSDate date];
+	NSDate *deadline = [NSDate dateWithString:@"2019-01-01 00:00:00 +0000"];
+	NSComparisonResult result = [today compare:deadline];
+
+	NSString *title = [NSString stringWithFormat:@"Mac OS X version not supported"];
+	NSString *text = @"";
+
+	if (result == NSOrderedAscending)
+	{
+		text = @"Toggl Desktop will stop working with your current version of OS X from the 1st of January 2019.\n\n";
```
NghiaTranUIT: Off space ⚒

src/ui/osx/TogglDesktop/test2/UnsupportedNotice.h
```diff
@@ -0,0 +1,21 @@
...
+//
+//  Created by Indrek Vändrik on 23/11/2018.
+//  Copyright © 2018 Alari. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface UnsupportedNotice : NSAlert
```
NghiaTranUIT: IMO, it makes no sense that `UnsupportedNotice` is subclass from `NSAlert`, which is UI components.
The reason is that we're trying to support Singleton for a view, which is bad practice.

I suggest that `UnsupportedNotice` should be a true model, which subclass from `NSObject`.

src/ui/osx/TogglDesktop/test2/UnsupportedNotice.m
```diff
@@ -0,0 +1,52 @@
...
+
+	text = [text stringByAppendingString:@"Please upgrade your system to macOS 10.11 or later to continue using Toggl Desktop."];
+
+	[self setMessageText:title];
+	[self setInformativeText:text];
+	[self addButtonWithTitle:@"OK"];
+	[self setAlertStyle:NSWarningAlertStyle];
+	[self runModal];
+
+	if (result != NSOrderedAscending)
```
NghiaTranUIT: The code will terminal **before** the alert actually appear.
This one might be a mistake. According to our requirement
```
it closes the app immediately after the alert is closed.
```
I think that should be to make sure that user is able to read the reason why the app will terminal.
```objective-c
[alert beginSheetModalForWindow:self.window completionHandler:^(NSModalResponse returnCode) {
    if (returnCode == NSModalResponseOK) {
        // Handle terminal logic here
        return;
    }
   
}];
```

src/ui/osx/TogglDesktop/test2/UnsupportedNotice.m
IndrekV: Please elaborate what exactly do you mean by "Off space"?

src/ui/osx/TogglDesktop/test2/UnsupportedNotice.m
IndrekV: I double checked and nothing is executed that is after the `[self runModal];` before the "OK" is pressed in the modal. This should not be an issue.

src/ui/osx/TogglDesktop/test2/UnsupportedNotice.h
IndrekV: I don't understand the reasoning behind this. UnsupportedNotice is a simple Alert time popup for the user. Why would we need to create a true model out of it? It would make sense to create a model if we would decide to use it as a general notice that can be used with different scenarios. Currently, it is not the case and there is no reason to treat it that way.

src/ui/osx/TogglDesktop/test2/UnsupportedNotice.m
NghiaTranUIT: Ah, I just double check - It's good for now 👨‍💻

src/ui/osx/TogglDesktop/test2/UnsupportedNotice.h
NghiaTranUIT: I understand your reason. 

My concern is related to the change in the future. I predict that we might use this `UnsupportedNotice` class for many extensions soon. As a result, it's better be a true model, then next teammate, who are in charge, easily follow to extend new feature.

However, for now, it's okay to keep it as it is.

NghiaTranUIT:
Ultimately, the change look good to me. Feel free to merge it 👍

PR #2653, opened 2018-11-23
Use a stacked widget instead of showing and hiding vertically (linux)

MartinBriza:
This a part of my efforts of bringing the notifications in the line.

At this moment, the different views in the main window are laid out in one single layout and different parts (like TimeEntryList and TimeEntryEdit) get shown/hidden depending on the backend requests. This was not very easy to maintain because with the addition of more views, the programmer would have to check all the other views and make them hide when his new view goes into place.

QStackedWidget takes care of this as it's a widget wrapper that can contain multiple widgets and display only a single one of them at a time as requested. That means now with the addition of a new view, one would have to only request the QStackedWidget to display his new view and the currently shown one gets hidden without any extra effort.

As a nice bonus, everything seems much faster and more fluid when switching between views

### DIFF
--- a/src/ui/linux/TogglDesktop/loginwidget.cpp
+++ b/src/ui/linux/TogglDesktop/loginwidget.cpp
@@ -6,12 +6,10 @@
 
 #include "./toggl.h"
 
-LoginWidget::LoginWidget(QWidget *parent) : QWidget(parent),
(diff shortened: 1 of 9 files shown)


MartinBriza:
`ErrorViewController` still gets to be in a vertical layout because it is supposed to be displayed along with the other views at the same time.

PR #2654, opened 2018-11-23
[WIP] Use native Linux notifications instead of popping up QDialogs

MartinBriza:
Still a work in progress, TODOs:

- remove libXss dependency
- readd tray icon logic especially related to app shutdown
- and most importantly: rebase commits - split changes atomically and add meaningful descriptions

Depends on #2653 

What has been done:

- `IdleNotificationDialog` has been changed to a `QWidget` that takes the whole main window if an idle notification appears (to force user to decide what to do with the idle time).
- `IdleNotificationDialog` no longer depends on `XScreenSaver` to determine user idle time - now it relies on querying `org.freedesktop.ScreenSaver` for the same thing - it should work exactly the same but without a problem on Wayland, too.
- A `SystemTray` class has been added, wrapping up some of the past Tray Icon logic, but with the added possibility to handle notifications over `org.freedesktop.Notifications`, including sending images (which doesn't work for me with QSystemTrayIcon, at least in Plasma).


MartinBriza:
This PR is now ready for review

IndrekV:
Also I'll add this link here about "Idle detection in linux". I don't know if you considered it already but here it is http://crosstantine.blogspot.com/2009/08/idle-detector-c-linux.html

MartinBriza:
Hmm, you're right about the warnings from the stack layout, those seem to be benign but they shouldn't appear at all, I'll look into it.

MartinBriza:
As for the impossibility to use the DBus notifications and activity tracking, I'd say LXDE is pretty much abandoned at this point and when we talked about this last, you mentioned you want to support only "modern" environments. It seems your LXDE installation doesn't have a notification daemon. which should be installed by default (not sure why it is not in your case).

As for the https://unix.stackexchange.com/a/214074 link you've posted, the information there is actually wrong -  org.freedesktop.ScreenSaver is the baseline that should work everywhere, regardless of the environment - the ones mentioned further (gnome and kde) MAY implement something extra but they also HAVE TO expose org.freedesktop.ScreenSaver at the same time, so all apps can rely on that.

The idle detection from the blogspot link is how we've done it before and I've removed this functionality because it doesn't work well with Wayland and adds an extra library dependency we won't necessarily need.

I spent quite a bit of time investigating this today and I think supporting only the DBus APIs without keeping the legacy code (checking XScreenSaver and popping up QDialogs instead of notifications) is a good baseline for all modern DEs (GNOME, KDE, LXQt, Cinnamon and Mate). However, I can still reintroduce the old code as a fallback, if you want me to.

MartinBriza:
Next step: add a warning to the user if their environment doesn't  support notifications

MartinBriza:
Users will now be warned about the unavailable notifications in the preferences.

Also note that if this gets merged after CMake, tweaks to the CMakeLists.txt will be needed to add the QtDBus module.

IndrekV:
test

MartinBriza:
idle detection works on older systems as it did before but now it doesn't show a notification but a new view inside the application when you come back

IndrekV:
When I had reminder enabled before and switched to system without notification support it still showed reminder as enabled. When we show notification error the preferences should be disabled.
screenshot 2018-12-27 at 13 07 11

IndrekV:
I also tested idle detection on system with no notifications support. It just shows the idle form inside the app. I know that before it showed the popup in the front. Right now if the timer is not in the front you would not know it detected idle. I think we should bring the window to the front in this case so user is aware of the idle time detected by the app.

MartinBriza:
well, that's what we're trying to avoid with this patch - bringing attention to the app itself is a bad practice, focus stealing is something people don't like

IndrekV:
I remember focus stealing being this being one of the issues why we needed to rewrite this part of the app. Focus stealing is bad when user is in the middle of their work

Then again is it a problem with idle detection. Idle notification/popup should appear after user resumes from idle and at that point they are not doing anything as they are just starting to work again.

MartinBriza:
You're right this is not that big of an issue for idle detection. However, I thought we agreed that we don't need to support systems without the support for "modern" features like notifications.
I can certainly readd it so it will work as before but I'm not sure it'll work right in environments that active prevent focus stealing, like GNOME (more so under Wayland).

IndrekV:
Agreed. Let's keep it as it is and not put too much effort into supporting outdated systems. It's good we got to discuss all possible ways to approach this.

MartinBriza:
Some systems also support highlighting the icon in the status bar (if there is any) like the jumping icon on macOS or highlighted system bar entry in Windows.

IndrekV:
What changes are made to `third_party/openssl/include/openssl/opensslconf.h` in the last commit f9e8c3f900de68da779c778a7824e2f804f6c5b2 ?

MartinBriza:
no idea, i commited this file by mistake
i assume it's caused by the linter but i'm not sure
let me remove it from the commits real quick

MartinBriza:
i rebased the commits against master, to fix the merge conflicts
tell me if you want the commits squashed

NghiaTranUIT:
Regarding the change of `opensslconf.h`, It's generated as a result of `make deps`.

I had it too, but I manage to skip it when committing the code 🔨

IndrekV:
I mostly use `git add -p` to add files to the commit. That way I can look through all the changes and see if something strange has changed.

MartinBriza:
I do that too, sometimes, I just don't expect changes in files I did not touch, especially with smaller commits.

PR #2665, opened 2018-12-03
Fix the login widget layout (linux)

### LINKED ISSUE #2657: Login view looking compressed with small width
### 💻 Environment

Platform: Linux 

### 🐞 Actual behavior

UI items are vertically compressed together

### 💯 Expected behavior

UI should keep margins between elements

### 🔨 Steps to reproduce

1. Open up login view
2. Make window vertically as small as possible

### 📦 Additional info

screen shot 2018-11-26 at 12 19 43

MartinBriza:
### 📒 Description
Basically all spacing policies are removed, vertical spacers have been reorganized, two horizontal spacers were added for the checkbox and gaps are now handled using stylesheets instead of spacers (that created weird overlaps).

### 🕶️ Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)

### 🤯 List of changes
* Login widget is laid out in a bit more deterministic way

### 👫 Relationships
Closes #2657 

### 🔎 Review hints
Design review necessary, the layout is not 1:1 to how it was before.

### DIFF
--- a/src/ui/linux/TogglDesktop/loginwidget.cpp
+++ b/src/ui/linux/TogglDesktop/loginwidget.cpp
@@ -137,13 +137,13 @@ void LoginWidget::setCountries(
 void LoginWidget::on_viewchangelabel_linkActivated(const QString &link)
 {
     if (signupVisible) {
-        ui->signupFrame->hide();
-        ui->loginFrame->show();
+        ui->signupWidget->hide();
+        ui->loginWidget->show();
         ui->viewchangelabel->setText("<html><head/><body><a href='#' style='cursor:pointer;font-weight:bold;text-decoration:none;color:#fff;'>Sign up for free</a></body></html>");
         signupVisible = false;
     } else {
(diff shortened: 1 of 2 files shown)


MartinBriza:
### Before 
layout_old_signin
layout_old_signup

MartinBriza:
### After
layout_new_signin
layout_new_signup

NghiaTranUIT:
I'm setting the Linux workspace. Will review this ticket in next hours 👍

IndrekV:
I think we shouldn't change the height of the buttons and fields. They were bigger before the changes and should remain the same. The issue was that the margins between them were too small with small width of the window.

MartinBriza:
I didn't notice the height change at all, sorry about that.
Keeping the login widgets widths is a tad more involved though, because some of them (email, for instance) are used in both views.

MartinBriza:
Jumping has been fixed now.

PR #2676, opened 2018-12-07
Linux autocomplete

MartinBriza:
### 📒 Description
This adds a more advanced autocomplete functionality to the Description and Project ComboBoxes in the app (and the Timer).

#### Future plans
This only builds up from what we have now without really changing a lot of stuff in other parts of the project.
Section headers are now done, not filtered and not selectable.

### 🕶️ Types of changes

### 🤯 List of changes
* An item model to represent all of the data
* A new QListView to show the data, with a custom delegate to render the data using HTML
  * This does not apply to Time Entries because the HTML subset in QTextDocument is pretty limited - these are rendered "by hand"
* A custom QComboBox class that contains all the new features
  * The ComboBox filters the data through a custom QSortFilterProxyModel and feeds it to a QCompleter subclass to handle the autocompletion

### 👫 Relationships
-

### 🔎 Review hints
* Test if user input works as intended
* Check if all shown data is represented as required

### DIFF
--- a/src/ui/linux/TogglDesktop/CMakeLists.txt
+++ b/src/ui/linux/TogglDesktop/CMakeLists.txt
@@ -29,7 +29,7 @@ include_directories(
 set(BINARY_SOURCE_FILES
     aboutdialog.cpp
     autocompletecombobox.cpp
-    autocompleteitemview.cpp
+    autocompletelistview.cpp
     autocompletelistmodel.cpp
     autocompleteview.cpp
     clickablelabel.cpp
--- a/src/ui/linux/TogglDesktop/CMakeLists.txt
+++ b/src/ui/linux/TogglDesktop/CMakeLists.txt
@@ -28,6 +28,9 @@ include_directories(
 # TogglDesktop sources
 set(BINARY_SOURCE_FILES
     aboutdialog.cpp
+    autocompletecombobox.cpp
+    autocompleteitemview.cpp
+    autocompletelistmodel.cpp
     autocompleteview.cpp
     clickablelabel.cpp
     colorpicker.cpp
--- a/src/ui/linux/TogglDesktop/TogglDesktop.pro
+++ b/src/ui/linux/TogglDesktop/TogglDesktop.pro
@@ -49,8 +49,8 @@ SOURCES += main.cpp\
     systemtray.cpp \
     idlenotificationwidget.cpp \
     autocompletelistmodel.cpp \
-    autocompleteitemview.cpp \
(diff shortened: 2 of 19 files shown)


IndrekV:
- [ ] When mouse clicking down arrow in edit view project field the app crashed. The same crash happened when selecting items from the list.
- [ ] When typing something in the timer description field and pressing down key the selected item is the name of the Workspace and it is not possible to move with arrow keys. Probably because the next item is also disabled.
- [ ] Projects category name should not be shown if there are not any projects under it. Right now when filtering and no projects are shown the projects category is still visible
- [ ] Project items are missing the color dot and text does not have the color of the project
- [ ] Fix category item styles. Fon size should be smaller.
- [ ] There should be more space between description/project name/client name
- [ ] Non-category items should have a slight margin from the left for better separation

IndrekV:
Things are much better and there are no crashes anymore. Here are the issues that I found in the functionality:

- [ ] Category items and workspace name items should not be filtered and always visible if there is data under them. Currently the workspace name is hidden when it does not match the filter. 
- [ ] Selecting items does not populate fields other than description. If you select from description autocomplete only description is filled in and nothing else. If you select from project autocomplete then nothing is filled in.
- [ ] Pressing down arrow in the input field should open up dropdown with current filter. Example: if I filter the project field with character `k` and then close autocomplete and then press down arrow it still shows the filtered results, actually it should show the unfiltered results as there is nothing in the input field at that point.
- [ ] Pressing on the arrow right of the should act the same as the down key press described in the previous point. Currently it does not show anything but an empty dropdown.

MartinBriza:
Pressing right/left arrows should control the input field, in my opinion (to let the user work with the actual input string). I can implement it to use them, but this feels more intuitive. Also, the Windows version handles right/left buttons this way, too.

MartinBriza:
I think now all points except the left/right arrow buttons are resolved and autocomplete works as designed.

IndrekV:
The description autocomplete in edit view should work the same as the description input in timer where it fills in all the info that an time entry has. Description, Project, tags, billable. The last commit added tags, billable and other things, but project setting is still missing.

IndrekV:
Good job @MartinBriza !

Now it sets the project properly. One last change and we can merge and start working on creating a release.

When the timer is running the `x` button to delete project in timer should not be visible. The whole border and the `x` should be hidden at that point.

screenshot 2019-02-01 at 11 40 35

IndrekV:
Great job @MartinBriza , Finally we've won the battle with the new autocomplete.

Issue #2675, opened 2018-12-07
Acquire CocoaPod (Dependency Management)


NghiaTranUIT:
### 📒 Description
It's a standard to adopt one of popular Dependency Management in the objc/swift community.

The potential candidate I would like to nominate is [CocoaPods](https://cocoapods.org), which is the most popular and reliable 💯

### 💻 Environment
Platform: macOS

### 💥Current problem
- No official dependency management in our macOS codebase.
- It become a heavy burden when we would like to update all dependencies.
- Repository is bigger because we have to include all 3rd-party source-codes in our repo.

### 👑 Benefits
- Repo is smaller. Cloning time in CI/CD is quickly. We only have to fetch pods whenever we need in our workspace.
- Easy to upgrade and manage versions.
- Project structure is neat and cleaner because of all 3rd-party dependencies are move out to Pod Project (not in main project)

### ⚒ Things to changes
- Move all dependencies to CocoaPod
- Update the build script on our CI/CD to integrate with CocoaPod
- Support Caching Pod folder in CI/CD
- Update macOS installation's documentation.

NghiaTranUIT:
Technically, CocoaPod only serves well for **macOS** platform, and it doesn't support for the rest.

Grand scheme of all things, **Submodule** (#2736) could consider as a potential candidate.

Thus, we close this ticket ❌

Issue #2651, opened 2018-11-23
Acquire Diff Algorithm for TimeEntryListViewController's TableView


NghiaTranUIT:
Regarding #2061 and #2643 

From my understand after playing-around the both tickets, I realize the root of problems.

In order to avoid the connection break between Popup and the entry when the list loads.

I think that we should consider about Diff algorithm in to our `timeEntriesTableView`

## Current Problem
1. Every-time we close the Edit Popup, this entry will save. Consequently, the `timeEntriesTableView` is reloaded completely with new list of entires (same GUID, but different instances)

## Drawback
It leads to visible consequences
1. Undo feature (#2643) is more difficult to implement because those instances are changed every-time
2. Popup feature (#2061) is hard to implement

## How Diff works
1. We shouldn't reload entirely the table.
2. We only update or add or remove a specific row, which changed from new DataSource.
3. ```By using Diff algorithm, the complex calculation is done automatically, no error-prone and super efficient.```

## Benefits
1. The instances of EntryTime from Entry List are same
2. Undo feature is now feasible, because this instance is alive and it's able to keep old value.
3. Popup feature is possible, because the table doesn't reload entire view, so the connection is maintained.

## Ref
- https://github.com/onmyway133/DeepDiff
- https://github.com/jflinter/Dwifft
- We might implement the lightweight Diff version (https://nghiatran.me/diff-in-real-world-ios-part-2/)

## Time Complexity
- `Wagner–Fischer` algorithm has O(mn) complexity, so it should be used for collection with < 100 items.
- `Heckel algorithm ` is O(n) 💯
- `LCS` is O(mn)

NghiaTranUIT:
After preliminary investigation about "How to integrate Swift Diff library to out Objc codebase", I recognize that there are certain problems we're going to facing.

I will explain the problem and suggest certain possible solution (with pros and cons). Then we can make wise decision in order to acquire diff algorithm.

## Diff libraries
There are certain kinds of diff library, which implement different diff algorith out there.
Here are the noticeable one I can find.
1. [DeepDiff](https://github.com/onmyway133/DeepDiff) (Swift)
Overall, It's pretty neat, well-organized and efficient in term of time-complexity. It implements `Heckel algorithm`, so the result is incredible ([Benchmarks](Heckel algorithm)). The core diff functionality is same [IGListKit](https://github.com/Instagram/IGListKit) from Instagram does

2. [Dwifft](https://github.com/jflinter/Dwifft) (Swift)
[...]

Issue #2643, opened 2018-11-22
Adding undo function to text fields


IndrekV:
Fields in edit view should have undo history per time entry. When pressing cmd+z it should revert to the previous value.

When a new entry is loaded into the edit view the Undo history list should be cleared.

Example of expected behavior:
- Open entry details
- Type in "Test" to the description field
- Move to the next field (description is saved)
- Move back to the description field
- Change text to "Undo" and press cmd+z
- The text should revert to "Test"

NghiaTranUIT:
Block 🚫 by #2669

NghiaTranUIT:
Close due to #2695

Issue #2061, opened 2016-11-29
Make the edit view point to the correct item


IndrekV:
### 💻 Environment
Platform: macOS

### 🐞 Actual behavior
Edit popup closes when entries list is reloaded

### 💯 Expected behavior
Edit view should not close when entries list is reloaded. The popup should relocate itself to the updated item in the list.

### 🔨 Steps to reproduce

1. Open item for editing
2. Change date
3. List is reloaded and the popup is closed (this happens if commit fc4d66b65fbf3a957b2d2c2d6a69159e2afcdb62 is commited)

Currently the popup just points to the mainwindow and not to a specific item at all.

### 📦 Additional info

The main issue is that when list is reloaded the item is deleted and the focus goes away from the edit popup.

This commit adds the functionality: fc4d66b65fbf3a957b2d2c2d6a69159e2afcdb62

IndrekV:
Similar to #1938

PR #2679, opened 2018-12-10
Diff algorithm for TimeEntryListViewController

NghiaTranUIT:
### 📒 Description
It's a dedicated effort by adopting Diff algorithm for `TimeEntryListViewController`.

By introducing this diff, TimeEntryListViewController is capable of distinguish the changes more accurately (Insert, Remove, Move) rather than forcing ReloadData completely.

It's also a key factor to tackle #2061 and #2643.

### 🕶️ Types of changes
**Breaking change**: It changes the way how `TimeEntryListViewController's TableView` reload data.

### 🤯 List of changes
- [x] Add Deep Diff 3rd party directly (We haven't had Dependency Management yet #2675)
- [x] Introduce TableViewDiffer, which is responsible for interacting DeepDiff library (There is a limitation of Objc ([Detail reason](https://github.com/toggl/toggldesktop/issues/2651#issuecomment-445153874)) )
- [x] Introduce `NSTableView+TimeEntryDiff`, which is exposed to our Obj-C codebase.
- [x] Override `- [TimeEntryViewItem hash]` in order to calculate the different of each objects.
- [x] Support Diff for TimeEntryListViewController

### 👫 Relationships
#2651 

### 🔎 Review hints
1. Add `print("Changed = \(changed)")` at [TableViewDiffer](https://github.com/toggl/toggldesktop/blob/8c9712fcbc0f300ca0933425807e5d7320c91e63/src/ui/osx/TogglDesktop/Diff/TableViewDiffer.swift#L29)
2. Run the app and wait until it loads EntryItem successfully 
3. Click on any Entry, then close
4. If the log is 
```
Changed = []
```
It means the next data source, which is from Server, doesn't change the data source. In comparison with `master`. It will reload with new instance.


src/ui/osx/TogglDesktop/TimeEntryViewItem.m
```diff
@@ -168,4 +168,8 @@ - (NSString *)description
 			self.ProjectColor, self.billable, self.tags];
 }
 
+-(NSUInteger)hash {
+    return self.duration_in_seconds ^ self.Description.hash ^ self.ProjectAndTaskLabel.hash ^ self.TaskLabel.hash ^ self.ProjectLabel.hash ^ self.ClientLabel.hash ^ self.WorkspaceID ^ self.ProjectID ^ self.TaskID ^ self.duration.hash ^ self.ProjectColor.hash ^ self.ProjectGUID.hash ^ self.GUID.hash ^ self.billable ^ self.unsynced ^ self.started.hash ^ self.ended.hash ^ self.startTimeString.hash ^ self.endTimeString.hash ^ self.tags.hash ^ self.updatedAt.hash ^ self.formattedDate.hash ^ self.isHeader ^ self.dateDuration.hash ^ self.focusedFieldName.hash ^ self.CanSeeBillable ^ self.DefaultWID ^ self.CanAddProjects ^ self.WorkspaceName.hash ^ self.Error.hash ^ self.loadMore ^ self.confirmlessDelete ^ self.Group ^ self.GroupOpen ^ self.GroupName.hash ^ self.GroupDuration.hash ^ self.GroupItemCount;
```
NghiaTranUIT: It's super error-prone when dealing with `Hashable` or `Equatable` 💥. If we forget to update this func if we introduce new property, it could lead to untraceable issues. 

However, there is no choice to avoid this approach since Diff Algorithm use `Hashable` to distinguish each objects.

In the future, if we convert this class to Swift, we're able to adopt [Sourcery](https://github.com/krzysztofzablocki/Sourcery), and their [AutoHashable](https://github.com/krzysztofzablocki/Sourcery/blob/master/Templates/Templates/AutoHashable.stencil).

src/ui/osx/TogglDesktop/Diff/NSTableView+Diff.swift
```diff
@@ -0,0 +1,88 @@
...
+        completion?()
+    }
+
+    /// Execute removeRows/insertRows/moveRow appropriately
+    /// It relies on the input changesWithIndexPath
+    ///
+    /// - Parameter changesWithIndexPath: The IndexSet changes
+    private func internalBatchUpdates(changesWithIndexPath: ChangeWithIndexPath) {
+
+        changesWithIndexPath.deletes.reversed().forEach {
```
NghiaTranUIT: Regarding `reversed()`, we have to reverse the deletes class, because NSTableView's removeRow works differently than UITableView. Otherwise, it causes crash 💥 due to incorrect indexSet.

NghiaTranUIT:
I've jut rebased it. It's ready to review 👍

NghiaTranUIT:
@IndrekV I've rebased the master as well as fixed the UI issue.

Please make sure that you **delete** your local branch before checking out the latest branch on origin.

```
git checkout master
git branch -D feature/diff
git checkout feature/diff
```

NghiaTranUIT:
Awesome 🙌 

Time to rebase the Undo Func #2695 after this PR is merged successfully 🤖

Issue #2687, opened 2018-12-12
Use CMake instead of just a Makefile


MartinBriza:
### 💻 Environment
Platform: Linux and macOS

I suggest applying this to Linux first and to macOS later if this proves to be the right way to go.

### 📒 Description
Current way of building Toggl Desktop is not configurable, cannot use systemwide libraries and takes an insanely long time (about one hour on my Kaby Lake R laptop) without the possibility to build in parallel (`make -j`).

Hence I suggest switching to a different build system - CMake seems to be a good choice for a project of this size and complexity. It will generate Makefiles for us automatically based on the description of the project. It is also completely multiplatform and allows configuring the build process.

I put a basic `CMakeLists.txt` into the `cmake` branch in git to show what a difference it can make. For now, it uses systemwide `poco`, `lua`, `json` and `Qwt` but I'll modify it to let us choose if we want to use systemwide or bundled libraries.

The biggest benefit so far: building the whole project from scratch now takes around 50 seconds on my machine.

PR #2696, opened 2018-12-13
[WIP] Add compile-time switches to choose between prod and dev servers (lib)

MartinBriza:
### 📒 Description
This possibly breaks build bots, need to investigate Windows and Mac. Also, README on how to do a release has to be changed.

However, it also makes building the app a bit simpler because we don't have to tweak the stuff by hand, it will now be a simple compile-time switch in Makefile, QMake, VS or in CMake

When building the library, one can now specify:
* `-DTOGGL_PRODUCTION_BUILD=1` for a production build
* `-DTOGGL_ALLOW_UPDATE_CHECK=1` to allow the app to self-update
Not defining these variables makes the lib do the opposite.

### 🕶️ Types of changes

### 🤯 List of changes
Let production/development build type be defined by a single commandline switch when building

### 👫 Relationships
Makes stuff easier for #2687

### 🔎 Review hints
Check Mac and Windows builds and if buildbots can handle this instead of commenting out pieces of code "by hand". - I may do this myself tomorrow when I try to reach the buildbots again (or at least the Linux one).

### DIFF
--- a/src/context.h
+++ b/src/context.h
@@ -27,6 +27,18 @@
 #include "Poco/Timestamp.h"
 #include "Poco/Util/Timer.h"
 
+#ifdef TOGGL_ALLOW_UPDATE_CHECK
+# define UPDATE_CHECK_DISABLED false
+#else
+# define UPDATE_CHECK_DISABLED true
+#endif
+
+#if defined(TOGGL_PRODUCTION_BUILD) && !defined(APP_ENVIRONMENT)
+# define APP_ENVIRONMENT "production"
+#elif !defined(APP_ENVIRONMENT)
+# define APP_ENVIRONMENT "development"
+#endif
+
 namespace toggl {
 
 class Database;
--- a/src/context.cc
+++ b/src/context.cc
@@ -72,11 +72,11 @@ Context::Context(const std::string app_name, const std::string app_version)
 , next_update_timeline_settings_at_(0)
 , next_wake_at_(0)
 , time_entry_editor_guid_("")
-, environment_("production")
+, environment_(APP_ENVIRONMENT)
 , idle_(&ui_)
 , last_sync_started_(0)
 , sync_interval_seconds_(0)
-, update_check_disabled_(false)
+, update_check_disabled_(UPDATE_CHECK_DISABLED)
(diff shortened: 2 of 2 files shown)


MartinBriza:
This issue needs changes in the build process on all platforms (maybe in the Branding repository), feel free to push directly into the `prod-devel-compiletime` branch if you find an issue.

src/context.h
```diff
@@ -27,6 +27,18 @@
 #include "Poco/Timestamp.h"
 #include "Poco/Util/Timer.h"
 
+#ifdef TOGGL_ALLOW_UPDATE_CHECK
+# define UPDATE_CHECK_DISABLED false
+#else
+# define UPDATE_CHECK_DISABLED true
+#endif
+
+#if defined(TOGGL_PRODUCTION_BUILD) && !defined(APP_ENVIRONMENT)
```
IndrekV: I don't think testing `!defined(APP_ENVIRONMENT)` here is necessary. if `TOGGL_PRODUCTION_BUILD` is defined it is always "production" otherwise it's "development"

IndrekV:
For Mac the definitions can be added:
`
xcodebuild GCC_PREPROCESSOR_DEFINITIONS='$GCC_PREPROCESSOR_DEFINITIONS TOGGL_PRODUCTION_BUILD=1 TOGGL_ALLOW_UPDATE_CHECK=1'
`

src/context.h
IndrekV: I'm not sure I follow. Overriding what exactly, APP_ENVIRONMENT? Can you elaborate on a scenario when we would use it so I get a better idea?

MartinBriza: Well, having this in place doesn't really break anything, it just gives an opportunity to override the `production` / `development` choice by hand if necessary in the future, for instance if the servers get renamed, if we add a third one or in some other scenario.

IndrekV: all good, let's keep it.

NghiaTranUIT:
This PR impacts how Windows buildbot works. 

Here is the fix : https://github.com/toggl/toggldesktop-branding/pull/16

PR #2743, opened 2019-01-08
Restore / Save current windows frame in Linux

### LINKED ISSUE #1644: Save window height, width and location
saving is present in lib we should just use it.

NghiaTranUIT:
### 📒 Description
It's an effect to introduce "memory 📝" to Windows. 

Thus, Toggl Desktop can restore to latest sessions appropriately in term of size and position of windows 🖥 

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Introduce two funs (getWindowsFrameSetting and setWindowsFrameSetting) in `TogglApi`
- [x] Save current frame during CloseEvent
- [x] Restore previous frame during initialization.

### 👫 Relationships
Closes #1644 

### 🔎 Review hints
1. Open app and freely resize or change position of Entry List Windows
2. Close app by tapping X
3. Exit app on Menu
4. Open app again

If the position and size are same with last session -> Correct 💯 

### Note
- Don't stop app suddenly in Qt Debug because `void MainWindowController::closeEvent(QCloseEvent *event)` will never be executed.

### DIFF
--- a/src/ui/linux/TogglDesktop/mainwindowcontroller.cpp
+++ b/src/ui/linux/TogglDesktop/mainwindowcontroller.cpp
@@ -107,7 +107,6 @@ MainWindowController::MainWindowController(
     connect(TogglApi::instance, SIGNAL(updateContinueStopShortcut()),  // NOLINT
             this, SLOT(updateContinueStopShortcut()));  // NOLINT
 
-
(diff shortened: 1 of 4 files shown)


IndrekV:
If everything is good, merge it in @MartinBriza!

PR #2753, opened 2019-01-10
Refactor NotificationCenter

### LINKED ISSUE #2704: [Proposal] Refactor how we use NSNotificationCenter
### 📒 Description
In order to make our life (Toggl's fellows) is easier, I suggest that we adopt a elegant trick to ensure
- Observer notifications on Main thread
- Able to see the full Stack Tree
- Wipe out many boilerplate code 🍝 .

### 💻 Environment
Platform: macOS

### ❌  Problems
- The Toggl's Stack Trace is omitted unintentionally. As a result, it's super difficult to track back where are the caller during debugging session or Bugsnag or Crash report
 
screen shot 2018-12-14 at 17 43 34

The problem is that we're currently using 
```
[NSObject performSelectorOnMainThread:withObject:waitUntilDone:];
```
in order to performance the task in Main Thread. It's reasonable approach. However, the Stack Trace is sketchy and impossible to determine where are the parent caller.

For instance,
https://github.com/toggl/toggldesktop/blob/06af7edae0f8d447f87898e0d3d585c2c897b7ec/src/ui/osx/TogglDesktop/TimeEntryListViewController.m#L164-L169
https://github.com/toggl/toggldesktop/blob/06af7edae0f8d447f87898e0d3d585c2c897b7ec/src/ui/osx/TogglDesktop/TimeEntryListViewController.m#L302-L307

#### 1. NSNotificationCenter extension
We introduce helper func in order to send the notification on Main Thread appropriately.
```swift
extension NotificationCenter {

    func postOnMainThread(with aName: NSNotification.Name, object anObject: Any?) {
        runOnMainThreadIfNeed {[unowned self] in
            self.post(name: aName, object: anObject)
        }
    }
}

public func runOnMainThreadIfNeed(_ block: @escaping () -> Void) {
    if Thread.isMainThread {
        block() // If it's main thread, just execute 
        return
    }

    // Or run async on Main Thread later
    DispatchQueue.main.async {
        block()
    }
}
``` 
#### 2. The poster 
```obj-c
[[NSNotificationCenter defaultCenter] postOnMainThreadWithName:kDisplayTimeEntryList object:cmd];
```

#### 3. The Observer
The observer-sides don't change anythings. Just keep it as it is.
```obj-c
		[[NSNotificationCenter defaultCenter] addObserver:self
												 selector:@selector(startDisplayTimeEntryList:)
													 name:kDisplayTimeEntryList
												   object:nil];
```

### 🍰  Result
1. The observed code is executed in main thread
2. Able to see full Stack Trace

screen shot 2018-12-14 at 17 42 50

NghiaTranUIT:
### 📒 Description
This PR is the implementation from the [NotificationCenter proposal](https://github.com/toggl/toggldesktop/issues/2704)

### 🕶️ Types of changes

### 🤯 List of changes
- Introduce helper methods for NSNotificationCenter, to guarantee that it posts on Main Thread
- Replace all `[self postNotificationName:object:userInfo]`
- Replace all `[self performSelectorOnMainThread:withObject:waitUntilDone]`

### 👫 Relationships
Closes #2704 

### 🔎 Review hints
- Put a checkpoint in any Observers -> If we able to see the parent caller from TogglDesktopLibrary in Stack Tree -> 💯 
- No crash and the logics are remain 💯

### DIFF
--- a/src/ui/osx/TogglDesktop/AutocompleteDataSource.m
+++ b/src/ui/osx/TogglDesktop/AutocompleteDataSource.m
@@ -59,9 +59,7 @@ extern void *ctx;
 
 - (void)startDisplayAutocomplete:(NSNotification *)notification
 {
-	[self performSelectorOnMainThread:@selector(displayAutocomplete:)
-						   withObject:notification.object
-						waitUntilDone:NO];
+	[self displayAutocomplete:notification.object];
 }
 
 - (void)displayAutocomplete:(NSMutableArray *)entries
--- a/src/ui/osx/TogglDesktop/LoginViewController.m
+++ b/src/ui/osx/TogglDesktop/LoginViewController.m
@@ -12,6 +12,7 @@
 #import "UIEvents.h"
 #import "const.h"
 #import "NSTextField+Ext.h"
+#import "TogglDesktop-Swift.h"
 
 @interface LoginViewController ()
 @property AutocompleteDataSource *countryAutocompleteDataSource;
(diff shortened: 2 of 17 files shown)


NghiaTranUIT:
Rebase is done. Ready to review 🙌

PR #2763, opened 2019-01-17
Safe thread container for Toggl Desktop Library

### LINKED ISSUE #2747: Review all Read/Write problem on RelatedData class
### Description
Regarding https://github.com/toggl/toggldesktop/issues/2506#issuecomment-452302800, we have to take a look on this issue on all **RelatedData's** variables

### 🐞 Actual behavior
- There are many Read/Write problem on Related Data
- RelatedData is **not** safe-thread
- RelatedData share publicly its variables
https://github.com/toggl/toggldesktop/blob/9ea1df83aa324adfc54e04d01cbee21622dbd17f/src/related_data.h#L37-L47
- There are no any lock mechanism.

### 💯 Acceptance Criteria
- RelatedData class should be safe-thread

NghiaTranUIT:
### 📒 Description
According #2747, there is major concern about race-condition when iterating and modifying vector in different thread simultaneously, which causes many untraceable crashes.

This PR will introduce safe-thread container to mitigate the problem significantly.

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Introduce TBB as a submodule
- [x] Build on Makefile
- [x] Support concurrency_vector in library
- [x] Work on macOS
- [x] Work on Windows
- [x] Work on Linux

### 👫 Relationships
Close #2747 

### 🔎 Review hints
- Play around all logics of the app
- Able to build this PR on all platforms 💻 
- Create new entry with new project, new client. 
- Re-assign project / client to an existing entry.


NghiaTranUIT:
This PR has been done after testing carefully on all platforms. Thus, It's ready to review 👍 

Pls, make sure we're checking the latest commit on `origin` since I rebase/squash certain small commits together 📦

NghiaTranUIT:
I've resolved the conflict. Feels free to continue your review 😄 (don't forget to checkout the origin branch)

NghiaTranUIT:
@MartinBriza I understand your concern. I admit that it's silver-bullet to fix completely this issue.

Let me try to introduce some lock, so it could guarantee that there is no `clean()` during iterators.

NghiaTranUIT:
> Maybe using a smart pointer with a mutex to protect the entire vector could be a clearer solution.

Can you elaborate your idea? I'm not familiar with c++ yet, so I couldn't wrap my head about it, but overall, this idea could be better than `std::concurrency_vector`.

The thing we should take into account that the solution shouldn't introduce more boilerplate code during iterating, push_back or clear.

MartinBriza:
Ok, smart pointers probably won't be necessary, I think something like putting those vectors into the `private` part of the class and returining a structure with a mutex guard and a pointer/reference to the vector could be sufficient. That would mean having a `std::mutex` for each of the vectors and return a *probably* `std::unique_lock` along with the pointer to the `vector`. Then when we would be done with modifying the vector, we would just delete this structure and the destructor of `std::unique_lock` would take care of unlocking the mutex.
I'm not exactly sure yet about the technicalities, it's been a while since I've written anything with mutexes, si I'll think about it a bit more. 
Also, I'm not sure how to solve the problem of simultaneous READ accesses from different threads because these can occur at the same time without harm. WRITING is the problem we're trying to protect against...
So as I wrote, I'll come back to this issue after I'll have thought about it some more, unless you come up with something.

Issue #2427, opened 2018-04-12
Mac Desktop App Dock Icon Issue on 4K monitor


VeljkoTN:
A user reported that the dock icon for Toggl Desktop for Mac is not showing up (instead there's a generic icon in its place) we're suspecting it's because of that 4K pixel density, here's how it shows: 

![image](https://user-images.githubusercontent.com/36152180/38668095-a08b900a-3e43-11e8-8a04-c813901a08d5.png)

The user tried uninstalling and re-installing the app and it didn't help.

Let me know if you need any additional information.

case/675299

PR #2769, opened 2019-01-23
Upgrade the Toggl AppIcon by using assets

IndrekV:
This is a second try of the PR #2631 

## ❓ What's this?
There is a report that the AppIcon doesn't work well in 4K (toggl/toggldesktop#2427). However, I couldn't reproduce the issue by using [SwitchResX](http://www.madrau.com) to simulate 4K screen and the codebase looks okay.

According to [App Icon Attributes](https://developer.apple.com/design/human-interface-guidelines/macos/icons-and-images/app-icon/), Apple recommends that we should use `ImageAsset` and 
```
Don't provide app icons in ICNS or JPEG format. Add de-interlaced PNG files in the app icon fields of your Xcode project's asset catalog.
```
We're using **ICNS** to set the App Icon directly, so I **believe** that it could be the issue in some edge scenario.

## 💾 How is it done?
Following the Apple's recommendation, we have to remove ICNS and old code then adopting new `ImageAsset`.

## 🤯 Changelogs 
- [x] Remove old `app.icns` and `app_inactive.icns`
- [x] Adopt ImageAsset and add AppIcon (Xcode is smart enough to generate it automatically from our ICNS)
- [x] Introduce `AppIconFactory` to handle App Icon appropriately.
- [x] Remove old code in AppDelegate
- [x] Replace old code with new image from `AppIconFactory`

## 👫 Relationships
Closes toggl/toggldesktop#2427

## 🔎 Review hints
- Run the build and check the icon before and after start the timer. If it change from "Grey" to "Red" when tracking. It's correct behavior 💯


NghiaTranUIT:
Here is the testing result 📖 

- [x] Has proper inactive_icon and active_icon when start and stop the timer.
- [x] Has `AppIcon.icns` and `Assets.car` (has icons) in DestopToggl.app, which are built from `make app` and `Build -> Archive`
- [x] Has icons appropriately in macOS 10.14 
- [x] Has icons appropriately in macOS 10.11 (Parallels) (Fixed https://github.com/toggl/toggldesktop/pull/2769/commits/5d723908725b3c6bde1f181c146f8a70a6610878)

src/ui/osx/TogglDesktop/test2/AppIconFactory.m
```diff
@@ -0,0 +1,32 @@
...
+{
+	NSImage *icon;
+
+	switch (type)
+	{
+		case AppIconTypeActive :
+			icon = [NSImage imageNamed:@"AppIconActive"];
+			break;
+		case AppIconTypeDefault :
+			icon = [NSImage imageNamed:NSImageNameApplicationIcon];
```
NghiaTranUIT: It's better to use `NSImageNameApplicationIcon` rather than "AppIcon", because it will get correct size icon depend on the screen resolution.

If not, the icon quality is pretty bad, because it gets 16px icon on **OS 10.11**.

## Before
screen shot 2019-01-24 at 10 39 04 am

## After
screen shot 2019-01-24 at 10 40 02 am

IndrekV:
If everything is ok, merge it in @NghiaTranUIT

NghiaTranUIT:
Sure, but let me resolve the conflict firstly 👍

PR #2770, opened 2019-01-24
Support custom error message from Response

### LINKED ISSUE #2431: Add support for json error messages
Api will implement json error message in the future and we should add support of that to the lib.

json structure:

```
{
    "message" : "error message that is shown to the user"
}
```

NghiaTranUIT:
### 📒 Description
This PR will introduce the minimal changes in order to adopt the customizable error message, which is bundled from Toggl's API responses.

Technically, it will check whether `error_message` key is existed or not:
- If not, it will fallback to current implementation.
- If it presents, we extract it and present it by combining with the current local error by following the format: 
```
{local_error_message} + {error_message}(if presents)
```
For instance, if user tries signing in with incorrect credentials. 

`Invalid email/password. Your account email is expired` will be present. 

`Your account email is expired` is customizable error message from Toggl Server.

By following this way, we **diminish** significantly the rate of change our current codebase, since the error mechanism in Desktop Library is so complicated.

### 🕶️ Types of changes
**New feature**

### 🤯 List of changes
- [x] Extract `error_message` and pass it out
- [x] Fallback to current error logic if `error_message` is not presented. 
- [x] Update `toggl_on_error` interface on all platforms.

### 👫 Relationships
Closes #2431 

### 🔎 Review hints
- Since the backend team doesn't return any `error_message`, so it's hard to test it on production. Thus, we have to fake it by hard-coding it.
- On the other hand, all current error logic should be remained.

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -970,6 +970,10 @@ bool Context::isPostponed(
 }
 
 error Context::displayError(const error err) {
+    return displayError(err, "");
+}
+
+error Context::displayError(const error err, const std::string message) {
     if ((err.find(kUnauthorizedError) != std::string::npos)) {
         if (user_) {
             setUser(nullptr);
@@ -996,6 +1000,9 @@ error Context::displayError(const error err) {
         }
     }
 
+    if (!message.empty()) {
+        return UI()->DisplayError(message);
+    }
     return UI()->DisplayError(err);
 }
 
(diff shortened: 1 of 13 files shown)


IndrekV:
I used a more simple solution. The logic behind it is.

- We currently use only the status code to translate responses to our own error messages.
- We never show the error received from API straight to the user
- If there will be translated error messages in the future it will be handled the same way. (converting `error_code` received from API to a translation inside the app.
- there is no need to have 2 error message strings as introduced in #2770

Issue #2374, opened 2018-01-29
Update titlebar design


IndrekV:
Update titlebar to blend into the app.

Add to mainwindowcontroller`windowDidLoad`
```
    self.window.titleVisibility = NSWindowTitleHidden;
    self.window.titlebarAppearsTransparent = YES;
    self.window.styleMask |= NSFullSizeContentViewWindowMask;
```

https://github.com/robin/TitlebarAndToolbar

Should be tested if works with older systems.

PR #2774, opened 2019-01-28
New Login and Sign up UI

### LINKED ISSUE #2767: Login view design update
### 💻 Environment
Platform: macOS

### 📒 Description
The functionality will remain the same, only the design will change. All the info about design will be in the design files.

Things to note:
- We have light and dark design
- There are two views: login and signup
- Error message display logic has been changed. The new logic shows error message as a floating item in the bottom. This error message logic will be used across the whole app with slight adjustments in the main app view.
- Also as this is the first thing we'll implement then this will be the time to update the window design.   We decided to not use the default grey window title and will use a slightly lighter window header. Changing header title logic is described in issue #2374

NghiaTranUIT:
### 📒 Description
This PR introduces the new login and sign up design and keep majority of current logics.

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Implement Floating Error Box
- [x] Implement new design for Login and Sign uo
- [x] Adopt light/dark theme appropriately (title, background, pattern, ...)
- [x] Floating Error box's width should be fixed and stick on right side

### 👫 Relationships
Close #2767 

### 🔎 Review hints
1. Double check the UI with the design (in both light and dark theme)
2. Error floating behaves properly.
3. Try to login, sign up, search country and login with google.
4. Try to click on Term and Privacy link


NghiaTranUIT:
Photos:
screen shot 2019-01-28 at 15 20 59
screen shot 2019-01-28 at 15 21 31
screen shot 2019-01-28 at 15 21 25
screen shot 2019-01-28 at 15 21 14

NghiaTranUIT:
I'm not sure if #2374 is still available because it doesn't match the design @IndrekV 🤔 

If we follow #2374, there are 1 issues:

1. There is a overlap in Entry List
screen shot 2019-01-28 at 15 45 45

2. Missing the Title Bar compared with the Design
screen shot 2019-01-28 at 15 45 24

NghiaTranUIT:
The conflict is gone 💨 

It's ready for reviewing 🐝

NghiaTranUIT:
Regarding `* Fix tabbing and focusing all items`.

From what I see from `Stable TogglDesktop 7.4.336`, here is how the tab works
### 1. Login screen
- Tab move around between the Email and Password textfield
### 2. Sign-up screen
- Tab move around between the Email, Password and Country Dropdown.

So, if we intend to extend the tab func, I will do it now 😄 👍

NghiaTranUIT:
Regarding `* The focus ring is smaller than the field and looks strange`, the problem is because we set the height of NSTextField, which is 30pt. So, the FocusBox doesn't render appropriately.

It leads to be smaller than the outer frame.

Here is example
## Original NSTextField
Just drag it down to storyboard, the default height is 22pt.
The Focus box draws perfectly 💯 
screen shot 2019-02-05 at 14 42 52

## Our customize NSTextBox
In order to respect the design, I set the TextBox's height is 30pt. So it results in the mis-place of FocusBox (zoom in to see clearly)
screen shot 2019-02-05 at 14 42 59

## Conclusion
I agree that we might remove the FocusBox temporary until we figure out how can manipulate to the drawing process of FocusBox 😄

NghiaTranUIT:
> Is it possible to remove the bottom border in dark mode.

I'm aware it too, but I couldn't get rid of it unless I have to override and do a drawing process by my own.

NghiaTranUIT:
All issue from your review are resolved @IndrekV 👍 

Don't forget to enable "Tab on All Control" in Keyboard (https://superuser.com/a/372690) 🍺

IndrekV:
Added some minor UX tweaks but overall it's good.

PR #2782, opened 2019-01-30
Use CMake instead of Make to build the Linux package

MartinBriza:
### 📒 Description
This changes the package.sh script for linux to use CMake instead of plain Makefile to create the Linux tarball.

### 🕶️ Types of changes

### 👫 Relationships
Depends on #2696


MartinBriza:
After reviewing this, toggl/toggldesktop-branding#14 should be reviewed/merged

PR #2785, opened 2019-02-01
Background thread for APIs

### LINKED ISSUE #2765: Execute Toggl APIs on main thread
Technically ,TogglDesktop is usually **frozen** during bad connection.

After investigation, it reveals that there are few APIs which are executed on Main Thread.

### 💻 Environment
Platform:  All platform macOS/Windows/Linux 
Toggl Version: Latest commit on Master

### 🐞 Actual behavior
- There are certain APIs, which are executed on Main Thread.
- Cause frozen app during bad network connection.

### 💯 Expected behavior
- Should be executed in background thread.
- Main thread should set free.

### 🔨 Steps to reproduce

1. Download [Network Conditioner](https://developer.apple.com/download/more/?q=Additional%20Tools)
2. Set **3G** profile
3. Open app and do actions, which execute the following APIs

- Login
- List Country in Sign up menu
- Sign up
- Report (/api/v8/desktop_login_tokens)

4. App is frozen in short time during requesting data from server.

### 📦 Additional info
For instance, Login API is running on Main Thread.
screen shot 2019-01-23 at 11 47 33

NghiaTranUIT:
### 📒 Description
There is currently issue that TogglDesktop is always frozen during login and sign-up in unreliable connection.
 
This PR will introduce the fix by executing those APIs in background thread.

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Introduce async version for `Login`, `Sign-up`, `Report` and `GoggleLogin`, `PullCountry` in library.
- [x] Introduce Loader during fetching data on background in LoginViewController (mac).
- [x] Same functionality for Windows.
- [x] Same functionality for Linux.

### 👫 Relationships
Close #2765 

### 🔎 Review hints
1. Download [Network Conditioner](https://developer.apple.com/download/more/?q=Additional%20Tools)
2. Set **3G** profile
3. Open app and do actions, which execute the following APIs

- Login
- List Country in Sign up menu
- Sign up
- Login with Goggle
- Report (/api/v8/desktop_login_tokens)

=> The logic of Login, Signup, Login with Google and Report are remained => 💯 
=> The app doesn't frozen anymore (The loader is shown and hide appropriately) => 💯

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -3779,6 +3779,14 @@ void Context::SetSleep() {
     }
 }
 
+error Context::AsyncOpenReportsInBrowser() {
+    std::thread backgroundThread([&]() {
+        return this->OpenReportsInBrowser();
+    });
+    backgroundThread.detach();
+    return noError;
+}
+
 error Context::OpenReportsInBrowser() {
     // Do not even allow to open reports
     // else they will linger around in the app
--- a/src/context.cc
+++ b/src/context.cc
@@ -56,6 +56,8 @@
 #include "Poco/Util/TimerTask.h"
 #include "Poco/Util/TimerTaskAdapter.h"
 #include <mutex> // NOLINT
+#include <thread>
+#include <future>
 
 namespace toggl {
 
@@ -2169,6 +2171,16 @@ error Context::attemptOfflineLogin(const std::string email,
     updateUI(UIElements::Reset());
 
     return save();
+}
+
+error Context::asyncLogin(const std::string email,
+                          const std::string password) {
+    std::future<error> resultFromDB = std::async([&](std::string email, std::string password) {
+        return this->Login(email, password);
+    }, email, password);
+
+    //error result = resultFromDB.get();
+    return noError;
 }
 
 error Context::Login(
--- a/src/context.cc
+++ b/src/context.cc
@@ -57,7 +57,6 @@
 #include "Poco/Util/TimerTaskAdapter.h"
 #include <mutex> // NOLINT
 #include <thread>
-#include <future>
 
 namespace toggl {
 
@@ -2175,11 +2174,10 @@ error Context::attemptOfflineLogin(const std::string email,
 
 error Context::asyncLogin(const std::string email,
                           const std::string password) {
-    std::future<error> resultFromDB = std::async([&](std::string email, std::string password) {
+    std::thread backgroundThread([&](std::string email, std::string password) {
         return this->Login(email, password);
     }, email, password);
-
-    //error result = resultFromDB.get();
+    backgroundThread.detach();
     return noError;
 }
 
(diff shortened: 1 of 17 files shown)


NghiaTranUIT:
After introducing `async` interface in https://github.com/toggl/toggldesktop/pull/2785/commits/365086cd559dd9516d3d0998af96c84f42fa0452, the Windows version works as we expected without any problem.

Because Windows 💻  has already support run on main thread from client side.
https://github.com/toggl/toggldesktop/blob/a74fd9ca4236c8d75d41beeb49947af1f2ef7bb4/src/ui/windows/TogglDesktop/TogglDesktop/ui/views/LoginView.xaml.cs#L178-L191

NghiaTranUIT:
I understand your concern. I agree that we have to handle the error from async work. It's good practice when dealing with concurrency app.

The problem is that the original approach doesn't handle any Response Error at all. All error are passed through the `Context::DisplayError` func. Then the app will present the alert.

## Current error mechanism in Toggl Desktop Library 📚 
1. We haven't handled any error from sync or async API request yet.
For examples:
- Original login ignore any errors, which is returned.
https://github.com/toggl/toggldesktop/blob/a74fd9ca4236c8d75d41beeb49947af1f2ef7bb4/src/ui/osx/TogglDesktop/LoginViewController.m#L81-L84

- Async `LoadMore` func doesn't care any error response
https://github.com/toggl/toggldesktop/blob/a74fd9ca4236c8d75d41beeb49947af1f2ef7bb4/src/ui/osx/TogglDesktop/test2/LoadMoreCell.m#L43

- And so many

2. We are using `Context::Display` as a global func to notify the app whenever there is any error from `any` source. IMO, it's super terrible idea, since we don't know that is the source of error. For that reasons, there are many un-related `if` statement to check where is the source.

https://github.com/toggl/toggldesktop/blob/a74fd9ca4236c8d75d41beeb49947af1f2ef7bb4/src/context.cc#L972-L1000

=> It's also the obstacle when I want to support `Error from JSON` from #2770 

For instance,
- If the `Login API` has errors (Incorrect username/pass, ...), it will execute the `Context::DisplayError`.
https://github.com/toggl/toggldesktop/blob/a74fd9ca4236c8d75d41beeb49947af1f2ef7bb4/src/context.cc#L2219-L2238

- There are many clumsy logics in `Context::Display` func as well.

## Then
For all of existing problems, I couldn't extend the scope of my ticket in order to refactor and improve our current error mechanism. 

I assume that the workload for this error refactor is bigger and more risks.

So, I just prefer to keep it as minimal as possible. Then we can refactor the error mechanism later.

NghiaTranUIT:
All platforms seems to be good right now 🙋‍♂️ Let make it count guys 👍

MartinBriza:
Yes, I think using displayError (or a similar function) would be reasonable. We just have to take into account that it is running in a different thread when using it.

NghiaTranUIT:
@IndrekV if you're reviewing this ticket, please checkout new branch from origin (I've just rebased from `master` and integrate the LoaderView in new Login design) 👍

PR #2796, opened 2019-02-07
Fix the missing GUID error in Linux

### LINKED ISSUE #2750: Missing GUID warning in linux
### Description
There are GUID Warning ⚠️  in Linux during development.

### 💻 Environment
Platform: Linux
OS Version: Ubuntu 18.04.1 LTS
QTCreator: 4.8.0
Tq: 5.11.1
Toggl Version: https://github.com/toggl/toggldesktop/commit/9ea1df83aa324adfc54e04d01cbee21622dbd17f

### 🐞 Actual behavior
- There are GUID Warning ⚠️  in Linux during development.

### 💯 Expected behavior
- GUID Warning should be gone

### 🔨 Steps to reproduce

1.  Check commit https://github.com/toggl/toggldesktop/commit/9ea1df83aa324adfc54e04d01cbee21622dbd17f in master
2. Debug with Qt Creator
3. Select any existing entry
4. The Edit Entry will present
5. The ⚠️ appears

MartinBriza:
### 📒 Description
Some changes in the time entry editor are handled by the event filter. These can be triggered even in the moment when there is no GUID assigned, yet. This PR solves this by not setting anything when the description field has been focused out and there is no GUID yet.

### 🕶️ Types of changes

### 🤯 List of changes
- "Missing GUID" messages now show the function where it is missing, too
- the missing GUID issue is fixed for the Linux client

### 👫 Relationships
Closes #2750 

### 🔎 Review hints
See if edge cases are handled

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -2722,7 +2722,7 @@ TimeEntry *Context::Continue(
     }
 
     if (GUID.empty()) {
-        displayError("Missing GUID");
+        displayError(std::string(__FUNCTION__) + ": Missing GUID");
         return nullptr;
     }
 
@@ -2768,7 +2768,7 @@ error Context::DeleteTimeEntryByGUID(const std::string GUID) {
     }
 
     if (GUID.empty()) {
-        return displayError("Missing GUID");
+        return displayError(std::string(__FUNCTION__) + ": Missing GUID");
     }
 
     TimeEntry *te = nullptr;
@@ -2810,7 +2810,7 @@ error Context::SetTimeEntryDuration(
     const std::string GUID,
     const std::string duration) {
     if (GUID.empty()) {
-        return displayError("Missing GUID");
+        return displayError(std::string(__FUNCTION__) + ": Missing GUID");
     }
 
     Poco::Mutex::ScopedLock lock(user_m_);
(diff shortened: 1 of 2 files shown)


src/context.cc
```diff
@@ -2722,7 +2722,7 @@ TimeEntry *Context::Continue(
     }
 
     if (GUID.empty()) {
-        displayError("Missing GUID");
+        displayError(std::string(__FUNCTION__) + ": Missing GUID");
```
NghiaTranUIT: It's good for debugging easier, but

I'm not sure that the present of `__FUNCTION__ + "Missing GUID"` to user interface is a good idea 🤔 since it's meaningfulness in term of ordinary user perspective 👀 

I suggest that we can remove it, and the fix `!guid.isEmpty()` is sufficient 👍

NghiaTranUIT:
Overall, the fix works well in all platforms. I didn't see any "Missing GUID" anymore 👍

src/context.cc
MartinBriza: `__FUNCTION__` is a macro that will convert to the wrapping method name: "Continue" in this case. User would see this string: "Continue: Missing GUID" in the error widget.

IndrekV: The text "Missing GUID" itself is also not very easy to understand for the user. So I think it's ok to add this detail so users can give us more specific error message when they contact us with the issue.

PR #2809, opened 2019-02-13
Improve and flatten menu bar handling (linux)

### LINKED ISSUE #2660: Make Tray Icon Menu One Level Deep
### 💻 Environment

Platform: Linux (openSUSE Leap 15.0 Mate Desktop)

### 📒 Description
When right clicking on the tray icon, a menu pops up with only one row, "Toggl Desktop >", and then clicking on that opens another menu level with everything you can do from the app.
toggl-menu

The "Toggl Desktop" level should be removed so that right clicking only shows the main menu.

An alternative would be to move the primary activity options (`New`, `Continue`, `Stop`, `Show`, `Quit` and/or `Logout`) into the top level and keep the rest as it is (or remove it, since it's less likely to be needed from the tray anyway).

### ⭐️ Why do you want this?
It's an extra click, and it's annoying to have to do every time I want to start or stop my timer.

MartinBriza:
### 📒 Description
This fixes the tray icon menu to display just as one level instead of two (in some environments where that happened). It also simplifies the menu creation process by quite a bit.

### 🕶️ Types of changes

### 🤯 List of changes
The weird reassignment of QActions in MainWindowController is not present anymore and the actions are now accessed directly through the `ui` member.

### 👫 Relationships
Closes #2660

### 🔎 Review hints
See if the tray has only 1 level and is the same as the menu in the main application (not like in #2660)

### DIFF
--- a/src/ui/linux/TogglDesktop/mainwindowcontroller.cpp
+++ b/src/ui/linux/TogglDesktop/mainwindowcontroller.cpp
@@ -38,15 +38,6 @@ MainWindowController::MainWindowController(
   togglApi(new TogglApi(nullptr, logPathOverride, dbPathOverride)),
   tracking(false),
   loggedIn(false),
-  actionEmail(nullptr),
-  actionNew(nullptr),
-  actionContinue(nullptr),
-  actionStop(nullptr),
-  actionSync(nullptr),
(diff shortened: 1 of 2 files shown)


IndrekV:
Well done!

PR #2816, opened 2019-02-18
[For Reference only] Safe thread container for Toggl Desktop Library 2.0

### LINKED ISSUE #2747: Review all Read/Write problem on RelatedData class

MartinBriza:
### 📒 Description
Everything is now returned as a `std::pair` through getters

`first` element of that pair is a lock that gets closed automatically when scope ends or explicitely by using `pair.first.unlock()`

`second` element is the actual pointer to the vector itself. Everything is now accessed through pointers - this will require some changes to the other methods, too, to increase efectiveness because in a lot of places, the vectors were passed by value.

### 🕶️ Types of changes

### 🤯 List of changes
Accessing members in `related_data` is now done through getter methods. It is absolutely necessary to NOT discard the returned `std::pair` until done with working with the vectors because the `first` member is the mutex lock (destroying it releases the mutex).

### 👫 Relationships
Another take on #2763

Closes #2747

### 🔎 Review hints
Verify build on all platforms and look for potential crashes.

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -809,10 +809,8 @@ void Context::updateUI(const UIElements &what) {
         if (what.display_autotracker_rules && user_) {
             if (UI()->CanDisplayAutotrackerRules()) {
                 // Collect rules
-                for (std::vector<toggl::AutotrackerRule *>::const_iterator
-                        it = user_->related.AutotrackerRules.begin();
-                        it != user_->related.AutotrackerRules.end();
-                        it++) {
+                auto ATRs = user_->related.GetAutotrackerRules();
+                for (auto it = ATRs.second->begin(); it != ATRs.second->end(); it++) {
                     AutotrackerRule *model = *it;
                     Project *p = user_->related.ProjectByID(model->PID());
                     Task *t = user_->related.TaskByID(model->TID());
@@ -3546,7 +3544,8 @@ error Context::AddAutotrackerRule(
             rule->SetPID(p->ID());
         }
         rule->SetUID(user_->ID());
-        user_->related.AutotrackerRules.push_back(rule);
+        auto ATRs = user_->related.GetAutotrackerRules();
+        ATRs.second->push_back(rule);
     }
 
     error err = save();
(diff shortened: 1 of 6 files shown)


MartinBriza:
Now, there is a lot of space for improvement.
Initially, I tried to create a class that would contain the lock and the pointer at the same time so we wouldn't have to modify as much code as we do at this point.
I got a lot of crashes when passing the `lock` through `std::move` to a move constructor of the class so I opted for using `std::pair` in the end.

MartinBriza:
Oh, I also changed a lot of for loop iterators from the `std::vector<very>long::name` types to `auto` along the way

MartinBriza:
I'm not able to reproduce this in Linux nor Mac. Can you please specify what actions seem to trigger this? At least a rough idea of what to do to make it freeze would suffice.

NghiaTranUIT:
## Description
Create new Client cause frozen 

## How to reproduce
1. Open the app, and select any Time Entry
2. Click "Add new Project" -> "Add new Client"
3. Type any name, and tab on "Add new Client"
4. 💥 

## GIF

MartinBriza:
Thanks, this happens in Linux too. I'll look into it.

MartinBriza:
I have resolved the reported freezing issue (and other possible freezing sources). Now everything should work right. I'll just rebase the branch against current master.

MartinBriza:
Mhm, the other issue here seems to be adding projects to clients doesn't work neither here, nor in master.

NghiaTranUIT:
@MartinBriza is it possible to rebase this branch from `master` again? Since we had a huge merge from TimeEntry Editor days ago 😄 

I just do a quick tests and it seems to be good for now. No freeze when creating new client/project/time entry.

MartinBriza:
Sure, done.

NghiaTranUIT:
All good now 👍 I was able to do some basic actions without any problems. Hopefully it solved many known-crashes when quitting the app.

IndrekV:
This PR is here only for reference and is **not** to be reviewd or merged. This solution does not fix the problem it was meant to fix and we'll keep this open as a reference until we figure out an alternative solution.

PR #2819, opened 2019-02-19
Elide the text in the timer widget (linux)

### LINKED ISSUE #2661: Dropdown should compress from the right/end of text
### 💻 Environment
Platform: Linux (openSUSE Leap 15.0 Mate Desktop)

### 📒 Description
The text in the "What are you doing?" drop down box is anchored at the end, so when it is too long, the beginning of the text is pushed out and only the end of the task name is visible.
long-name-task
I think the opposite should happen, with the text fixed at the start and ellipsized from the end (or in the middle — as is the case with inactive drop-down entries).
toggle-dropdown

MartinBriza:
### 📒 Description
The timer showed only the end of the current time entry's label. Now it will show the beginning and elide the end of the string if it's too long.
The dropdown arrow will be also hidden.

### 🕶️ Types of changes

### 🤯 List of changes
The timer widget now contains a QLabel in addition to the AutocompleteComboBox that shows the current time entry text when the timer is running.

### 👫 Relationships
Closes #2661 

### 🔎 Review hints
Check if everything looks right and behaves as expected in the timer widget.

### DIFF
--- a/src/ui/linux/TogglDesktop/timerwidget.cpp
+++ b/src/ui/linux/TogglDesktop/timerwidget.cpp
@@ -45,6 +45,9 @@ selectedProjectId(0) {
             this, SLOT(descriptionBillableChanged(bool)));
     connect(ui->description, SIGNAL(tagsChanged(QString)),
             this, SLOT(descriptionTagsChanged(QString)));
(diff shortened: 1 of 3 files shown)


IndrekV:
Well done!

PR #2823, opened 2019-02-21
Make opening running TEs by keyboard possible on Linux

### LINKED ISSUE #2262: Add shortcuts to the Linux version
When [logging](https://support.toggl.com/wp-content/uploads/2012/03/Screenshot_2017-01-02_19-00-57.png), please make a keyboard shortcut to open the [details](https://support.toggl.com/wp-content/uploads/2012/03/Screenshot_2017-01-02_19-01-29.png) of the running time entry under Linux. This could be accomplished f.i. with `ctrl+enter`.

Ideally, the focus should then be on the `duration` field or at least on `description`. Currently, only after using `tab` once, the description field is focused.

The following `ctrl+enter` could close the details view, i.e. emulating `done`. Additionally, `ctrl+backspace` could trigger `delete`. This would allow to edit time entries conveniently, without using the mouse at all.

MartinBriza:
### 📒 Description
With this PR, it will be possible to use Ctrl+Enter to enter a running Time Entry from the title screen (timer widget).
It also changes enter behavior on the Time Entry Editor: Every element that's not a button of some sort (that means: regular comboboxes, checkboxes and actual buttons) will move the focus to the next element in chain with every Enter keypress. Ctrl+Enter on this screen will click the DONE button.

### 🕶️ Types of changes

### 🤯 List of changes
Enter on the timer widget opens a running Time Entry
Enter on the Time Entry Editor will now move focus to the next element if the current focused item is not a button. If it is a button, it will activate it.

### 👫 Relationships
Closes #2262 - except the deleting part, I consider it too easy to destroy something. I'm open to suggestions though.

### 🔎 Review hints
Check if behavior of all elements is as expected and predictable.

### DIFF
--- a/src/ui/linux/TogglDesktop/timeentryeditorwidget.cpp
+++ b/src/ui/linux/TogglDesktop/timeentryeditorwidget.cpp
@@ -38,6 +38,12 @@ projectModel(new AutocompleteListModel(this, QVector<AutocompleteView*>(), AC_PR
     ui->description->installEventFilter(this);
     ui->project->installEventFilter(this);
 
+    connect(ui->description, &AutocompleteComboBox::returnPressed, [this]() { focusNextChild(); });
+    connect(ui->project, &AutocompleteComboBox::returnPressed, [this]() { focusNextChild(); });
+
+    connect(ui->newProjectWorkspace, QOverload<int>::of(&QComboBox::activated), [this](int) { focusNextChild(); });
+    connect(ui->newProjectClient, QOverload<int>::of(&QComboBox::activated), [this](int) { focusNextChild(); });
+
     toggleNewClientMode(false);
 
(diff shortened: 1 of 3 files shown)


IndrekV:
- [ ] Enter+Ctrl opening the edit view for running entry. This should focus the description field in edit view. Right now nothing is focused.
- [ ] Enter+Ctrl to trigger `Done` button in edit view. This does not work when in description or project field. 
- [ ] I think we should add the `Delete` button shortcut also(`Ctrl+Backspace`). It's not so dangerous as we show confirmation before executing the action

The one part that is missing is jumping to the listing and moving between items and then opening edit for selected item. I'll create a separate issue for that.

IndrekV:
Issue for listing actions: #2865

PR #2831, opened 2019-02-27
New Main App design

### LINKED ISSUE #2813: New design for Main App Window
### 📒 Description
It's time to implement new design for Main App Window from Zeplin 🎨 

### 💻 Environment
Platform: macOS

### 💯 Acceptance criteria 
+ Implement respectfully the designs for Main app, which includes:
+ Overall Main app layout
+ Day expand 
+ Load More
+ Group and Ungroup layout

NghiaTranUIT:
### 📒 Decision
This PR will introduce new Main App design as well as the change from NSTableView to NSCollectionView.

### Why do I select NSCollectionView than NSTableView
After preliminary experiment, our NSTableView has many limitations which we have to overcome by doing tricks if we would like to implement 100% new design.

Here is the list of reasons:

1. Shadow underneath of cell
- I tried to support Shadow in TimeEntryCell (NSTableView), but it turns out that the shadow of one cell will overlap with next cell -> The shadows is darker in top and bottom, but less in left and right.
- NSCollectionView is capable of adding DecorationView, which is underneath of each section, so it's pretty easy to draw a shadows without messing with another section.

2. Header
- NSTableView doesn't support native Header, so currently, we have both same class (TimeEntryCellWithHeader and TimeEntryCell)
- NSCollectionView officially supports Header as a supplement view. So it's easier to implement TimeEntryHeader and TimeEntryCell separately.
- Additionally, it's easier to implement the collapsible section because TimeEntryHeader is different view.

3. TimeEntryCell
- TimeEntryCell is now separated with Header, so there is no duplicated code / design if comparing to old solution (NSTableView)

4. Section Inset 
- NSTableView is impossible to support section inset, since every row will render sequently or we have to do many `tricks`, which I'm totally discouraged.
- NSCollectionView officially support section inset from [NSCollectionViewFlowLayout](https://developer.apple.com/documentation/appkit/nscollectionviewflowlayout?language=objc).

5. Left/right top and left/right bottom corners
- NSTableView cell is super **difficult** to support Left/Right Top corner in the first cell in section and left/right bottom for the last cell in section. 
- By introducing the DecorationView, it's super easy, because we only have to round 4 corners.

6. Row / Section data structure
- If we look around the design, it easily recognize that the pattern of EntryTimeList must be section-based list.
- NSCollectionView support section by default. Otherwise, NSTableView isn't

7. Collapsible Section
- If we have TimeEntryHeader, it's straightforward to implement collapsible Section, since the header view and the cell are different.
- NSCollectionView do section very well, not NSTableView.

=> I made decision to move on our old list to NSCollectionView, so we could respect our design and easier to implement new feature.

screen_shot_2019-02-27_at_17_46_18
screen_shot_2019-02-27_at_17_46_20
screen_shot_2019-02-27_at_17_49_45

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Drop out NSTableView and adopt NSCollectionView for Time Entry List
- [x] Support TimeEntryCell, TimeEntryHeader
- [x] Implement new design for TimeEntryCell and header
- [x] Introduce Decoration View for shadows and corners.
- [x] Drag and drop
- [x] Migrate all old logic to new one 
- [x] Support Dark-mode and complex hover state for cell
- [x] Introduce TimeEntryDataSource, which is responsible for categorize the array of TimeEntryItem to section-based structure.
- [x] Remove unnecessary code from NSTableView

### Known-issue
- [x] Keydown func is conflicted with didSelectedItem (fixed with 
https://github.com/toggl/toggldesktop/pull/2831/commits/7c0f11be5ba28318f2fd46644fdd94a82c4c07a7)
- [ ] Differ - DeepDiff doesn't support section-based diff.
- [x] Image of cell during dragging doesn't work with Darkmode (Background of cell is white color)
- [x] Strange jump after closing the Popover

### 👫 Relationships
Close #2813 

### 🔎 Review hints
- Double check all logics

## Skeleton
screen_shot_2019-02-27_at_18_10_47


NghiaTranUIT:
Fixed the conflict successfully, pls make sure you guys check out the branch from origin 👍

NghiaTranUIT:
Basically, this PR is ready for reviewing (except Diff func, it should be separate ticket) 👍

IndrekV:
On first load the listing is not rendered correctly. Lines between rows are from side to side and the hover color is also larger than the list. After a reload it looks ok again.

Screenshot 2019-03-11 at 12 32 31

IndrekV:
After app start it is not possible to edit running entry.

Scenario:

- Start app
- continue an entry from past
- Try to click on an entry to edit it
- Nothing happens
- The app fails at this spot in `TimeEntryListViewController.m`
https://github.com/toggl/toggldesktop/blob/cb7a324984b9ebd1811aab13189c7df955d6b4e8/src/ui/osx/TogglDesktop/TimeEntryListViewController.m#L244
```
clickedIndexPath is nil when triggered from timer
```

IndrekV:
- [ ] When editing the date of the time entry the popup does not point to the correct time entry row after edit.
- [ ] App always shows billable icon in entry item (even when billable is false)
- [ ] When deleting the last item of the day I got error
```
Fatal error: Index out of range
```
Probably removing section is not handled correctly

in **`let sectionItem = sections[section]`**
```
extension TimeEntryDatasource: NSCollectionViewDataSource, NSCollectionViewDelegate, NSCollectionViewDelegateFlowLayout {

    func numberOfSections(in collectionView: NSCollectionView) -> Int {
        return sections.count
    }

    func collectionView(_ collectionView: NSCollectionView,
                        numberOfItemsInSection section: Int) -> Int {
        let sectionItem = sections[section]
        if sectionItem.isOpen {
            return sectionItem.entries.count
        }
        return 0
    }
```
- [ ] Unsunced icon missing from entry item
Screenshot 2019-03-11 at 13 31 50

NghiaTranUIT:
> Unsynced icon missing from entry item
This icon intended to support in "New design of System Message" (https://github.com/toggl/toggldesktop/pull/2837).

NghiaTranUIT:
> On first load the listing is not rendered correctly. Lines between rows are from side to side and the hover color is also larger than the list. After a reload it looks ok again.

I couldn't reproduce this issue, do you remember how to trigger that? I tried to open TimeEntryList after logging, start/stop (no running entry), start/stop with running entry, but I couldn't 🤔 

> After app start it is not possible to edit running entry.

This one too, it seems that both issues were gone.

Additionally, I fixed couple bugs above and leave comment about the "Error icon in Time Entry Cell".

-----
It's ready for reviewing in second round @IndrekV, please make sure remove your current branch and checkout the `feature/design-app-window` from `origin`, because I rebased in order to resolve the conflict from `master` 👍

IndrekV:
>After app start it is not possible to edit running entry.

This is still broken. I'll add gif to demonstrate the issue. When I move the mouse over the running entry when it runs I click on it multiple times to start editing.

mac_unable_to_edit_running


Scenario:

- Start app (with no running time entry)
- continue an entry from past
- Try to click on the running entry to edit it
- Nothing happens
- The app fails at this spot in `TimeEntryListViewController.m`
https://github.com/toggl/toggldesktop/blob/cb7a324984b9ebd1811aab13189c7df955d6b4e8/src/ui/osx/TogglDesktop/TimeEntryListViewController.m#L244
```
clickedIndexPath is nil when triggered from timer
```

IndrekV:
Also can we not make the login page background image be visible as the scrollbar background.
Screenshot 2019-03-13 at 12 43 39

IndrekV:
Jep the last commit fixed the edit popup opening issue. Good Job!

IndrekV:
Tested out keyboard shortcuts in the listing view and some functionalities are not working properly.
All keyboard shortcut logic in mac is described in the bottom of the article here: https://support.toggl.com/desktop-apps/toggl-desktop-for-mac

- [ ] Pressing SPACE will not start the selected entry from the list
- [ ] Moving up in entries list. When the selection goes from one day to another it will jump to the first item of the section. It should be the last item of the day to make it smooth.

IndrekV:
@NghiaTranUIT I want your feedback on https://github.com/toggl/toggldesktop/pull/2831/commits/f97f93bb89e2f7bc1cdc5eee28d98c37a7e0cb1f

I had to use a variable to store the last selected indexPath because for some reason the collectionView inside the `shouldSelectItemsAt` was always `NULL` and had no info about selected indexes.

IndrekV:
One more issue that occurred.
- [ ] When you delete an item in list the selection is not update to the previous item. The list does not have a selected item anymore. It should probably be handled in `setFirstRowAsSelected` method inside `TimeEntryCollectionView`

src/ui/osx/TogglDesktop/Feature/TimeEntry/Views/TimeEntryCollectionView.m
```diff
@@ -138,8 +138,7 @@ - (void)deleteEntry
 		{
 			if (toggl_delete_time_entry(ctx, [cell.GUID UTF8String]))
 			{
-				[self deleteItemsAtIndexPaths:[[NSSet alloc] initWithArray:@[self.latestSelectedIndexPath]]];
```
NghiaTranUIT: I deleted this crash code because when `toggl_delete_time_entry` is triggered, it will reload entire collection view with new Entry List.

The old code tries to remove the invalid `self.latestSelectedIndexPath` -> crash 💥

NghiaTranUIT:
> When you delete an item in list the selection is not update to the previous item. The list does not have a selected item anymore. It should probably be handled in setFirstRowAsSelected method inside TimeEntryCollectionView

I fixed it in 
https://github.com/toggl/toggldesktop/pull/2831/commits/0259004db2b80d6fbc2eee4345edeac168d2afc3. However, after re-selecting the previous cell, the CollectionView isReloading and lost the current selection.

IndrekV:
I like the improved solution for moving up in the list. Good job @NghiaTranUIT

IndrekV:
Opening and closing groups with keyboard (pressing LEF/RIGHT) loses selected item highlight. On both cases the selected item should remain the same as before the action.

mac_list

IndrekV:
List sections rendering issues. It's same issue I noted before https://github.com/toggl/toggldesktop/pull/2831#issuecomment-471495721

It seems that with updates to the listing structure the loading of the margins is broken.
In the gif I click on grouped items to open and close the group. When all scrollbar is shown or hidden the rendering breaks. Sometimes is rerenders and fixes the issue in few seconds. If you resize the window it also re-renders and fixes it.

mac_list_glitch

IndrekV:
Another point. The background color of selected item and the subitems in group are the same, they should differ a bit so that user understands when they are highlighting the grouped subitem. I suggest we make the subitems a bit lighter (in dark mode) than the unselected item.

NghiaTranUIT:
This 
https://github.com/toggl/toggldesktop/pull/2831/commits/478fb5270bb5fc5fa65a9e8c95a2ef9d060aa648 fix the UI issue from https://github.com/toggl/toggldesktop/pull/2831#issuecomment-473281341.

NghiaTranUIT:
@IndrekV All bugs are gone 🥇 It's ready for next review phrase 😍

When you're happy with it. Please let me know, I will squash some unnecessary commits before merging. Then I will continue to rebase next PR to make it clean and reviewable.

NghiaTranUIT:
#### Change-logs (from last review)
- Fix jump bug
- Lighter background for selected Sub-Item cell
- Keyboard func works (move up/down and delete (re-select cell after deleting))
- Scroller bar UI issue

IndrekV:
Good Job! @NghiaTranUIT

PR #2863, opened 2019-03-13
Use the absolute value of time to set confirmless delete (all)

### LINKED ISSUE #2825: Confirmless delete removes running entries instantly without confirmation
### 💻 Environment
Platform: macOS/Windows/Linux 
Toggl Version: master

### 🐞 Actual behavior
Running entries are deleted instantly

### 💯 Expected behavior
Running entries also should show confirmation when the duration is longer than 15 seconds

### 🔨 Steps to reproduce

1. Start entry
2. Open running entry edit view
3. Click Delete
4. Entry deletes instantly

### 📦 Additional info
We keep the duration of running items in negative time and need to calculate the 15-second limit differently from the stopped entries.

MartinBriza:
### 📒 Description
This changes the way confirmless delete is determined on all three platforms.
To be determined: Should we do this in the library?

### 🕶️ Types of changes

### 🤯 List of changes
Deleting currently running time entries (longer than 15s) will issue a prompt asking if the user is sure to delete them.

### 👫 Relationships
Closes #2825

### DIFF
--- a/src/ui/osx/TogglDesktop/TimeEntryViewItem.m
+++ b/src/ui/osx/TogglDesktop/TimeEntryViewItem.m
@@ -136,7 +136,7 @@
 		self.Error = nil;
 	}
 	// If duration is less than 15 seconds delete without confirmation
-	self.confirmlessDelete = (self.duration_in_seconds < 15);
+	self.confirmlessDelete = te->ConfirmlessDelete;
 
 	// Grouped mode
 
--- a/src/toggl_api.h
+++ b/src/toggl_api.h
@@ -74,6 +74,8 @@ extern "C" {
         // If syncing a time entry ended with an error,
         // the error is attached to the time entry
         char_t *Error;
+        // Short time entries can be deleted without user confirmation
+        bool_t ConfirmlessDelete;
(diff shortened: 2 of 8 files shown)


src/ui/osx/TogglDesktop/TimeEntryViewItem.m
```diff
@@ -136,7 +136,7 @@ - (void)load:(TogglTimeEntryView *)te
 		self.Error = nil;
 	}
 	// If duration is less than 15 seconds delete without confirmation
-	self.confirmlessDelete = (self.duration_in_seconds < 15);
+	self.confirmlessDelete = (self.duration_in_seconds > -15 || self.duration_in_seconds < 15);
```
NghiaTranUIT: Can you elaborate why it must be greater than `-15`?

MartinBriza: whoops, this is actually wrong :D

IndrekV:
- [x] Works on Mac
- [x] Works in Linux
- [ ] Works in Windows ( unable to test properly because breakpoints don't seem to work )

PR #2881, opened 2019-03-20
Handle adding tags inside the app (linux)

### LINKED ISSUE #1973: Linux app adding tags
Like the title says, it's annoying that native linux app doesn't provide an option to add tags to a task - only a checklist to choose from. Any plans to implement it?
BTW: great app!

### LINKED ISSUE #2193: Make it possible to create Tags within Linux Desktop app
This is currently not possible, a user submitted a Feature request to make this possible in some of the future versions:

>Is not possible to add/create tags in linux version yet and the edition panel of timers could be improved. Any plans for these changes be implemented? 

Case: #485175

MartinBriza:
### 📒 Description
A hacky, workaround-y solution that stores the newly created tags until they actually come from the server side.

This is to prevent them being overwritten (and flashing) inside the app, because after creating a new tag, the library or the backend sends multiple updates with old ones before they start to contain the newly created one.

### 🕶️ Types of changes

### 🤯 List of changes
- Adding tags is now possible inside the Linux app
- Spacing and layout inside the Linux Time Entry edit view is now more consistent

### 👫 Relationships
Closes #1973

### 🔎 Review hints
Check if tags are added correctly and the UI behaves as expected

### DIFF
--- a/src/ui/linux/TogglDesktop/timeentryeditorwidget.cpp
+++ b/src/ui/linux/TogglDesktop/timeentryeditorwidget.cpp
@@ -498,8 +498,12 @@ void TimeEntryEditorWidget::on_tags_itemClicked(QListWidgetItem *item) {
             tags.push_back(widgetItem->text());
         }
     }
+    if (item) {
+        tags.push_back(item->text());
+    }
     tags.sort();
     QString list = tags.join("\t");
+
     if (previousTagList != list) {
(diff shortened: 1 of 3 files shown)


IndrekV:
I did 2 fixes:
- If new tag was the same as one of the tags in the list checking the item in list did not work
- The link text was wrong "Add new client" should have been "Add new tag". Classic copy paste mistake.

Now it's all good, works as expected.

PR #2885, opened 2019-03-22
Fix/offline state

### LINKED ISSUE #2884: Improve offline capabilities of the apps
### 💻 Environment

Platform: All (library)

### 📒 Description
- Load more button keeps spinning (it should either not show the button offline or just hide it after the click)
- Entries do not get unsynced icons for some reason (fixed with f8a7644)
- We should validate data before sending so we can show proper error message without a network connection.
Example errors:
   - Too big duration. If the app is offline strange things starts to happen. The timer starts to count down. It happens even if you select a stopped entry and change duration to some big number (more than 999 hours that is allowed). it will start to count down.

    - Start/Stop time errors
`Stop time year must be between 2006 and 2030`

IndrekV:
### 📒 Description
App should work offline exactly the same as online. No functionality should be missing. This mostly goes to data validation and syncing errors.

### 🕶️ Types of changes

### 🤯 List of changes
- Render entries list right away after sync fails. This means that you see unsynced icons right when the sync fails
- ... To be added

### 👫 Relationships

Closes #2884

### 🔎 Review hints

- Turn off internet connection
- Do a bunch of actions. (edit entries, start, stop, delete)
- See if unsynced icons are shown properly. Or if validation errors are displayed.
- If the internet is turned off the app should show unsynced icons for entries right after app start
- The app should not allow saving duration larger than 999 hrs. (3 596 400 seconds). It makes the duration 999hrs if the value is bigger.
- Error should be shown if the selected date is not between the year 2006 and 2030.
- Error should be shown if the description text is longer than 3000 characters.

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -2897,7 +2897,7 @@ error Context::SetTimeEntryDate(
             Poco::Timestamp::fromEpochTime(te->Start()));
 
         // Validate date input
-        if (date_part.year() < kMinimumAllowedYear || date_part.year() > kMinimumAllowedYear) {
+        if (date_part.year() < kMinimumAllowedYear || date_part.year() > kMaximumAllowedYear) {
             return displayError(error(kInvalidDateError));
         }
 
@@ -2953,7 +2953,7 @@ error Context::SetTimeEntryStart(
     Poco::LocalDateTime local(Poco::Timestamp::fromEpochTime(te->Start()));
 
     // Validate time input
-    if (local.year() < kMinimumAllowedYear || local.year() > kMinimumAllowedYear) {
+    if (local.year() < kMinimumAllowedYear || local.year() > kMaximumAllowedYear) {
         return displayError(error(kInvalidStartTimeError));
     }
 
--- a/src/context.cc
+++ b/src/context.cc
@@ -4358,6 +4358,7 @@ void Context::syncerActivity() {
                 err = pushChanges(&client, &trigger_sync_);
                 trigger_push_ = false;
                 if (err != noError) {
+                    user_->ConfirmLoadedMore();
                     displayError(err);
                     return;
                 } else {
@@ -4385,6 +4386,7 @@ void Context::syncerActivity() {
 
                 error err = pushChanges(&client, &trigger_sync_);
                 if (err != noError) {
+                    user_->ConfirmLoadedMore();
                     displayError(err);
                 } else {
                     setOnline("Data pushed");
@@ -4709,6 +4711,8 @@ error Context::pushChanges(
                 api_token,
                 *toggl_client);
             if (err != noError) {
+                // Hide load more button when offline
+                user_->ConfirmLoadedMore();
                 // Reload list to show unsynced icons in items
                 UIElements render;
                 render.display_time_entries = true;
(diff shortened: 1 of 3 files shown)


src/formatter.cc
```diff
@@ -370,9 +370,12 @@ int Formatter::ParseDurationString(const std::string value) {
 
     // 15
     if (input.find(".") == std::string::npos) {
-        int minutes = 0;
-        if (Poco::NumberParser::tryParse(input, minutes)) {
-            return seconds + (minutes * 60);
+        Poco::Int64 minutes = 0;
+        if (Poco::NumberParser::tryParse64(input, minutes)) {
+            if (minutes > kMaxDurationSeconds) {
+                return kMaxDurationSeconds;
```
NghiaTranUIT: I'm suppose that If it's incorrect when we return `kMaxDurationSeconds` in `Formatter::ParseDurationString`.

Let take a following code as a typical example:

https://github.com/toggl/toggldesktop/blob/91d8d9fa02ea56ddb3fc42527a8584f2a31e808f/src/context.cc#L2790-L2793

Let imagine that in some scenario, the`Formatter::ParseDurationString` return `kMaxDurationSeconds`, but the 
```
return displayError(error(kOverMaxDurationError));
```
is never executed.

---------
I could be fixed by replacing with
```
if ((minutes * 60) > kMaxDurationSeconds) {
        return kMaxDurationSeconds + 1;
}
```

src/context.cc
```diff
@@ -2941,6 +2952,11 @@ error Context::SetTimeEntryStart(
 
     Poco::LocalDateTime local(Poco::Timestamp::fromEpochTime(te->Start()));
 
+    // Validate time input
+    if (local.year() < kMinimumAllowedYear || local.year() > kMinimumAllowedYear) {
```
NghiaTranUIT: This condition is wrong in my case when I try to edit the Start Time from `17:30` to `17:20`. Therefore, it presents invalid Year error.

Screen Shot 2019-04-02 at 10 47 49

## Attempt:

Screen Shot 2019-04-02 at 10 47 35

The `local.year()` return `106` (in console), but `local._year` is `2019`

NghiaTranUIT: For some reason, the `local.year()` returns wrong value event the TimeEntry's StartDate is correct 💭

src/context.cc
```diff
@@ -2890,6 +2896,11 @@ error Context::SetTimeEntryDate(
         Poco::LocalDateTime time_part(
             Poco::Timestamp::fromEpochTime(te->Start()));
 
+        // Validate date input
+        if (date_part.year() < kMinimumAllowedYear || date_part.year() > kMinimumAllowedYear) {
```
NghiaTranUIT: Same with my previous comment, we should double check whether `date_part.year()` returns correct value or not 🤔

src/context.cc
IndrekV: local.year actually returns the correct value. I'm not sure why console show other number. The issue was that in second check there should be `kMaximumAllowedYear`.

src/formatter.cc
IndrekV: I think the `kMaxDurationSeconds + 1` is not needed we can just use
```
if (seconds >= kMaxDurationSeconds) { 
     return displayError(error(kOverMaxDurationError)); 
 }
```
In `context.cc` and it will work. Fixed with (43093445d8947dff89c22e8e44c41685894d1b91)

IndrekV:
I've fixed all the reported issues. It's now ready for a second review.

Issue #1862, opened 2016-03-17
Clicking on the right side of the description textbox does not put cursor in the end of text


IndrekV:
> When using the desktop-based app and entering text about a task, clicking in the text box to the right of the last entered character does not automatically set the cursor to the end of the string.

If I have set the cursor to the middle of the edit vide description field then clicking on the right side will not move it to the end. Event if another field is focused and I click on the right side of description field the field is focused but the cursor is still where I last had it (in the middle of the text for example)

korg250:
Everything related with cursor position and text selection does not follow the default behaviour (like any other Windows program). Another examples:
- The cursor "remembers" its position for every single field, even if I close and open another task.
- Double click on a word, does not select the word, but the whole sentence. The same for time fields.

JamieGruener:
Came here to open this very issue. Excited to see it's already been defined, saddened to see it's been open for 2+ years.

IndrekV:
Closed with #2894 2894

PR #2894, opened 2019-03-26
Fix cursor placement issue #2097 (windows)

### LINKED ISSUE #2097: Cursor Placement Issue On Windows Desktop

skel35:
### 📒 Description
This PR fixes the issue with the TextBox when clicking to the right of the text did not change the cursor placement.

### 🕶️ Types of changes

### 👫 Relationships
Closes #2097, #1862 

### 🔎 Review hints
This change affects the base TextBox style of the application, but it is actually used only in EditView, other windows have their own styles. I did check all TextBoxes anyway, but you can also have a look just in case.

HorizontalAlignment of the ScrollViewer in TextBox's template has to be "Stretch" (default value) in order for the cursor placement to work properly. Setting HorizontalContentAlignment for TextBox still works properly, apparently it applies to a more internal element of the native template.

Overriding of cursor icon was removed so that the user is not fooled by thinking that cursor can be placed somewhere where it actually can not. This currently affects 2px of margin at the edges of the TextBox and 16px of margin to the right of the "tags" textboxes in the EditView.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/Resources/SharedStyles.xaml
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/Resources/SharedStyles.xaml
@@ -37,13 +37,11 @@
                                     BorderThickness="{TemplateBinding BorderThickness}">
 
                         <ScrollViewer 
-                                    Cursor="IBeam"
                                     Padding="{TemplateBinding Padding}"
                                     BorderThickness="0" 
                                     IsTabStop="False" 
                                     Margin="2" 
                                     Background="{x:Null}"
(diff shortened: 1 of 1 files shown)


IndrekV:
Everything seems to be working. Good Job! 👍

Issue #2860, opened 2019-03-12
Edit view redesign - General


IndrekV:
### 💻 Environment
Platform: macOS

### 📒 Description
This is the main issue for Edit view redesign on Mac. 

**Description Input** 
Description field works similarly as in old design. The difference is that we need to use NSPopover so that the autocomplete dropdown can be bigger than the edit window if needed.

**Project Input**
Described in detail in a separate issue: #2857 

**Tags Input**
Described in detail in a separate issue: #2851 

**Time Input**
Described in detail in a separate issue: #2858

**Date Input (Calendar)**
Described in detail in a separate issue: #2859

**Billable Input**
Billable is a simple checkbox flag.

**Delete Button**
Delete button is a simple clickable text label

Issue #2859, opened 2019-03-12
Edit view redesign - Date input (Calendar)


IndrekV:
### 💻 Environment
Platform: macOS

### 📒 Description
The calendar has 4 different components:
- Previous day button
- Text input
- Next day button
- NSPopover with date selection

#### Functionalities
- Pressing `Previous Button` will switch to previous day
- Pressing `Next Button` will switch to next day
- The text input is editable (we have to decide what format it accepts. Maybe look at the current solution and build on top of that)
- **Popover**
* Scrolling vertically will show more dates
* Current month dates have darker color, next/previous months are lighter
* Every first day of the month is shown with month name

Issue #2858, opened 2019-03-12
Edit view redesign - Time input 


IndrekV:
### 💻 Environment
Platform: macOS

### 📒 Description
Time input has 3 component:
- Duration input
- Start -> Stop label
- Time editing NSPopover

#### Time editing flow

When duration, start or stop field is focused the popover is shown.

Depending on what was clicked the correct field is focused inside the NSPopover.

Time edit Popover has 4 parts:
- Duration field
- Start time field
- Stop time field
- Time selector

Each of the fields should be able to define which of the time slots is selected (hours or minutes).
User should be able to edit the fields manually.

**Time selector**

Time selector has two different number states: Hours and minutes.
When hours display is visible and user selects the number it switches to minutes display.

When hovering the numbers the numbers show darker gray circle to indicate hovering
When user click the position of the green circle changes and the input data updates. Same is when user keeps mouse button pressed and moves around the numbers display. (it always moves the green circle and updates the field)

IndrekV:
When starting this we understood that the implementation of the custom round time input is very complex and needs big custom component development we decided to skip this part in the current project and maybe pick it up some time in the future. Great work on this @NghiaTranUIT !

Issue #2857, opened 2019-03-12
Edit view redesign - Project input


IndrekV:
### 💻 Environment
Platform: macOS

### 📒 Description
Projects input has 4 different components:
- Text input
- Autocomplete (NSPopover)
- Add Project Form
- Project color select Form
- Add Client Form

**Text Input**
It's an input field with a down arrow button on the right side. Clicking on the down arrow should open the autocomplete list (similar to current implementation)

**Autocomplete**
Typing text in the input will filter the autocomplete list. The `+ Create a new Project` row is always visible in the bottom of the autocomplete. It is not part of the items list and does not scroll.

#### Add Project Flow
When the user presses the `+ Create a new Project` with mouse or activates it with the keyboard the contents of the NSPopover are replaced with the `Add Project Form` content. The text that was typed into the input field is copied to the project name field in the `New Project` form.

**New Project Form**
New Project has the following components:

- Cancel Button - Clears the "New Project Form" and goes back to initial autocomplete listing
- Add Button - triggers adding the new project (validation and saving the project). Then selects the newly created project for the time entry, clears the "New Project Form" and closes popup.
- Project name Input - value is initially copied from the autocomplete input field. User can edit is as they please.
- Project color dot - described in detail in: #2856
- Workspace Select - It's a regular select box without autocomplete. (Similar to current solution)
- Client autocomplete - Autocomplete of client names. If the inserted text has no matches text `No matching items`

**New Client Flow**
If user types in client name that does not have exact match a new line with text `+Create Client "Client Name"` will appear. If the client name is too long it shows `...` at the end of the name.
Clicking this new row will create client and nothing else.

PR #2906, opened 2019-04-02
Time Entry editor

NghiaTranUIT:
### Description
This PR is a combination of many sub-tickets (See change-logs). Each tickets should be reviewed and approved separately before reviewing in the final round 🚀 

### 🤯 List of changes
It's a big PR for Time Entry Editor. It's also a parent for many following sub-tickets:
- [x] General logic for Popover #2860 
- [x] Project Creation: Generic AutoComplete, AutoComplete for Project, Client, Workspace, basic Color Picker and all relevant logic to fetch them from Library (in-progress) #2857 
- [x] Tag input
- ~[ ] Time Input #2858~ We decided to postpone this ticket until we have the consensus regarding the Clock View 
- [x] Date input #2859 
- [x] Project Color Picker #2856 

### 🕶️ Types of changes

### 👫 Relationships
Project (https://github.com/toggl/toggldesktop/projects/1)

### 🔎 Review hints
- All logic is remains
- Double check the designs


NghiaTranUIT:
Rebased 👍

NghiaTranUIT:
@IndrekV @MartinBriza All sub-tickets are already approved and merged appropriately. 
So, This huge Editor PR is already for reviewing in the final round 👏

IndrekV:
- [ ]  Description field autocomplete does not show any suggestions.
- [ ]  Billable flag checkbox should be visible only when the time entry is in a non-free workspace
- [ ]  Start and end times are cut off. We need to make the edit popup a bit wider to accommodate the values in these fields.
- [ ]  The arrow between the start and stop time should not be focusable with tab.
- [ ]  Editing start or stop time and pressing tab to move to the next field does not update other time fields. Editing duration and pressing tab will update stop time accordingly.

**Additional screenshots**
Empty autocomplete in edit view:
mac-ac-empty

Start and stop time cut off:
cutoff_fields

NghiaTranUIT:
@IndrekV, I've updated new logic for Description AutoComplete as well as resolved few issues, which you mention in last comment 👍 



## Description AutoComplete's capabilities
- Able to Enter to update new Description
- Tab to move to next field & update description
- Search & auto complete
- Select existing Description from keyboard, mouse -> Update Description and Project if need
- Layout for Description Cell: Description Lbl + Color Dot View + Project or Client label

IndrekV:
I added some fixes:
- fixed the start stop field width issue without the render triggering. This render triggering cause the size of the popup to change when the width of fields changed because of the content. New solution sets a fixed with that can always fit the correct format of the time.
- Added "refuse first responder" to the arrow between time fields so it doesn't get focused ever.

All other issues seem to be fixed.

Now for some more tweaks before we can merge this in:

- [ ] Calendar seems slow somehow. There seems to be a delay opening the calendar popup. Also switching days seems to take 1 second to execute.
- [ ] Calendar hover color is gray but should be light green (see Zeplin for the color)
- [ ] Calendar selected date color text should be white not black

Screenshot 2019-05-17 at 11 56 30

- [ ] When you filter tags and then select one of them the list reloads and discards the filter. The list should not reload and should keep the same filtered state after selection.

mac-tags


If these issues are fixed we need to rebase against master and we can go on with merging this in and preparing for the grand release. Good job Nghia with all the implementations!

NghiaTranUIT:
The investigation on unresponsive Calendar when selecting new date, reveals that this following code causes the problem:

https://github.com/toggl/toggldesktop/blob/60a279c33f4e8d74f52153c398147ccf33eff460/src/ui/osx/TogglDesktop/Feature/Calendar/CalendarDataSource.swift#L76

Basically, I calculate the start and end date from the current date, then rendering the Calendar. Each cell will calculate the date by adding the index -> This action burdens the CPU -> 1s delay

I will come up alternative idea to resolve it.

NghiaTranUIT:
@IndrekV this PR is ready for reviewing 👍. I did a merge because I was unable to resolve the conflict when rebasing (git rebase -i master) the huge number of commits in this branch.

## What's news
- Re-implement the Calendar logic to render appropriately without using NSCalendar too much. In the original approach, every time the cell is rendered, it will convert the NSIndexPath to date by using NSCalendar -> It means that it burdens the CPU.

## Better approach
Here is the idea of new approach:
1. Give selected date -> Calculate the first day of week (replies on the First Day of Week in System Preference)
2. Get the `from` date by moving by offset weeks
3. Get the `to` date by adding offset weeks
4. Get the number of day for all months from `from` -> `to` (Use NSCalendar to get correctly since the number of day relies on what kind of years, ....)
4. Calculate all Date struct (consist of day, month, year: Int) by adding 1 day from `from` -> `to` => By this solution, I don't use NSCalendar to construct the NSDate, because it's inefficient-process
5. Flat the map 
6. Render on the Collection View

## Result
- Scrolling is better, no lagging like previous version
- Frozen 1 seconds is gone, however, it isn't disappear completely. Every time we update the Date of Time Entry, there are heavy things are executed in main Thread (Store on db, logic in Library, Reload the TIme Entry cell, calculate the Position of Popover, ...) -> Still lag, but acceptable

IndrekV:
All the issues pointed out are addressed and now work correctly. Good job @NghiaTranUIT !

2 last issues I noticed that are different in old design
When editing a running entry:
- Stop time input should be hidden (as it can not be edited)
- Calendar should be disabled as you can track only right now not in the past or the future

IndrekV:
Ok everything seems in order. Drum roll please..... It's time to merge this bad boy in!

PR #2854, opened 2019-03-12
Fix Focus Jump when editing Time Entry

NghiaTranUIT:
### 📒 Description
During editing the Time Entry, the focus always is lost and focus on the TimerBar Input.

### Problem
- We accidentally focus on TimerBar many time when the TimeEntryList is rendering => During editing the Time Entry, the focus will be lost if the TimeEntryList is executed.

### 💯 Expected behavior
- [x] During editing the Time Entry, make sure **100%** that the Focus on Edit View shouldn't be lost.
- [x] After closing the Edit Popover, we should focus on Timer Bar
- [x] During editing the Time Entry, the focus is still remaining even we click "Load More" (In current build, the focus is lost)
- [x] After switching Manual / Timer Mode, the Timer Bar should be focus on AutoCompleteInput.

### 🕶️ Types of changes
 **Bug fix** (non-breaking change which fixes an issue) 

### 🤯 List of changes
- [x] Remove focus TimerBar during TimeEntyList displays
- [x] Introduce EditorPopover and provide func to close and focus on Timerbar if need
- [x] Work on macOS
- [x] Work on Window (WIP 🚧 )

### 👫 Relationships
Closes #2833 

### 🔎 Review hints
- Check the Expected behavior 
- When editing the Time Entry in Popover -> 100% not lost focus.
- No jump anymore.

PR #2911, opened 2019-04-09
Focus jump on Time Editor if there is any running entry

NghiaTranUIT:
### 📒 Description
- Continue from #2854 PR, it's completely fixed for Focus Jump whether there is any running jump.

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Don't override the TimerMode when the running entry is running. If there is any running entry, the Desktop Library constantly posts `kDisplayTimerState`, and it accidentally override the TimerMode Setting. Thus, it focus to the Timer Bar again -> Lost jump focus if we're editing
- [x] Remove focus jump when selecting different TimeEntry if the Editor is opening.

### 👫 Relationships
Related to PR #2854

### 🔎 Review hints
- There is no jump any more in 2 scenarios: No running entry and running entry.

### DIFF
--- a/src/ui/osx/TogglDesktop/test2/AppDelegate.m
+++ b/src/ui/osx/TogglDesktop/test2/AppDelegate.m
@@ -1103,8 +1103,6 @@ BOOL onTop = NO;
 - (IBAction)onShowMenuItem:(id)sender
 {
 	[self.mainWindowController showWindow:self];
-	[[NSNotificationCenter defaultCenter] postNotificationOnMainThread:kFocusTimer
-																object:nil];
 	[NSApp activateIgnoringOtherApps:YES];
 }
 
--- a/src/ui/osx/TogglDesktop/test2/AppDelegate.m
+++ b/src/ui/osx/TogglDesktop/test2/AppDelegate.m
@@ -1103,6 +1103,8 @@ BOOL onTop = NO;
 - (IBAction)onShowMenuItem:(id)sender
 {
 	[self.mainWindowController showWindow:self];
+	[[NSNotificationCenter defaultCenter] postNotificationOnMainThread:kFocusTimer
+																object:nil];
 	[NSApp activateIgnoringOtherApps:YES];
 }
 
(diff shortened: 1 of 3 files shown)


NghiaTranUIT:
If you're happy with this fix, I suppose that we can prioritize it and release in BETA soon (only this ticket). There are many users, who are using Stable and Beta are encountering this annoying bug 🐞

src/ui/osx/TogglDesktop/test2/AppDelegate.m
```diff
@@ -1103,8 +1103,6 @@ - (IBAction)onViewChangelogMenuItem:(id)sender
 - (IBAction)onShowMenuItem:(id)sender
 {
 	[self.mainWindowController showWindow:self];
-	[[NSNotificationCenter defaultCenter] postNotificationOnMainThread:kFocusTimer
```
NghiaTranUIT: I deleted this code since it's **unnecessary** to focus to Timer whenever we open the Time Entry Edit View.

I don't know what is the decision behind the scene, but if this func is executed whenever I click on TimeEntry in TimeEntryList in order to open the Editor.

As a result, sometime, it focus (NSWindow.makeFirstResponder(_)) on Timer Bar, not the Description Text Field in Editor.

Hence, I removed it.

IndrekV: The main function of `onShowMenuItem` is to show app window and at that point it has to focus the timer. Maybe it's wrong to call `onShowMenuItem` from the flow of selecting time entry for editing.

NghiaTranUIT: I revert last commit and do a fix in library at https://github.com/toggl/toggldesktop/pull/2911/commits/ab70ddc51b7c36f86a3d1ef434eb3b6bb062575b

I means that we don't need to call `DisplayApp()` if we just open the Editor. I just tested it. The output is retained. Please let me know your thought @IndrekV 🏄‍♂️

PR #2907, opened 2019-04-02
Project Input

NghiaTranUIT:
### 📒 Description
It's a huge ticket, which is trying to introduce the complete new logic for Project Input.

### 🕶️ Types of changes
- **New feature** (non-breaking change which adds functionality)
- **Breaking change** (fix or feature that would cause existing functionality to change)

### 🤯 List of changes
- [x] Reusable AutoCompleteView, AutoCompleteDataSource, AutoCompleteTextField
- [x] Menu Windows: Able to expand cross the main Windows
- [x] AutoComplete logic for Project, Client and Workspace
- [x] Basic layout and logic for Color Picker
- [x] Fill TimeEntry data from selected TimeEntry from the list
- [x] Logic for create new Client and Project from Library (The UI and basic logic are done. It's time to call func from library to make it works)
- [x] Finalize the final design: We have few minor issues. Such as "Borderless Windows", "Shadow on windows", ...
- [x] Save Time Entry Name after editing
### 👫 Relationships
Close #2857 

### 🔎 Review hints
- Main logic of Project Input should work

PR #2913, opened 2019-04-12
Tag Input

### LINKED ISSUE #2851: Edit view redesign - Tags input
### 💻 Environment
Platform: macOS

### 📒 Description
Tag input has 3 components:
- Tags input field
- The list of already selected tags
- Tags autocomplete dropdown (popover)

**Display**
If no tags are selected - Show only tags input field with placeholder "Add tags"
if there are tags selected - Show as many tags as fits in a row. The ones that do not fit are displayed as label with "..."

**Opening Autocomplete**
When there are tags already selected for entry then clicking on these tags will open the autocomplete. If no tags are selected tags input is visible and just focusing the tags input field will open the tags autocomplete

When autocomplete is opened it is on top of the selected tag labels. We will not do the "removing form cross" for the tag elements.

When autocomplete is opened the selected tags are always listed at the top so it's easy to remove them if needed.

**Adding new tags**
If you type in tag name that does not have an exact match a new line is added to the bottom of the list with the text `+ Create new tag "tag name"`

**Keyboard**
It should be possible to do all the actions using only keyboard:
- select/deselect item from row
- Create a new tag

It makes sense to use similar logic as in autocomplete. (up/down keys do actions in listing and other actions are captured in input)

NghiaTranUIT:
### ⚠️ Warning
~This PR is branched off from "Project Input" (https://github.com/toggl/toggldesktop/pull/2907), so it's critical to merge it before reviewing this PR.~

~I will in charge of doing it~

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Implemented new design for Tag View
- [x] Implement new design and logic for Tag Token (Including hover, X button and gradient)
- [x] Quick hack 💥 to support TagTextField inside the "TagAutoCompleteView" (See next section for the detail)
- [x] Logic for fetching Tags from Library
- [x] Logic for Tag DataSource (Including the Tag Selection (Select/De-select))
- [x] Logic for creating new tag (but I have a on-going bug, which is the Tag in Library doesn't reflect correctly) (see the Hack 3 section below)
- [x] Unable to Tab to focus on Create new tag  (see the Hack 2 section below)
- [x] Logic for replacing long tags to "..." tag

### 👫 Relationships
Close #2851 

#### 1. Move TagTextField to AutoCompleteWindow
Technically, there is a hidden Tag TextField in EditorController. Then it is “moved” deliberately to AutoCompleteWindows after presenting and move back after the windows is dismissed. By doing this hack, the original logic for presenting the AutoComplete is retained, and the state of TagSearchTextField is retained too. At the first experiment, I try to “clone” itself, but by this approach, more boilerplate codes.
#### 2. Tab to focus on Tag Create Button doesn't work
It turns out that when pressing the Tab button, the TagAutoCompleteView is collapsed unexpectedly, so it's impossible to focus on Tag Create Button.
The hack is simply ignore the TextFieldDidEndEditing in TagTextField, so the Windows is still remained (https://github.com/toggl/toggldesktop/pull/2913/commits/2b05a2768bf8cbe9139128a9348d197fbe1190fe)
#### 3. New tags doesn't appear after creating
After creating new Tag, for some reasons, the Library doesn't return any new tags (https://github.com/toggl/toggldesktop/blob/ca782683aabc180ab3332473f2e33406445461ca/src/ui/osx/TogglDesktop/test2/AppDelegate.m#L1631) (even the Syncer is done). Until I have to close the TimeEntry, then trigger the `on_tag` again.
The hack is just simply keeping all newTags, and appending it if need.

### 🔎 Review hints
- Able to create new tag
- Able to select / de-select tag and the UI is corresponding properly 
- Able to filter tag
- Same Design


NghiaTranUIT:
Updated ### Hack section. It's ready to review @IndrekV

PR #2950, opened 2019-05-10
Incorrect selected on the first Time Entry

### LINKED ISSUE #2933: Mac app shifts focus to edit the previous time entry instead of the selected one
### 💻 Environment

Platform:  macOS
OS Version:  10.14.3
Toggl Version:  7.4.373

### 🐞 Actual behavior

Clicking on the newly saved entry to edit it, the popup opens the edit window for the previous entry instead

### 💯 Expected behavior

The entry user wants to edit should be edited and not the previous one which hasn't been selected.

### 🔨 Steps to reproduce

1. Start the timer and then stop it after several seconds
2. Click to edit it, you can already see that the previous TE has been opened
3. Any changes would apply to the previous time entry and not the one user intended to

### 📦 Additional info

Screencast of the issue is in the [case](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/2759100/conversations/21767864701)

NghiaTranUIT:
### 📒 Description
There are many [reports](https://github.com/toggl/toggldesktop/issues/2933) that the users couldn't select the first Time Entry after using the app awhile.

This bug isn't 100% reproducible, it took me couple minutes to reproduce it by playing around the app (create new Time Entry, select different row, open/close popover, ...)

### Problem
After carefully investigation, it turns out that this following logic will return wrong cell. 
Although the `row = 0`, but the `RowView` isn't the first row. It's the second row during debugging.
https://github.com/toggl/toggldesktop/blob/9077fd0aa5ef2a9ed464cd52aae2fd5b81f38ec9/src/ui/osx/TogglDesktop/TimeEntryListViewController.m#L523-L524

I believe that this logic isn't the good way to get the cellRow from given row index. 

### Solution
By using `-[NSTableView viewAtColumn:row:makeIfNecessary]`, we can get the cell directly.
```objc
NSView *cellView = [self.timeEntriesTableView viewAtColumn:0 row:row makeIfNecessary:YES];
```
=> The bug is gone.

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Replace the `[NSTableView rowViewAtRow:]` with `-[NSTableView viewAtColumn:row:makeIfNecessary]`

### 👫 Relationships
Closes #2933 

### 🔎 Review hints
- Select first Time Entry should open the Time Popover correctly.

### Warning
In the master, we don't use old `TimeEntryListViewController.m` anymore, so this bug doesn't affect on the version in `master`, which has new Time Entry design.

So, if we gonna release the hot-fix for `v.7.4.373`, we could merge this ticket. Otherwise, we can close it, since it's unnecessary @IndrekV

### DIFF
--- a/src/ui/osx/TogglDesktop/TimeEntryListViewController.m
+++ b/src/ui/osx/TogglDesktop/TimeEntryListViewController.m
@@ -520,24 +520,21 @@ extern void *ctx;
 		return nil;
 	}
 
-	NSView *latestView = [self.timeEntriesTableView rowViewAtRow:row
-												 makeIfNecessary  :YES];
+	NSView *cellView = [self.timeEntriesTableView viewAtColumn:0 row:row makeIfNecessary:YES];
 
-	if (latestView == nil)
+	if (cellView == nil)
 	{
 		return nil;
 	}
 
 	self.selectedEntryCell = nil;
 
-	for (NSView *subview in [latestView subviews])
+	if ([cellView isKindOfClass:[TimeEntryCell class]] || [cellView isKindOfClass:[TimeEntryCellWithHeader class]])
 	{
-		if ([subview isKindOfClass:[TimeEntryCell class]] || [subview isKindOfClass:[TimeEntryCellWithHeader class]])
-		{
-			self.selectedEntryCell = (TimeEntryCell *)subview;
-			return self.selectedEntryCell;
-		}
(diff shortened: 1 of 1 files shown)


IndrekV:
Let's push forward with new design and not do a hotfix.

Stosiu:
@IndrekV When can we expect a version with a new design? This bug is super annoying.

Issue #2589, opened 2018-10-10
TE description drop-down matching suggestions list reloading issue


VeljkoTN:
User: 1464634
User agent: windows_native_app/7.4.252-obm-87-obm-98-obm-99-obm-101
Version: 7.4.252 
UpdateChannel: stable
Intercom conversation: conversations/18940031416

The user reported that when they first start typing an entry description the list of drop-down suggestions is accurate at first, but then after a couple of seconds the list is not filtered and all time entries show up as suggestions.

I was not able to reproduce this, clearing local data and a fresh install of the latest stable version did not help

skel35:
Duplicates #2839

skel35:
Fixed by #2960

PR #2960, opened 2019-05-16
Filter autocomplete based on entered text when setting controller if …

### LINKED ISSUE #2839: Time entry description autocomplete issue - Windows - 7.4.345
### 💻 Environment

Platform: Windows
OS Version: Windows 10
Toggl Version: 7.4.345 

### 🐞 Actual behavior
The user shared a very detailed description of the issue, I'll summarize here but you can get a lot more info and a video of the issue being reproduced in the case - **[link 🔗](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/unassigned/conversations/20977139173)** 

When you start typing a time entry description, initially correct autocomplete recommendations show up, if you stop of a second or two, the list of recommendations 'refreshes' itself with suggestions that have no connection with what was initially typed in. 

I tested this out on my end and was able to reproduce the same. This only seems to be happening if you start creating a new TE right after stopping one - if you allow some time to pas (like 5-10 sec for the sync to be over) everything works as expected. 

### 💯 Expected behavior
The list shouldn't refresh if you stop typing with the previous TE still in the process of being synced (if that's the actual reason here)

### 🔨 Steps to reproduce

1. Stop a running time entry
2. Immediately start typing a description that was already used before, type in a few letters and wait
3. The autocomplete suggestions list will be correct/relevant at first and will refresh when you stop typing and give it a sec

### 📦 Additional info
A video of the issue recorded by the user in in the case linked above

skel35:
…autocomplete popup is currently open #2839 (win)

### 📒 Description
When autocomplete is open and the library sends UI update request it resets the controller for auto completion popup, which makes the popup show entire updated list without applied filtering.

The fix is to apply to the list coming from UI update request the same filtering that is already applied to the existing autocompletion popup.

### 🕶️ Types of changes

### 👫 Relationships
Closes #2839, #2589.

### 🔎 Review hints

Steps to reproduce are in https://github.com/toggl/toggldesktop/issues/2839 (may not reproduce every time, have to click really fast)

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/controls/AutoCompletionPopup.xaml.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/controls/AutoCompletionPopup.xaml.cs
@@ -202,7 +202,7 @@ namespace TogglDesktop
             this.controller = controller;
             this.needsToRefreshList = true;
             if (this.popup.IsOpen)
-                this.ensureList();
+                this.open(true);
         }
 
         public void OpenAndShowAll()


IndrekV:
Well done! The flickering and loading of full list is gone.

PR #2964, opened 2019-05-17
Disable autocomplete in Description if the timer is running

### LINKED ISSUE #2957: When timer is running the timer autocomplete should be disabled
### 💻 Environment
Platform: macOS
Toggl Version: master

### 🐞 Actual behavior
When timer is running user is able to focus the description field and start autocompleting. This is very confusing and it is actually not possible to close the autocomplete listing when it is opened in that state. Selecting with Enter key press will stop the entry

### 💯 Expected behavior
When the timer is started all the fields in timer are disabled and all editing at this point should be done through edit popup.

### 🔨 Steps to reproduce

1. autocomplete in timer and select only project
2. start timer
3. try to click on the description field and see that it is focusable and autocomplete works. Also, you are not able to close autocomplete without making a selection.

### 📦 Additional info
timer_decription_focus

NghiaTranUIT:
### 📒 Description
There is a confusing behavior that the AutoComplete appears during the Timer is active

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Disable editable capability of Description AutoComplete

### 👫 Relationships
Close #2957 

### 🔎 Review hints
- The autocomplete won't show if the timer is running 
- The Editor popover presents -> Able to type the name of Time Entry

### DIFF
--- a/src/ui/osx/TogglDesktop/TimerEditViewController.m
+++ b/src/ui/osx/TogglDesktop/TimerEditViewController.m
@@ -424,7 +424,6 @@ NSString *kInactiveTimerColor = @"#999999";
 	self.autoCompleteInput.stringValue = self.time_entry.Description;
 
 	// Make descriptionLabel is editable and focus
-	self.descriptionLabel.editable = YES;
 	self.descriptionLabel.stringValue = self.time_entry.Description;
 	self.displayMode = DisplayModeTimer;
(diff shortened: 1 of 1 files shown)


IndrekV:
This didn't work correctly. The issue was that the description label should be editable when you only select project from autocomplete and the timer is not running. I've made the adjustment to make it work correctly.

NghiaTranUIT:
Thank you for the fix 👍

Issue #2968, opened 2019-05-20
Windows app hanging when changing date to future date


skel35:
### 💻 Environment
Platform: Windows
Toggl Version: 7.4.345

### 🐞 Actual behavior
Application hangs.

### 💯 Expected behavior
Application should handle the use case without hanging or crashing.

### 🔨 Steps to reproduce

1. Start the timer
2. Set the date to a future date (e.g. 2 days after today)
3. Press Stop (if warning about future date appears, ignore or close it)

skel35:
When the issue #2523 was addressed the `SetTimeEntryStart` library method was fixed, but the `SetTimeEntryDate` method is still not handling the future dates. (Not saying it should be handled in the lib, might be easier to do it on a UI side, dunno yet).

skel35:
This specific bug will not be relevant after #2974 is done, but there is a similar issue.
When setting date to a future date on a stopped time entry, the entry just gets deleted automatically during the next sync without any notification or feedback to the user.

IndrekV:
Please create a new issue for the description in your last comment so we can look into this a bit more.

skel35:
Created #2976.
Closing this one can be closed as it will be handled in scope of #2974.

PR #2969, opened 2019-05-20
Fix UnixFromDateTime method to properly account for time zone and day…

### LINKED ISSUE #2012: Selecting date in Australian timezone saves the date of the day before
User reports:

> I selected 28 September 2016 in the date picker. This makes the date picker disappear, but then the date displayed in the date field is set to 27/09/2016.

case 358926

### LINKED ISSUE #2160: Desktop timezone issue
A few reports or the desktop app not choosing the correct time or date in app but appearing correct in browser.

> Switzerland we add 1 hour. But on Windows 10, this is considered as UTC;1 not UTC;2. On Toggl.com it;s considered as UTC;2.
> 
> The result is that in the desktop app, I create an entry at 5PM (UTC), it;s marked 6PM when it;s actually 7PM but on Toggl.com it;s correctly set as 7PM.

/case/450995

### LINKED ISSUE #2455: Windows Desktop App Editing Time Entry - Date Issue
A user has reported that when they edit time entries through the desktop app and change the date on an entry, the entry takes on the date of the day before the actual date. 

When the issue was reported the user was on Win 7, in the process of our correspondence they upgraded the laptop and the OS to Win 10, they have a fresh install and the same thing happens. 

We've checked with the user if their Toggl timezone setting matches their device timezone - the user confirms that it does. 

Backoffice shows that the users' latest known location was "Dallas, Texas, US" and that their time zone is "Australia/Adelaide" - I've asked about this, the user travelled to Dallas earlier this year but is now in Australia, not sure if this can be the cause of the issue (latest known location not updating in our system to his current location when he traveled back to Australia). 

Case: 672501

I've asked the user to select a different timezone, save the change, then change back to the correct timezone and save again to see if that will somehow affect things, but no result. 

Version: [Latest Stable, released at (05-03-2018)]  7.4.122

Let me know if you need anything else.

### LINKED ISSUE #2561: Timezone issue with the Windows Desktop app 7.4.252
A user and 2 of his colleagues located in New Zealand with the correct/matching time zone settings between Toggl and their devices are having the following issue: 

Time Zone set to Auckland, Wellington (UTC + 12)

1. Enter text for new task
2. Press Enter to start the clock
3. Click on entry to bring up details.
4. Click on the calendar icon to bring up the calendar
5. Select a date
 - At this point the date changes but to the day before the date I selected. E.g. I selected 3rd September and the 2nd gets selected.
- The entry gets moved to the 2nd and the elapsed time gets changed to 48 hours. Obviously, if I selected an earlier date it would be higher.

All of them are on Win 10

Clearing the apps local data manually and installing the latest stable version did not help, their previous version was 7.4.237 and they were having the same issue.

### LINKED ISSUE #2936: When Editing, Date Field Reverts To Previous Day
Nghia believes there was previously as issue closed related to this but was unable to find it and asked me to create a new one.

### 💻 Environment

Platform:  Windows

OS Version: 

windows_native_app/7.4.345-obm-87-obm-98-obm-99-obm-101 
Toggl Version:  Version: 7.4.345

### 🐞 Actual behavior

The User writes; 

"Start timer when it is PM, Stop, Edit that entry and set the start time to any time that is before 12:00 like 10:00, Notice that the Date field is automatically set to yesterday and you can&#39;t manipulate it to today again."

### 💯 Expected behavior

Date field reflects current date.

### 📦 Additional info

Case: /conversation/21775004221

skel35:
…light savings data (win)

### 📒 Description
`UnixFromDateTime` method was not working correctly when there was a different local time zone offset between 1st of January and the date of current time entry.
Current `UnixEpoch` was converted to local time and subtracted from local `DateTime`. This is a logical error, as there is no guarantee that `UnixEpoch` local time conversion produces the same time zone offset as local `DateTime` currently has. The subtraction should be performed in UTC.

The .NET `DateTimeOffset` actually represents a `DateTime` + offset, not just some offset as its name could mistakenly indicate. It is somewhat analogous to Poco's `LocalDateTime`.

### 🕶️ Types of changes

### 👫 Relationships
Closes #2936
Closes #2561
Closes #2455
Closes #2012
Closes #2160

### 🔎 Review hints
STR described in #2936.
Try different time zones, with daylight savings on and off.
I have attempted to thoroughly test the changed code before PR, but ofc I could miss some edge case.
Known bug related to future dates: #2968.
Note: When changing time zones during the time when Visual Studio is open, building stops working, even when no code has changed and you press F5. So restarting Visual Studio for every time zone change is recommended.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/Toggl.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/Toggl.cs
@@ -19,9 +19,9 @@ public static partial class Toggl
     public const string Description = "description";
 
     public const string TagSeparator = "\t";
-
-    private static readonly DateTime UnixEpoch =
-        new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
+    
+    private static readonly DateTimeOffset UnixEpoch = 
+        new DateTimeOffset(1970, 1, 1, 0, 0, 0, TimeSpan.Zero);
 
     private static IntPtr ctx = IntPtr.Zero;
 
@@ -1366,12 +1366,12 @@ public static partial class Toggl
 
     public static DateTime DateTimeFromUnix(UInt64 unix_seconds)
     {
-        return UnixEpoch.AddSeconds(unix_seconds).ToLocalTime();
+        return UnixEpoch.AddSeconds(unix_seconds).ToLocalTime().DateTime;
     }
 
     public static Int64 UnixFromDateTime(DateTime value)
     {
-        TimeSpan span = (value - UnixEpoch.ToLocalTime());
+        var span = new DateTimeOffset(value) - UnixEpoch;
         return (Int64)span.TotalSeconds;
     }
 


IndrekV:
Tested out 3 scenarios that were described in the issues and the issues did not appear. Looks good to me. 👍

nshtinkov:
Still have 1h offset of Toggl desktop app (Windows) from system time. System time zone is "(UTC-05:00) Eastern Time (US & Canada)" with "Adjust for daylight saving time automatically" ON. Time zone in Toggl account is "America/New York (UTC-4)". This is the same time zone and should show the same time.
Screenshot (30)

skel35:
Hi @nshtinkov, thanks for reporting the issue. We'll test out this scenario and get back to you.
Meanwhile, make sure you're running the latest stable version of the app which is 7.4.475 currently https://toggl.github.io/toggldesktop/.

nshtinkov:
Thanks @skel35 for the prompt reply; version 7.4.475 displays correct times so the issue seems fixed. I had the older version (7.4.422) because that's what I get from the download link on https://toggl.com/toggl-desktop/

skel35:
@nshtinkov Glad to know that the issue is fixed.
You're correct, we have indeed not updated the link on that page yet, will do that soon. 
Feel free to reach out to support@toggl.com if you happen to have any other issue. Have a nice day!

PR #2975, opened 2019-05-23
Disable date picker for the currently running time entry #2974 (win)

### LINKED ISSUE #2974: Calendar disabled when timer running
### 💻 Environment

Platform: Windows

### 📒 Description
On mac we currently have the calendar disabled when the timer is running. We should add this functionality to Windows also as editing date on a running timer can cause issues. For example, when time is set to future the error appears about "Stop time needs to be later than start time"

Also, the fact that tracking time should be a current event not in the far future.

If the user needs to have the start date set to yesterday they can do so with duration.

Things to do:

- When editing running timer -> Calendar input disabled
- When editing stopped timer -> Calendar input enabled.

skel35:
### 📒 Description
Disable date picker for the currently running time entry
Hide the calendar button when it is disabled 
Gray out the date when date picker is disabled

### 🕶️ Types of changes

### 👫 Relationships
Closes #2974


IndrekV:
👍

Issue #2958, opened 2019-05-16
Update bugsnag version and data


IndrekV:
### 💻 Environment
Platform: macOS/Windows/Linux

### 📒 Description
We should update our bugsnag versions (At least on windows it's old)

Also we should make sure we send all additional data in correct format so we can search the error logs properly.

We should also update the code so we only send bugnsnag notifications when in production to nowt crowd bugsnag with development exception reports.

skel35:
Also there is a problem that Windows 10 and Windows 8.1 are detected as Windows 8 in Bugsnag. In order for `Environment.OSVersion` to return a correct version, this version has to be enabled in the application manifest https://docs.microsoft.com/en-us/windows/desktop/SysInfo/targeting-your-application-at-windows-8-1.

IndrekV:
Latest NuGet package for windows app.
https://www.nuget.org/packages/Bugsnag/

IndrekV:
Sending user ID in #C is described in Callback section https://github.com/bugsnag/bugsnag-dotnet#callbacks

IndrekV:
Windows and mac are done. We now need to double check the linux implementation. Linux uses custom made solution. What we need to check is if Linux sends the userID correctly.

IndrekV:
This is done now, Linux PR is also already merged.

PR #2978, opened 2019-05-23
Update Bugsnag to 2.2.0 in TogglDesktopUpdater

skel35:
### 📒 Description
Update Bugsnag to 2.2.0 in TogglDesktopUpdater
Update year in AssemblyInfo

### 🕶️ Types of changes

### 👫 Relationships
Related #2958


skel35:
Unfortunately this won't help to fix the problem. Updater is written in a way that it keeps running during the installation. So it won't be able to replace itself and Bugsnag.dll that it is using.
If we don't want to break auto-updating functionality, one option could be to rename the Bugsnag DLL used by the client application.

PR #2984, opened 2019-05-27
Add at least some form of backtrace generation to Bugsnag in Linux

MartinBriza:
### 📒 Description
This patch adds a basic backtrace generation to our Linux app, so we know at least something about the crashes our Linux users are experiencing. Some more work related to this should be done in the future though, especially for other signals (SIGABRT, SIGILL, etc) and especially for exceptions, quite often there are crashes coming from POCO.
In the case of exceptions, we could, however, look into doing this as a part of our library code.

### 🕶️ Types of changes

### 🤯 List of changes
- Reported Linux crashes now contain backtrace information

### 🔎 Review hints
Listing the reports from the past hour on Bugsnag helps a lot.

### DIFF
--- a/src/ui/linux/TogglDesktop/main.cpp
+++ b/src/ui/linux/TogglDesktop/main.cpp
@@ -10,6 +10,9 @@
 
 #include <stdint.h>
 #include <stdbool.h>
+#include <stdio.h>
+#include <iostream>
+#include <unistd.h>
 
 #include "singleapplication.h"  // NOLINT
 
@@ -18,6 +21,8 @@
 #include "./genericview.h"
 #include "./mainwindowcontroller.h"
 #include "./toggl.h"
+
+MainWindowController *w = nullptr;
 
 class TogglApplication : public SingleApplication {
  public:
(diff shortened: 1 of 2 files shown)


src/ui/linux/TogglDesktop/main.cpp
```diff
@@ -10,6 +10,12 @@
 
 #include <stdint.h>
 #include <stdbool.h>
+#include <stdio.h>
+#include <iostream>
+#include <unistd.h>
+
+#include "Poco/ErrorHandler.h"
```
IndrekV: This gives me error: 
```
fatal error: Poco/ErrorHandler.h: no such file or directory
```

Nowhere in linux project has poco been referenced like this. We've only referenced Poco in library. I think it's not possible to reference Poco this way. Maybe someway through library.

MartinBriza: hmh, for some reason it does compile for me, i just forgot to remove this from when i tested stuff around poco errors - i'll remove it right away

IndrekV:
I'd remove the 2 safe thread commits (revert maybe) as they will be handled separetely. Also it is important to send the `userID` to Bugsnag so we can use it to filter and search issues.

MartinBriza:
Yeah, I'm planning to do user ID (and other stuff) separately with platform info data, too
Backtraces seem like a completely separate thing to me, so I wanted to handle them on their own.

PR #2996, opened 2019-06-03
Restrict max width of client name #2993 (win)

### LINKED ISSUE #2993: Restrict the maximum width of client name to 50% of total client+project name label width
### 💻 Environment
Platform: Windows

### 📒 Description
Restrict the maximum width of client name to 50% of total client+project name label width

### ⭐️ Why do you want this?
When the client name is too long, the project name is not visible. For the user it is important to see the project name in order to be able to differentiate between different projects of the same client.

image

skel35:
### 📒 Description
Restrict the maximum width of client name to 50% of total client+project name label width

### 🕶️ Types of changes

### 👫 Relationships
Closes #2993


MartinBriza:
Nice clean patch! I have just one little thing and I'm not even sure if it's worth the effort but I'll try anyway:

image

Notice how for B, the total length of the label is shorter than in the other cases. If fixing this is not overly complex, I'd probably try doing it in a way that will spread the client name all the way so it fits the complete width of the line.

However, if it's too hard, it's probably not worth it.

MartinBriza:
Also I'm not sure why I wrote 'en' instead of 'an' :/

IndrekV:
I think that it should be clear that the client name has been cut off. Right now it's just shortened without no indication. 

Possible solutions are:
- Same logic as at the end of the row. Fading out with a gradient
- Ellipsis. Three dots added to the end of the client name when it's cut off to indicate there is more text.

I also agree with Martin that if possible we should use the whole row, but make sure that if the client name is too big it is cut off.

I checked how we do it in mac and in mac, the project name is always first and we don't do anything special other than showing ellipsis three dots to indicate there is more info. Also, we have tooltips so if the user hovers the item they will see the whole text.

Screenshot 2019-06-06 at 11 02 19

skel35:
Having both fading out and ellipsis won't look good, fading out in the middle of the line could be okay.
But copying the Mac solution = changing the order of client<->project + adding a black dot in between sounds like the easiest solution.

IndrekV:
Ok, let's do that. Copy the mac solution and add a dot. 👍

skel35:
I did not increase the width of the bullet dot before the project name to the Mac version's width because I think such changes should rather be done when we make bigger UI updates, not such smaller ones. So I just put an even smaller dot before the client name.
image

PR #2997, opened 2019-06-03
Another set of improvements for Linux packaging

MartinBriza:
### 🕶️ Types of changes

### 🔎 Review hints
This makes building packages a lot easier in both Fedora and Ubuntu (especially for me, during development).
Also, the Flatpak release should now have proper metadata.


MartinBriza:
This bug will have a `toggldesktop-branding` counterpart, however this should be merged first as it doesn't introduce anything that would break the existing scripts.

MartinBriza:
The Branding counterpart is here: https://github.com/toggl/toggldesktop-branding/pull/29

IndrekV:
Looks good to me!

PR #3004, opened 2019-06-06
Feature/create private projects win

### LINKED ISSUE #2135: If non-admin creates a Project in Desktop, they can make it public
Non-admins should not be able to create Public projects but it seems it is possible to do so with the desktop app. The issue is that the checbox for Public shouldn’t be showing for them, there should be no way for them to create a public project.

App allows everyone to create public projects, however, only admins can edit them.
User created a Project in Desktop app, it was created as Public and when User wanted to edit it in the web app, she was unable to due to not being an admin of the Workspace.

Case: #434908

### LINKED ISSUE #2967: Add possibility to create private projects in Windows app
### 💻 Environment
Platform: Windows

### 📒 Description
There should be a checkbox named `Public (visible to the whole team)` that allows to choose if the created project is public or private

Same logic is present in mac app

Screenshot 2019-05-20 at 10 54 03

skel35:
### 📒 Description
Adds `Public (visible to the whole team)` checkbox when adding new project.

### 🕶️ Types of changes

### 👫 Relationships
Closes #2967.

### 🔎 Review hints
Note: this does not fix #2135. We probably have this issue on all platforms, will have to look into it separately.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/views/EditView.xaml.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/views/EditView.xaml.cs
@@ -661,7 +661,7 @@ namespace TogglDesktop
             var ret = Toggl.AddProject(
                 this.timeEntry.GUID, this.selectedWorkspaceId,
                 this.selectedClientId, this.selectedClientGUID,
-                text, !isProjectPublicCheckBox.IsChecked.GetValueOrDefault(true), color) != null;
+                text, !isProjectPublicCheckBox.IsChecked.GetValueOrDefault(), color) != null;
 
             this.isCreatingProject = false;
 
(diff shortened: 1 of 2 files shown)


src/ui/windows/TogglDesktop/TogglDesktop/ui/views/EditView.xaml.cs
```diff
@@ -658,7 +661,7 @@ private bool tryCreatingNewProject(string text, string color)
             var ret = Toggl.AddProject(
                 this.timeEntry.GUID, this.selectedWorkspaceId,
                 this.selectedClientId, this.selectedClientGUID,
-                text, false, color) != null;
+                text, !isProjectPublicCheckBox.IsChecked.GetValueOrDefault(true), color) != null;
```
NghiaTranUIT:
```c#
.GetValueOrDefault(true)
```
I suggest that the default of `isProjectPublicCheckBox` should be a `false` because the project when creating is private project by default.

skel35: I agree, fixed.
Although this is just a backup value, it will not be used as it is a 2-state checkbox that always returns `true` or `false`.

PR #3019, opened 2019-06-10
UI Bug on macOS 10.11

### LINKED ISSUE #3002: Issues with new design and MacOs 10.11
### 💻 Environment
Platform: macOS

### 📒 Description

- [x] Mouse click on list items does not open edit popup (Using keyboard works ok and opening the edit popup for running timer is also ok)
- [x] In edit view the autocomplete dropdowns and tags dropdown does not work. Tags dropdown opens but at the very bottom of the screen. Probably an issue with fetching the coordinates. Might be needed to do it differently in older OS.
- [x] UI issues in Edit view. Some parts have gray background. I think it is caused by the fact that we use the system color in some places and in other places we use our predefined colors (like white). We should unify this and use the same color in all the places that need to have the same color.

Example: Delete button in edit view has gray background but the group button in the time entry list item has a correct white (or whatever the system color is) color.

I'll add a screenshot of the app on 10.11:

62057520_457795428304377_7975765500011479040_n

These issues should be fixed before the stable release.

NghiaTranUIT:
### 📒 Description
This PR will introduce some fixed for the UI on macOS 10.11

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Move the "Select cell logic" from `mouseUp ` to `collectionView(_ collectionView: NSCollectionView, didSelectItemsAt indexPaths`
- [x] Add isUserAction state to determine if it's from keyboard or mouse.
- [x] Fix background color for Editor
- [x] Fix missing location for AutoCompleteview

### 👫 Relationships
Close #3002 

### 🔎 Review hints
- Work on macOS 10.11

### Known-issues
- See the comment in code for further information

### DIFF
--- a/src/ui/osx/TogglDesktop/Feature/TimeEntry/Views/TimeEntryCollectionView.m
+++ b/src/ui/osx/TogglDesktop/Feature/TimeEntry/Views/TimeEntryCollectionView.m
@@ -23,24 +23,11 @@ extern void *ctx;
 - (void)awakeFromNib
 {
 	[super awakeFromNib];
-	self.isUserAction = NO;
 }
 
 - (BOOL)isFlipped
 {
 	return YES;
-}
-
-- (void)mouseUp:(NSEvent *)event {
-	[super mouseUp:event];
-	self.isUserAction = NO;
-}
-
-- (void)mouseDown:(NSEvent *)event {
-	[super mouseDown:event];
-	self.isUserAction = YES;
-
-	[self handleReslectSelectedRowWithEvent:event];
 }
 
 - (void)handleReslectSelectedRowWithEvent:(NSEvent *)event
--- a/src/ui/osx/TogglDesktop/Feature/TimeEntry/Views/TimeEntryCollectionView.m
+++ b/src/ui/osx/TogglDesktop/Feature/TimeEntry/Views/TimeEntryCollectionView.m
@@ -30,7 +30,32 @@ extern void *ctx;
 	return YES;
 }
 
-- (void)handleReslectSelectedRowWithEvent:(NSEvent *)event
+- (void)mouseUp:(NSEvent *)event {
+	[super mouseUp:event];
+
+	if (@available(macOS 10.12, *))
+	{
+		// >= macOS 10.12, the mouse Up is executed when selecting the cell
+		[self handleMouseSelectionWithEvent:event];
+	}
+}
+
+- (void)mouseDown:(NSEvent *)event {
+	[super mouseDown:event];
+
+	if (@available(macOS 10.12, *))
+	{
+        // Do nothing
+	}
+    else
+    {
+        // In macOS 10.11, the -mouseUp doesn't execute, so we have to handle the logic here
+        // if we call this logic in macOS >= 10.12 => The selection doesn't update
+        [self handleMouseSelectionWithEvent:event];
+    }
+}
+
+- (void)handleMouseSelectionWithEvent:(NSEvent *)event
 {
 	if ([event clickCount] > 1)
 	{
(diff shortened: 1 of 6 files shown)


src/ui/osx/TogglDesktop/Feature/TimeEntry/Views/TimeEntryCollectionView.m
```diff
@@ -32,33 +33,30 @@ - (BOOL)isFlipped
 
 - (void)mouseUp:(NSEvent *)event {
```
NghiaTranUIT: This method doesn't execute in macOS 10.11 when the user selects the cell by mouse. It results in the first issue in this ticket, which is "Unable to select cell in Time Entry List"

Therefore, I moved this "Selection logic" to 
```swift
func collectionView(_ collectionView: NSCollectionView, didSelectItemsAt indexPaths: Set<IndexPath>)
```
So it works.

However, I'm struggling when I couldn't re-select the cell again on macOS 10.11 because this logic
https://github.com/toggl/toggldesktop/blob/b95cdd9fff681f2406dafad1a7511801e3ce5a09/src/ui/osx/TogglDesktop/Feature/TimeEntry/Views/TimeEntryCollectionView.m#L43
 doesn't execute

lise-toggl:
[case](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/conversation/21918796562)

IndrekV:
- Opening edit view for stopped entries works
- Autocompletes work as expected
- Edit view UI items still have a strange grey background.

Screen Shot 2019-06-13 at 08 51 50

NghiaTranUIT:
@IndrekV do you have any special config for macOS 10.11?

I'm running macOS 10.11 on Parallels, and there is no grey background.

Screen Shot 2019-06-13 at 15 09 49

NghiaTranUIT:
It's ready for reviewing. Please make sure get the latest branch (I rebased).

Regarding the cell selection. I will write down the detailed note to explain what the problems is and how to workaround it in next post 👍

IndrekV:
No special setup I just have a physical macbook air with 10.11 running on it. Sadly I can't test anymore as I'm far away from that macbook. I'll pick this up again when I'm back.

NghiaTranUIT:
### Acceptance criteria
Before going to the detail, here is the goal we have to achieve
1. Able to select Time Entry Row in macOS 10.11 and latest OSX
2. If we already select the cell (Grey color) by mouse, then we select it again -> The Editor view should be presented. 
3. Select on group cell => should expand / collapse the sub-items
4. Select row by keyboard shouldn't present the Editor View

### Problem
Unable to select row on macOS 10.11 https://github.com/toggl/toggldesktop/pull/3019#discussion_r291992596

### Original logic
Before implementing new TimeEntryList, we used NSTableView and use this func to handle the `click` on TableViewCell

https://github.com/toggl/toggldesktop/blob/6a48d7fe802d2c7585a84c408db9d7e239ad7901/src/ui/osx/TogglDesktop/TimeEntryListViewController.m#L478

However, this func is **unavailable** in NSCollectionView => That why we have to find alternative approach.

### Solution 1
In order to handle row selection, it's trivial to use [collectionView:didSelectItemsAtIndexPaths:](https://developer.apple.com/documentation/appkit/nscollectionviewdelegate/1528236-collectionview?language=objc)

#### Pros
- Work natively on all macOS

#### Cons
- 2nd and 3rd criteria don't fulfill since this selection delegate doesn't execute if we select on the row, which is already selected.
- We have to workaround to fulfill the 2nd and 3rd criteria

### Solution 2
We mimic the `-performClick` by overriding the `-mouseUp` in `TimeEntryCollectionView`, and handle the selection logic
https://github.com/toggl/toggldesktop/blob/5183ac505886a21fc8520060faab5bb7dc5d196a/src/ui/osx/TogglDesktop/Feature/TimeEntry/Views/TimeEntryCollectionView.m#L33

#### Pros
- 2, 3 and 4 criteria are fulfilled.

#### Cons
- macOS 10.11 doesn't work (as the main isssue). The problem is: `-mouseUp` doesn't execute on macOS 10.11.

### Solution 3
Override the `-mouseDown` and handle Selection logic

#### Pros
- Work perfectly on 10.11

#### Cons
- on >= macOS 10.12, the `-mouseDown` is executed before the selection has been made, so the selection logic is broken 

### Solution 4: The hack
Basically, it's the combination of the 2nd and 3rd solution. We will introduce some #available to check the OS version.

#### Pros
- All critera are fulfilled in all OS X
https://github.com/toggl/toggldesktop/blob/d99369c55c0c42f390328d0978155c8a310a617e/src/ui/osx/TogglDesktop/Feature/TimeEntry/Views/TimeEntryCollectionView.m#L33-L56
#### Cons
- It's a hack.

NghiaTranUIT:
@IndrekV ☝️ it's the explanation why the Hack on this PR exists. Please let me know if we have better idea.

IndrekV:
All the issues with selecting time entry item are fixed. Sadly the grey background issue still shows on my macOs 10.11.6 machine. As the timer description field seems to work as expected maybe we can use the same logic in edit view fields to fix them. I'll look into it later today.

Screenshot of the grey fields.
Screen Shot 2019-06-26 at 07 45 07

NghiaTranUIT:
@IndrekV If you're looking on it, here is the hints

## Background box


## TextField


## In master version
- In master version, I don't override the background color for "Background Box" -> Causes the background issues

## In this PR
- I deliberately override it, so it must be matched with the background of Editor View.

## macOS 10.11.6
- However, on your OS 10.11.6, the background doesn't seem work, but It works on macOS 10.11.
- "Dark menubar and Toolbar" in General couldn't reproduce this bug.

IndrekV:
Thank you for the detailed explanations about the decisions made and the solution that was chosen.

nidarasheed:
[Case](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/conversation/22673589041)

Issue #2735, opened 2019-01-03
Can't run Windows Toggl Desktop on Debug mode 


NghiaTranUIT:
### 💻 Environment
Platform: Windows
OS Version: Windows 10
Toggl Version: master
Visual Studio: 2017 (VS 2013 for toolset v120)

### 🐞 Actual behavior
- Unable to run on Debug mode
- Able to run on Release_VS mode

### 💯 Expected behavior
- Should work on both modes

### 🔨 Steps to reproduce
1. Clone master
2. Run Toggl Desktop on Debug mode
3. 💥 

### 📦 Additional info
There are mis-configurations on Debug mode
<img width="351" alt="screen shot 2019-01-03 at 16 14 56" src="https://user-images.githubusercontent.com/5878421/50630289-d7a34280-0f72-11e9-9ebb-0a3e3fb6acd4.png">

Technically, it should use relative path, like Release_VS does

<img width="357" alt="screen shot 2019-01-03 at 16 15 16" src="https://user-images.githubusercontent.com/5878421/50630303-e8ec4f00-0f72-11e9-8ea6-659f11c21aa0.png">

NghiaTranUIT:
Close due to #3022

PR #3022, opened 2019-06-11
64-bit config for Windows

### LINKED ISSUE #3025: Create 64bit configuration and successfully build the app in 64bit
Part 1 of #2954 

### 💻 Environment
Platform: Windows

### 📒 Description
Create a 64bit configuration and successfully build the app in 64bit.

AC: It is possible to build the app both in 32bit and in 64bit versions. 

To check if the running app is 32 or 64bit you can look into Task Manager and see if the app has a "(32 bit)" ending like here 
image

skel35:
### 📒 Description
Adds a 64-bit build configuration for Windows

A corresponding PR from `branding` repo has to be merged together with this one so that 32-bit release script is not broken https://github.com/toggl/toggldesktop-branding/pull/30 

### 🕶️ Types of changes

### 🤯 List of changes
- Only 2 platform solution-level configurations are now available: `x86` and `x64`, the solution file was cleaned up heavily
- 64-bit config of POCO added as a separate vcxproj files for simplicity (such files were already present in POCO repository), 64-bit projects are ignored when building `x86` and vice versa so it causes no performance hits during the build process.
- 64-bit configs added for TogglDesktopDLL, TogglDesktop, TogglDesktopUpdater
- cleaned up unused dependencies in TogglDesktopDLL.vcxproj
- removed an assertion of `is_sqlite_threadsafe` - seemed useless, this parameter is explicitly specified in default vcxproj or makefile.

### 👫 Relationships
Closes #3025.

### 🔎 Review hints
- `git clean -xdf`
-  try `Release_VS` and `Release` configs, `x86` and `x64` platforms


skel35:
This PR also makes it possible to run Debug build, although with a non-crashing heap corruption error (includes same changes that Nghia and I had done previously). I suggest that after this PR we close #2735 because Debug version becomes possible to run if needed and open a new issue about heap corruption error as it may be caused by an actual issue in the code.

NghiaTranUIT:
@skel35 Can you elaborate some acceptance criteria for this ticket? 

https://github.com/toggl/toggldesktop/issues/3025 doesn't describle any thing, which I have to check

skel35:
@NghiaTranUIT Acceptance criterion is: it is possible to build and run both a 32bit and a 64bit version of Toggl Desktop on Windows. Added a little more info in #3025. 
Also try building it from a clean repo, using `git clean -xdf` for example.

skel35:
There are also issues with Timeline and Autotracker that 64-bit build is supposed to get fixed, but I suggest to do it later because I haven't found how to reproduce those issues yet, Timeline and Autotracker work on my 64bit Windows 10 even with 32bit TogglDesktop.

src/database.cc
```diff
@@ -50,19 +50,6 @@ Database::Database(const std::string db_path)
 
     session_ = new Poco::Data::Session("SQLite", db_path);
 
-    {
-        int is_sqlite_threadsafe = sqlite3_threadsafe();
```
NghiaTranUIT: IMO, I don't think that we should remove this assertion for couple reasons:
1. This logic will affect all other platforms: Although it's useless in Windows, but it might be not in linux or macOS.
2. This removal is out of scope of this ticket, which is 64-bit config (not refactor)

If we would like to remove it (especially in Library), I suggest that we could open ticket, then we will evaluate it carefully.

skel35: Agree that it's out of scope, will revert these changes

NghiaTranUIT:
I'm genuinely happy with this PR 👏

All redundant configs are gone and it's now supported 86bit and 64bit too.

Confirmed that I'm able to
- Run and build successfully for x86
- Run and build successfully for x64
- Basic functionalities are remained.
- Clean all caching file before testing and all are good.

@skel35 please review my previous feedback, when it's done. I'm glad to push it forward 👍 
Good job @skel35 💯

NghiaTranUIT:
Additionally, https://github.com/toggl/toggldesktop-branding/pull/30 was approved and merged 👍

src/database.cc
skel35: Done

skel35:
Thanks 👍

Issue #3041, opened 2019-06-24
Description is focused after clicking "+ Add project" or the tag icon


skel35:
### 💻 Environment
Platform: macOS
OS Version: macOS 10.14
Toggl Version: 7.4.430

### 🐞 Actual behavior
Edit view is open and the Description field is focused

### 💯 Expected behavior
When "+ Add project" is clicked in the time entry list, the Edit view should be open with Project field being focused. Same applies to the tag icon or the Client name.

### 🔨 Steps to reproduce
1. In the Time entry list click on "+ Add project"
![description_is_always_focused](https://user-images.githubusercontent.com/16451391/60008369-a3385580-967c-11e9-8ec7-af8187238eb8.gif)

NghiaTranUIT:
To be clear, the Editor should focus on a specific textfield, which we clicked in Time Entry. For instance:
1. Click on Description -> Editor will focus on Description
 2. So all for Project and Tag

skel35:
Fixed by #3042

PR #3042, opened 2019-06-25
Focus textfields on Editor

NghiaTranUIT:
### 📒 Description
This PR will introduce the feature to focus on specific textfields, which we select on Time Entry List. It's old behavior of Time Entry Editor, but we forgot to bring it up to new home.

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Focus on text fields, depend on current selection on TimeEntryCell
- [x] Support focus on Tag Token views

### 👫 Relationships
#3041 

### 🔎 Review hints
1. Select following labels in Time Entry Cell, and observe the focus ring in TimeEntry Editor
- Click on Title label -> The Description textfield should be focused
- Click on Project label -> The Project textfield should be focused
- Click on Duration -> The duration should be focused
- Click on Tag -> The first tag view should be focused
- Otherwise -> The Description textfield should be focused

### DIFF
--- a/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/EditorViewController.swift
+++ b/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/EditorViewController.swift
@@ -48,6 +48,7 @@ final class EditorViewController: NSViewController {
         didSet {
             fillData()
             registerUndoForAllFields()
+            setFocusOnTextField()
         }
     }
     private var selectedProjectItem: ProjectContentItem?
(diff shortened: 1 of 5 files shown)


skel35:
@NghiaTranUIT After changing a project, tag, or duration using a keyboard, the focus now jumps to the description field. This worked fine on `master`.

NghiaTranUIT:
@skel35 Sure, I will take a look on it. The problem is that when we update the content (Project, Description, ...) the Editor will re-assign the TimeEntry instance from Library again -> Cause the SetFocus func resets to default behavior, which is Description label. Let me see how to workaround it 👍

NghiaTranUIT:
@skel35 This last commit resolves our issue. Let check it out 👍

PR #3053, opened 2019-06-28
Disable logging for QtWebEngine

### LINKED ISSUE #2941: Over-flooding error logs in Linux
### Description
There is a user's report that Toggl accidentally floods the system with huge number of message log. It logs repeatedly every single 11 microsecond, so it results in out of disk space.

We have to investigate it on Desktop Library, since it potentially affects all platform, not just Linux.

### 💻 Environment
Platform: Library and Lunix in particular

### 🐞 Actual behavior
- Flooding the system with tons of line error message.

### 💯 Expected behavior
- Only write 1 error message log.

### 📦 Additional info
conversations/21895615070

MartinBriza:
### 📒 Description
This should help with log flooding in case of Snapcraft uninstall. I haven't actually tested this yet, personally, as I don't have a working Snapcraft configuration on my computer at this moment (can't download an Ubuntu image using `multipass` while running `snapcraft`)

### 🕶️ Types of changes

### 🤯 List of changes
- Logging is now disabled for QtWebEngine

### 👫 Relationships
Closes #2941

### 🔎 Review hints
While TogglDesktop installed using Snap is running, uninstall it. Without this patch, you should notice the process logging huge amounts of error messages in your system log. With this patch, it should not log anything.

### DIFF
--- a/src/ui/linux/TogglDesktop.sh
+++ b/src/ui/linux/TogglDesktop.sh
@@ -14,6 +14,7 @@ if [ -f "$dirname/QtWebEngineProcess" ]; then
   QTWEBENGINEPROCESS_PATH=$dirname/QtWebEngineProcess
   export QTWEBENGINEPROCESS_PATH
 fi
+export QTWEBENGINE_CHROMIUM_FLAGS="--disable-logging"
 
 # Xubuntu, i3 and Cinnamon tray icon fix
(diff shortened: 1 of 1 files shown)


IndrekV:
Seems to be nonbreaking commit, so let's just merge it.

PR #3054, opened 2019-06-28
Fix compilation warnings coming from the library

### LINKED ISSUE #2701: Fix lib warnings
### 💻 Environment
Platform: macOS/Windows/Linux

### 📒 Description
There's a lot of warnings coming from the library, especially unused variables and wrong initialization order in the classes which makes catching real issues pretty hard.

MartinBriza:
### 📒 Description
Fix most of the warnings coming from the library when building in gcc and clang.

### 🕶️ Types of changes
- Slight enhancement

### 👫 Relationships
Closes #2701 

### 🔎 Review hints
Since the tests are very limited, maybe postpone reviewing this PR for when there is more of them.
There should be no to very little issues but considering the signed/unsigned integer change, we should think of testing stuff thoroughly.
I'm going for a vacation next week, so expect movement in the test department from my side when I return.

### DIFF
--- a/src/analytics.cc
+++ b/src/analytics.cc
@@ -21,6 +21,7 @@
 
 namespace toggl {
 
+
 void Analytics::Track(const std::string client_id,
                       const std::string category,
                       const std::string action) {
--- a/src/analytics.h
+++ b/src/analytics.h
@@ -96,9 +96,9 @@ class GoogleAnalyticsSettingsEvent : public Poco::Task {
         : Poco::Task("GoogleAnalyticsSettingsEvent")
     , client_id_(client_id)
     , category_(category)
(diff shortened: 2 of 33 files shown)


IndrekV:
Just to summarize the changes of this PR:

- using `nullptr` instead of `(0)` and `NULL`
- Remove argument names from method declarations
- remove `return noError` as it is not needed and removing it is the same as returning empty string that the `noError` represents
- changing `int` values to or from `uint` to avoid conversions and castings in code

NghiaTranUIT:
Just wondering that whether this PR is ready for reviewing 🤔 since I would like to take down 1 Linux PR today 🌮

MartinBriza:
Yes, both this and #2861 are ready for review

PR #3056, opened 2019-07-01
Do not record timeline when PC is locked

### LINKED ISSUE #1554: Timeline keeps tracking "Windows Default Lock Screen" while being idle
Windows 10 user has reported following:
Previously, when I was away from my PC, nothing was logged in Timeline and I saw a gap between the blue bars in the "Timeline" plot, but now I see blue bars even when I'm away from the PC,
indicating that "Windows Default Lock Screen" was running.
Desk case #199401

### LINKED ISSUE #1965: Timeline tracks LockApp.exe - "Windows Default Lock Screen" while in standby. [bug][win]
Expected behavior: When computer locked/asleep, timeline should show no activity/idle.

Actual behavior 1: When computer changes to S1-S4, LockApp.exe is listed for the entire time at which system "asleep".

Actual behavior 2: When computer is locked, LockApp.exe is listed for entire time at which system is locked. Not listed as idle.

Assumption: Similar behavior would likely manifest in S0 low-power idle, in systems supporting this.

This problem exists as of release 7.3.326.

This is similar to #1554 in that the behavior is similar and the problem may have the same cause; however, I have insufficient time to explore possible causes.

I believe a solution lies in handling a SystemEvents.PowerModeChanged event, for the case of sleep. I can't remember if handling the SessionSwitch system event would work for this. This could probably also be handled by merely displaying entries for LockApp.exe as idle; however, that would be sloppy.

### LINKED ISSUE #2914: Entry stops running when locking computer
### 💻 Environment

Platform: Windows
OS Version: Windows 10 version 1803
Toggl Version: 7.4.345

### 🐞 Actual behavior
"Idle detection" is on and "Stop running entry on computer sleep/shutdown" as well. 
An entry is running and I lock my computer. When I log back on I see that the time entry has stopped running.

Also no "Idle" message is being displayed.

### 💯 Expected behavior
The preference states that the entry should only stop on "sleep" or "shutdown". 
I expect the time entry to continue running (it is not because my computer is locked that I stop working on the running item). When logging back on to my computer, it should ask me if I would like to add the "Idle" time to my current running entry.

### 🔨 Steps to reproduce
1.  Start a time entry
2. Lock your computer
3. Login again
4. The time entry has stopped working

skel35:
### 📒 Description
Timeline data should not be recorded when the PC is locked or in sleep mode

### 🕶️ Types of changes

### 🤯 List of changes
- [x] add toggl_set_locked() and toggl_set_unlocked() to the API (lib)
- [x] extend WindowChangeRecorder to not record while locked or sleeping (lib)
- [x] trigger SetSleep(), SetWake(), SetLocked() and SetUnlocked() correctly (win)

### 👫 Relationships
Closes #1965
Closes #2914
Closes #1554

### 🔎 Review hints
Check that:
- Timeline recording still works on Windows, MacOS, and Linux

On Windows:
- During the time when PC is locked/asleep the timeline data is not being recorded
- When "Stop running entry on computer sleep/shutdown" is checked - entry should not be stopped if PC is just locked

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -3780,6 +3780,7 @@ void Context::SetSleep() {
     if (!isHandled) {
         logger().debug("SetSleep");
         idle_.SetSleep();
+		window_change_recorder_->SetIsSleeping(true);
     }
 }
 
@@ -3986,6 +3987,7 @@ void Context::onWake(Poco::Util::TimerTask& task) {  // NOLINT
         }
 
         idle_.SetWake(user_);
+		window_change_recorder_->SetIsSleeping(false);
 
         Sync();
     }
@@ -3998,6 +4000,16 @@ void Context::onWake(Poco::Util::TimerTask& task) {  // NOLINT
     catch (const std::string& ex) {
         logger().error(ex);
     }
+}
+
+void Context::SetLocked() {
+	logger().debug("SetLocked");
+	window_change_recorder_->SetIsLocked(true);
+}
+
+void Context::SetUnlocked() {
+	logger().debug("SetUnlocked");
+	window_change_recorder_->SetIsLocked(false);
 }
 
 void Context::SetOnline() {
--- a/src/context.cc
+++ b/src/context.cc
@@ -3780,7 +3780,7 @@ void Context::SetSleep() {
     if (!isHandled) {
         logger().debug("SetSleep");
         idle_.SetSleep();
-		window_change_recorder_->SetIsSleeping(true);
+        window_change_recorder_->SetIsSleeping(true);
     }
 }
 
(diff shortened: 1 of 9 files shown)


NghiaTranUIT:
@skel35 let me know your thought about this big file, then we could proceed.

Btw, Good job @skel35 for the first Library-related ticket 👍 🎉

skel35:
@NghiaTranUIT Thanks for the review 👍 
`TogglApi.cs` is auto-generated from `toggl_api.h`. The difference in indentation is probably due to differences between how autogeneration works in Windows and Mac/Linux. I think because it's auto-generated we don't need to read it and it's fine to make sure that `toggl_api.h` file is correct.

NghiaTranUIT:
It's reasonable. Thank for the answer 👍

PR #3059, opened 2019-07-02
UX improvements: Double click to select word, triple click to select entire field

### LINKED ISSUE #1826: Improve windows edit view UX
Use regular windows actions:
- double click selects word
- triple click the whole field

when you select (by pressing tab/enter) a suggested task, the cursor stays at the begining of the phrase should be at the end.

User suggest:

> I wish the edit window would work like Firefox URL field: typing some word would show suggestions that, as I navigate through them with the cursor, are already selected; hitting tab would select and move to the next field, while hitting backspace would edit the currenct selection from the last character

skel35:
### 📒 Description
Double click should select a single word, triple click should select the entire field.
Currently, double click selects the entire field and, what's annoying, there's no quick way to select a single word.

### 🕶️ Types of changes

### 👫 Relationships
Closes #1826

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/controls/ExtendedTextBox.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/controls/ExtendedTextBox.cs
@@ -28,6 +28,11 @@ namespace TogglDesktop
                 this.SelectionLength = 0;
             }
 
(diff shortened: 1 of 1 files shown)


skel35:
@NghiaTranUIT It is implemented by default, but we have overridden it for some reason 😄 
There was probably a need to override some special case, and doing it broke the default behavior.

PR #3071, opened 2019-07-08
Small improvements to a Windows developer environment

skel35:
### 📒 Description
- added some stuff to `.gitignore` 
- .csproj minor cleanup


.gitignore
```diff
@@ -173,6 +176,12 @@ src/ui/windows/TogglDesktop/TogglDesktopDLLInteropTest/bin
 # poco binaries
 third_party/poco/bin
 third_party/poco/bin64
+third_party/poco/lib
```
skel35: @IndrekV I remember you reverted this change a while ago, but I totally didn't get why 😃 These are the build artifacts clearly and it's inconvenient to have them in unstaged files all the time. I have these paths in my `exclude` file anyway.

IndrekV: As it states in https://github.com/toggl/toggldesktop/commit/c6e8c86abd834b7f4db0b000d79f594619ccdf24
We used static libraries for libcyrpto and libssl libraries. To test if this works ok. You need to do a clean clone after these changes and try to build the app. Maybe with all the updates we''ve changed the way libcrypo and libssl are used.

skel35: I can test it with a clean clone ofc, although I don't get the issue with ignoring this folder either way. There's nothing in this folder in the repo right now and the files would get copied there during the build regardless. We never have to commit any files from this folder to the repo anyway 🤷‍♂

IndrekV: Jep, I think it's ok. Tried to build it and everything works ok.

skel35: Great, thanks 👍

PR #3073, opened 2019-07-08
Switch from packages.config to PackageReference (win)

skel35:
### 📒 Description
Switches from the old `packages.config` method of referencing packages to the new `PackageReference` method. 
The benefits are:
- NuGet restore from `PackageReference` is built-in to MSBuild
- There is no need for an extra `packages.config` file
- The `.csproj` files look more concise as well
- More advanced behavior is supported (which we don't use for now)

See https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files .

### 🕶️ Types of changes

### 🔎 Review hints
- Building from VS works
- Building from the command line using MSBuild works. Should be run with `/restore` option (can work without this option depending on the local configuration in `%APPDATA\NuGet`)
Example command:
`MSBuild.exe src/ui/windows/TogglDesktop/TogglDesktop.sln /property:Configuration=Release /p:Platform="x64" /t:Clean,Build /restore`


IndrekV:
It work very well with updates in https://github.com/toggl/toggldesktop-branding/pull/33

PR #3081, opened 2019-07-11
Fix for Idle popup not visible

### LINKED ISSUE #2479: Idle detection notification scaling issue on a Surface Book 2 Windows
Case 718426

> Due to display peculiarities, when I unlock my Surface Book 2, a lot of things on the screen change size. With Toggle, this causes a problem because when the idle time detection dialog displays, sometimes it gets sized so small that only the title bar is visible, and I can&#39;t interact with it or resize it. Could you verify whether that dialog has a minimum size set? If not, could you set one? If it already has one, then it may take more investigation to figure out why it shrinks to title-bar size. It might also be helpful to make that dialog resizable, so that at least if it shrinks too small to work with, I can manually resize it.

### LINKED ISSUE #3009: Idle popup too small
### 💻 Environment
Platform: Windows
OS Version: Windows 10
Toggl Version: Master

### 🐞 Actual behavior
The idle popup is very small. As big as it's title bar.

### 💯 Expected behavior
Idle popup should have a minimum size so that it is always possible to interact with the buttons in the popup

### 📦 Additional info
User feedback:
>Idle dialog still needs a minimum size specified. When it appears on a locked Windows 10 system that uses DisplayPort and multiple monitors, it can appear as a tiny box no bigger than its title bar, and then it can't be interacted with.

skel35:
### 📒 Description
`updateMaximumSize()` was the only place in the code that can change the size of the window, so I made sure it's not called for our tool windows, as even the largest of tool windows does not have to be resized.
Also set the minimum size for the idle notification window, just in case.

### 🕶️ Types of changes

### 👫 Relationships
Closes #3009
Closes #2479 

### 🔎 Review hints
I wasn't able to reproduce this exact issue, but the `updateMaximumSize()` method triggered some other methods that are not guaranteed to work correctly.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/chrome/ChromelessWindow.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/chrome/ChromelessWindow.cs
@@ -96,13 +96,14 @@ namespace TogglDesktop
 
         protected override void OnLocationChanged(EventArgs e)
         {
-            this.updateMaximumSize();
-
-            if (!this.isToolWindow
-                && this.WindowState != WindowState.Maximized
-                && this.ResizeMode != ResizeMode.CanResize)
+            if (!this.IsToolWindow)
             {
-                this.ResizeMode = ResizeMode.CanResize;
+                this.updateMaximumSize();
+                if (this.WindowState != WindowState.Maximized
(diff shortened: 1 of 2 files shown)


IndrekV:
Looks good! 👍

PR #3085, opened 2019-07-12
Fixes ugly grey input backgrounds in macos 10.11

### LINKED ISSUE #3076: The grey background issue on macOs 10.11.
### 💻 Environment
Platform: macOS
OS Version: macOS 10.11.6
Toggl Version: 7.4.444 (latest beta)

### 🐞 Actual behavior
Sadly the grey background issue still shows on my macOs 10.11.6 machine. As the timer description field seems to work as expected maybe we can use the same logic in edit view fields to fix them.

Screenshot of the grey fields.
Screen Shot 2019-06-26 at 07 45 07

_Originally posted by @IndrekV in https://github.com/toggl/toggldesktop/pull/3019#issuecomment-505717439_

IndrekV:
### 📒 Description
changed the Appearance of the edit view to always be `aqua` when on a system that does not support dark mode

### 🕶️ Types of changes
**Bug fix**

### 👫 Relationships
Closes #3076

### 🔎 Review hints
Start the app on macos 10.11 and make sure the app looks like this 

Screen Shot 2019-07-12 at 23 39 03

and not like on the screenshot in the issue

Screen Shot 2019-06-26 at 07 45 07

Some small grey corners are still visible in the input edges in the edit form. I just could not figure out why they are there. Maybe it's also connected to the appearance setting, but I tested and was not able to detect the cause. Luckily these small corners are very very small and we can ship the new version with them. Maybe once @NghiaTranUIT is back he can take another look at the corners.

### DIFF
--- a/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/EditorPopover.swift
+++ b/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/EditorPopover.swift
@@ -26,6 +26,7 @@ final class EditorPopover: NoVibrantPopoverView {
 
     @objc func prepareViewController() {
         let editor = EditorViewController(nibName: NSNib.Name("EditorViewController"), bundle: nil)
+        editor.view.appearance = appearance
         contentViewController = editor
     }
 


NghiaTranUIT:
Thank for that fix. The problem is that NSPopover is actually embedded the NSVisualEffect and enable the Vibrant for all children textlable by default 

Additionally, the background of the Editor from the design is solid color -> Solid background color + vibrant textlable -> Unexpected color.

PR #3088, opened 2019-07-15
When adding an existing tag, just make it checked

### LINKED ISSUE #3084: Unchecking a tag doesn't remove it from the entry.
### 💻 Environment
Platform: Linux
OS Version: Ubuntu 19.04 x64
Toggl Version: 7.4.456 (installed from a deb)

### 🐞 Actual behaviour
When unchecking already checked tag nothing happens. It's not synced over to the server and state never changes. You can add only new tags and not remove them.

### 💯 Expected behaviour
When unchecking a tag it should get removed from the tags list on the time tracker entry.

### 🔨 Steps to reproduce
1. Create new entry and start tracking it.
2. Tick some of the tags.
3. Click done to verify it saved.
4. Edit the task again.
5. Uncheck one of the added tags.
6. Click done.
7. Edit the entry again and see that tag wasn't removed.

MartinBriza:
### 📒 Description
This fixes how adding tags is handled - now when adding an existing tag, it's forced to be checked (no unchecking, though).

### 🕶️ Types of changes

### 🤯 List of changes
- linux: adding an existing tag makes it selected for the current time entry

### 👫 Relationships
Closes #3084

### 🔎 Review hints
See if adding (new and existing) tags works as expected. Adding an existing tag should make the tag checked.
Also see if checking and unchecking of other (previously existing) tags works right when (re)opening the time entry editor.

### DIFF
--- a/src/ui/linux/TogglDesktop/timeentryeditorwidget.cpp
+++ b/src/ui/linux/TogglDesktop/timeentryeditorwidget.cpp
@@ -491,9 +491,7 @@ void TimeEntryEditorWidget::on_tags_itemClicked(QListWidgetItem *item) {
             tags.push_back(widgetItem->text());
         }
     }
-    if (item) {
-        tags.push_back(item->text());
-    }
+
     tags.sort();
     QString list = tags.join("\t");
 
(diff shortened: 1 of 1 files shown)


IndrekV:
Works great! Nice job.

PR #3098, opened 2019-07-18
Rework auto-update system to use github pages instead of v8 API

### LINKED ISSUE #3097: Rework auto-update system to use github pages instead of v8 API
### 💻 Environment
Platform: Windows/Linux (Mac is using Sparkle for auto-updates)

### 📒 Description
We should stop using v8 API for auto-updates and instead use `updates.json` file from https://toggl.github.io/toggldesktop/assets/releases/updates.json. 

### ⭐️ Why do you want this?
1. It brings us closer to our goal to stop using v8 API.
2. As we don't register new platforms on v8 API anymore, using `updates.json` will allow us to set up auto-updates for x64 Windows release.

skel35:
### 📒 Description
Rework auto-update system to use github pages instead of v8 API

### 🕶️ Types of changes

### 🤯 List of changes
- Use updates.json from github pages for auto-update instead of v8 API
- Auto-update from 32-bit versions on 64-bit systems to 64-bit versions on Windows
- Fixed error in `google-astyle` (`make fmt` didn't work on my machine due to this error)

### 👫 Relationships
Closes #3097

### 🔎 Review hints
- The Windows and Linux applications are now using https://toggl.github.io/toggldesktop/assets/releases/updates.json for update checking.
- On a 64-bit Windows users get a 64-bit Toggl Desktop update.
- On a 32-bit Windows users get a 32-bit Toggl Desktop update. 
- The app does not crash when the JSON file does not have the expected format.

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -1398,7 +1398,11 @@ error Context::downloadUpdate() {
             }
             auto latestVersion = root[shortOSName()][update_channel];
             url = latestVersion[installerPlatform()].asString();
-            version_number = latestVersion["version"].asString();
+            auto versionNumberJsonToken = latestVersion["version"];
+            if (versionNumberJsonToken.empty()) {
+                return error("No versions found for OS " + shortOSName() + ", platform " + installerPlatform() + ", channel " + update_channel);
+            }
+            version_number = versionNumberJsonToken.asString();
 
             if (lessThanVersion(HTTPSClient::Config.AppVersion, version_number)) {
                 std::stringstream ss;
@@ -1541,7 +1545,7 @@ bool Context::lessThanVersion(const std::string& version1, const std::string& ve
     int parsed1[4] {}, parsed2[4] {};
     parseVersion(parsed1, version1);
     parseVersion(parsed2, version2);
-    return std::lexicographical_compare(parsed1, parsed1 + 4, parsed2, parsed2 + 4);
+    return std::lexicographical_compare(parsed1, &parsed1[4], parsed2, &parsed2[4]);
 }
 
 void Context::TimelineUpdateServerSettings() {
--- a/src/context.cc
+++ b/src/context.cc
@@ -1492,6 +1492,11 @@ const std::string Context::linuxPlatformName() {
 const std::string Context::windowsPlatformName() {
 #if defined(_WIN64)
     return "windows64";
+#elif defined(_WIN32)
+    BOOL f64 = FALSE;
+    if (IsWow64Process(GetCurrentProcess(), &f64) && f64)
+        return "windows64";
+    return "windows";
 #else
     return "windows";
 #endif
--- a/src/context.cc
+++ b/src/context.cc
@@ -1373,11 +1373,8 @@ error Context::downloadUpdate() {
             return err;
         }
 
-        // Get update check URL
-        std::string update_url("");
-        err = updateURL(&update_url);
-        if (err != noError) {
-            return err;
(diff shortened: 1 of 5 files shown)


src/context.cc
```diff
@@ -1373,48 +1373,45 @@ error Context::downloadUpdate() {
...
-                }
-                return noError;
-            }
-
             Json::Value root;
             Json::Reader reader;
             if (!reader.parse(resp.body, root)) {
                 return error("Error parsing update check response body");
             }
+            auto latestVersion = root[shortOSName()][update_channel];
```
MartinBriza: What if the json format changes in the future or is generated wrong? I think we should be more defensive in this case.

src/context.cc
```diff
@@ -1373,48 +1373,45 @@ error Context::downloadUpdate() {
             return err;
         }
 
-        // Get update check URL
-        std::string update_url("");
-        err = updateURL(&update_url);
-        if (err != noError) {
-            return err;
+        if (HTTPSClient::Config.AppVersion.empty()) {
```
MartinBriza: Here I'd probably reword the error. This is good for us as the developers but if there's a broken version somewhere in the wild, the user won't know what's going on. I'd go for something like "This version cannot check for updates. This has been probably already fixed. Please check toggl.com for a newer version."
Also, reporting to bugsnag in this place would be a good idea so we know immediately.

src/context.cc
```diff
@@ -1543,6 +1528,22 @@ const std::string Context::shortOSName() {
...
+        parser.get(); //Skip period
+        parser >> result[idx];
+    }
+}
+
+bool Context::lessThanVersion(const std::string& version1, const std::string& version2) {
+    int parsed1[4] {}, parsed2[4] {};
+    parseVersion(parsed1, version1);
+    parseVersion(parsed2, version2);
+    return std::lexicographical_compare(parsed1, parsed1 + 4, parsed2, parsed2 + 4);
```
MartinBriza: Personally, I find this notion a tad confusing and very C-like. This may be just me though, I'd personally write `&parsed1[4]`, it seems a lot clearer to me.

src/context.cc
skel35: It won't fail as `Json::nullValue` when indexed will return `Json::nullValue`.
I can think of throwing an error when we can't find any version in the end, but that also means we would have an error when testing a new platform on a not-yet-released channel, which is okay I guess.

src/context.cc
skel35: 👌

src/context.cc
skel35: I think these are errors for developers, as they shouldn't happen for users (definitely not in beta or stable) so we're optimizing the developer-fix time instead of trying to explain stuff to the users. We have plenty of such errors elsewhere; actually, I just moved this piece of code. And it does get reported to Bugsnag, at least for Windows we subscribe to `toggl_on_error`.

skel35:
@MartinBriza I've addressed your comments, please check out the review again

src/context.cc
MartinBriza: If the message gets displayed in the UI (and this message will), the user will see it and get confused by it. 
It's not for optimization of the fix time (and making it more comprehensible doesn't change anything for us because we can still `grep` the string in the source code)

MartinBriza: Of course, the chance of this happening is **slim** but still, we can start handling app versions different in the future, or this can happen to somebody who got the app from some other source than us (like unofficial packages on Linux).

skel35: Updated the error message, @MartinBriza

PR #3113, opened 2019-07-22
Force focus to the delete button when backtabbing on the description field

### LINKED ISSUE #3105: Reverse tab in description field focuses date input
### 💻 Environment
Platform: Linux
Toggl Version: Master

### 🐞 Actual behavior
Pressing SHIFT+TAB in description field focuses date field.

### 💯 Expected behavior
Pressing SHIFT+TAB in description field should jump to the last item in the form - `Delete` button 

### 🔨 Steps to reproduce

1. Go to edit view
2. Focus description field
3. Press SHIFT+TAB to change focus to the previous field
4. See that the date field gets the focus

MartinBriza:
### 📒 Description
This PR forces backtabbing to focus the delete button when the user backtabs on the description field.

### 🕶️ Types of changes

### 🤯 List of changes
- Linux: backtabbing works as expected in the time entry editor

### 👫 Relationships
Closes #3105

### 🔎 Review hints
Test if tabbing and backtabbing works as expected in the time entry editor.

### DIFF
--- a/src/ui/linux/TogglDesktop/timeentryeditorwidget.cpp
+++ b/src/ui/linux/TogglDesktop/timeentryeditorwidget.cpp
@@ -342,6 +342,17 @@ bool TimeEntryEditorWidget::eventFilter(QObject *object, QEvent *event) {
                                                     "");
         }
     }
+    if (event->type() == QEvent::KeyPress) {
+        if (object == ui->description) {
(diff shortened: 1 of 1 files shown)


IndrekV:
I checked out a possible solution but in terms of tab index setup everything is correct and it should work, but it just doesn't. It's a hack but it works so it's good enough.

PR #3121, opened 2019-07-23
Save last Editor Window size

### LINKED ISSUE #3047: Remember the size of the edit popover
The follow-up to #3040 

### 💻 Environment
Platform: macOS

### 📒 Description
The width of the edit popover should be remembered, as it currently does on Windows. We already support it in API (`toggl_set_window_edit_size_width`, `toggl_get_window_edit_size_width`) so it should be trivial to add it to MacOS.

### ⭐️ Why do you want this?
For the convenience of the users who usually have really long project names and descriptions.

NghiaTranUIT:
### 📒 Description
This free feature has been implemented in library. This PR will adopt it in macOS Desktop.

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Get / Set editor window size. Return default size if the size is lesser than the default size
- [x] Set Editor size after resizing.
- [x] Override the Editor window size

### 👫 Relationships
Close #3047 

### 🔎 Review hints
- Resize the Editor -> Close app -> Open again -> If the Editor window size is remained -> Correct
- Try to set invalid size in database or in code => Editor window falls into default size -> Correct

### DIFF
--- a/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/EditorPopover.swift
+++ b/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/EditorPopover.swift
@@ -27,6 +27,8 @@ final class EditorPopover: NoVibrantPopoverView {
     @objc func prepareViewController() {
         let editor = EditorViewController(nibName: NSNib.Name("EditorViewController"), bundle: nil)
         editor.view.appearance = appearance
+        let size = DesktopLibraryBridge.shared().getEditorWindowSize()
+        editor.view.frame.size = size
         contentViewController = editor
     }
 
@@ -41,5 +43,9 @@ final class EditorPopover: NoVibrantPopoverView {
         if let editor = contentViewController as? EditorViewController {
             editor.timeEntry = timeEntry
         }
+    }
+
+    override func popoverDidResize() {
+        DesktopLibraryBridge.shared().setEditorWindowSize(contentSize)
     }
 }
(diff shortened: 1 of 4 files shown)


skel35:
@NghiaTranUIT Tried running from Xcode, the size does not get persisted between runs.
1. Run from Xcode
2. Resize the Edit view by making it wider
3. TogglDesktop->Quit Toggl Desktop
4. Run from Xcode again, open the Edit view
Actual result: Edit view has the default size, not the wider size

src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/EditorPopover.swift
```diff
@@ -27,6 +27,8 @@ final class EditorPopover: NoVibrantPopoverView {
     @objc func prepareViewController() {
         let editor = EditorViewController(nibName: NSNib.Name("EditorViewController"), bundle: nil)
         editor.view.appearance = appearance
+        let size = DesktopLibraryBridge.shared().getEditorWindowSize()
+        editor.view.frame.size = size
```
skel35: Could we also use `contentSize` here?

NghiaTranUIT:
I'm not sure what's wrong but your test scenario was exactly how I tested, and I couldn't reproduce it (GIF)

src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/EditorPopover.swift
NghiaTranUIT: Actually, we don't need to override the `contentSize`, since when we set the `contentViewController` in line 32, it will automatically override `contentSize`of Popover  in

https://github.com/toggl/toggldesktop/blob/29d6aee704e140c5b702c12ec4cd4ee508eacc09/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/ResizablePopover.swift#L37

and 

https://github.com/toggl/toggldesktop/blob/29d6aee704e140c5b702c12ec4cd4ee508eacc09/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/ResizablePopover.swift#L180

skel35: Ah, got it, thanks 👍

skel35:
@NghiaTranUIT Sorry, there was some issue with my dev environment.
Works well, LGTM 👍

PR #3122, opened 2019-07-23
Project Text is disappear when TE is running

### LINKED ISSUE #3038: Tabbing Causes Project Name Loss
### 💻 Environment

Platform: Mac
OS Version: N/A
Toggl Version: 7.4.430

### 🐞 Actual behavior
User primarily uses keyboard to start and edit time entries; when tabbing while starting a time entry, the action causes the project name he's entered to be weirdly cleared. 

### 💯 Expected behavior
Tabbing between fields should not cause field deletion

### 🔨 Steps to reproduce
See case link below for instructions and screenshot video of action

### 📦 Additional info
Case [link](https://app.intercom.io/a/apps/ayixs927/inbox/2137598/conversations/22409757603)

NghiaTranUIT:
### 📒 Description
If the TE is running, there is a chance that Project Text Field in 

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Ignore the Project Text Field update if the user is editing it
- [x] Ignore focus on Timer if a different Window (AutoComplete) becomes "Key". Otherwise, we can't `Tab` to focus on "Add new project" button
 
### 👫 Relationships
Close #3038 

### 🔎 Review hints
- Start empty TE -> Open the Editor -> Tab to Project TextField and type something quickly -> Observe that if the text in Project TextField doesn't update -> 💯 
- If we're editing the Project TextField -> Press "Tab" -> The "Add new project" button should be focus, not the Timer

### DIFF
--- a/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/AutoComplete/Input/ProjectAutoCompleteTextField.swift
+++ b/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/AutoComplete/Input/ProjectAutoCompleteTextField.swift
@@ -35,9 +35,11 @@ final class ProjectAutoCompleteTextField: AutoCompleteTextField {
     func setTimeEntry(_ timeEntry: TimeEntryViewItem) {
         projectCreationView.selectedTimeEntry = timeEntry
 
-        stringValue = timeEntry.projectLabel
-        layoutProject(with: stringValue)
-        applyColor(with: timeEntry.projectColor)
+        if currentEditor() == nil {
+            stringValue = timeEntry.projectLabel
+            layoutProject(with: stringValue)
+            applyColor(with: timeEntry.projectColor)
+        }
     }
 
     override func didTapOnCreateButton() {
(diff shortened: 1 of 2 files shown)


NghiaTranUIT:
## Bug on master bracnh

NghiaTranUIT:
## Technical explanation for the GIF bug
When the "Tab" key is pressed, it will trigger this action
https://github.com/toggl/toggldesktop/blob/ae53bbc3c47ca50b0fbf00dd0887aaa78c190cb8/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/AutoComplete/AutoCompleteView.swift#L150-L155

As we're editing the Project Text Field, so the EditorViewController is a key a Window. In order to focus on the "Create project button", we have to make the AutoCompleteView to become a keyWindow -> Then makeFirstRespinder on this button.

However, whenever any Windows become a Key, this action will unintentionally trigger 
https://github.com/toggl/toggldesktop/blob/ae53bbc3c47ca50b0fbf00dd0887aaa78c190cb8/src/ui/osx/TogglDesktop/TimerEditViewController.m#L102-L105

and it makes the Timer become first Responder -> Cause weird bug

IndrekV:
This bug was just introduced by me with https://github.com/toggl/toggldesktop/commit/06c20fa5f75814840cd23065ceb861914d3efab8
The goal was to always focus the timer description field when the user focuses the app. For example tabs between apps and then focuses Toggl.

We will soon update the timer focus logic to make it possible to easily switch between tasks with keyboard. More of that in this issue: https://github.com/toggl/toggldesktop/issues/2926#issuecomment-511784646

IndrekV:
Looks good. Everything working ok again, no focus jumping from edit view.

PR #3127, opened 2019-07-23
Force scrolling to items in the list only when tabbed to

### LINKED ISSUE #3120: List item continue button not working on first click
### 💻 Environment
Platform: Linux 
Toggl Version: Master

### 🐞 Actual behavior
If you click on a continue button of an entry that is not selected in the listing it will just select the item and will not trigger the continue action

### 💯 Expected behavior
Clicking on continue button in a listing entry continues the entry at first click

### 🔨 Steps to reproduce

1. Click continue arrow on any not selected item in the list
2. Entry will not start tracking

### 📦 Additional info
* First we click on an item that is not selected. It gets selected and the listing scrolls to make the selected item be in focus. The second click on the continue button triggers the timer start.

* Second time we select the entry we want with keyboard and then press continue button with the mouse. Now the timer is triggered right away.

linux-continue

MartinBriza:
### 📒 Description
This PR makes stops the time entry list from scrolling around when an item is clicked (scrolling is still enabled for tabbing). This also fixes the behavior of the Continue button.

### 🕶️ Types of changes

### 🤯 List of changes
- Linux - Clicking the continue button is necessary only once now

### 👫 Relationships
Closes #3120

### 🔎 Review hints
Check if tabbing through the time entry list, clicking the continue button and generally selecting items in the time entry list works consistently as expected.

### DIFF
--- a/src/ui/linux/TogglDesktop/timeentrycellwidget.cpp
+++ b/src/ui/linux/TogglDesktop/timeentrycellwidget.cpp
@@ -116,7 +116,9 @@ void TimeEntryCellWidget::deleteTimeEntry() {
 
 bool TimeEntryCellWidget::eventFilter(QObject *watched, QEvent *event) {
     if (event->type() == QEvent::FocusIn) {
-        focusInEvent(reinterpret_cast<QFocusEvent*>(event));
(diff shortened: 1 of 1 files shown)


IndrekV:
Works!

Issue #2059, opened 2016-11-23
MacOs grouped mode edit


IndrekV:
`PATCH` updates:

- [ ] Drag dropping group should move all entries to the other day
- [ ] When editing group a note of group edit should be visible in edit view
- [ ] Group edit changes all items (patch)
- [ ] Group delete removes all items (patch)

glen-84:
This should apply to Windows as well.

ctmaloney:
Any timeline for this?

IndrekV:
Sorry no timeline for this yet.

ctmaloney:
🙏

Still hoping for this to get some attention!

Might give it a go myself if I get some extra time... any hints on where this functionality would live in the codebase?

lise-toggl:
[User request](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/2931915/conversations/23316180285)

th0rgall:
I had inadvertently posted a duplicate of this bug (closed now, #3987). Consider this a +1 vote!

PR #3141, opened 2019-07-26
Improve keyboard

### LINKED ISSUE #2926: macOS timer selection broken
Timer selection allows users (👋) to easily delete, rename and generally navigate Toggl without having to use a mouse. This is done by keying up and down the list to change the highlighted item. One expects to be able to press <kbd>Enter</kbd> on a highlighted item to open up the editing window or <kbd>Del</kbd> to remove the item (with, perhaps, an alert appearing).

This is a standard HIG that you can see on all time tracking apps in competition with Toggl, and across macOS more generally.

### 💻 Environment

Platform: macOS
OS Version: 10.14.4 (18E226)
Toggl Version: 7.4.373

### 🐞 Actual behavior
1. Beginning selection is nearly impossible. The only way to begin the selection process, whereby the up and down arrows change timer selection, is to first click a parent item containing duplicated timer names.

    This is because clicking actual timer entries opens up their editing screens. That's expected. However, when a user _closes_ the editing screen, the timer in question should then be selected and up/down arrows should act as expected.

    There should also be ways in which users can select timers without opening an editing screen. This includes clicking and dragging a timer, right-clicking a timer (which lacks a contextual menu, oddly), etc.
2. Up and down arrows do not allow navigating into timer entries from adjacent days. The selection process halts at the first and last timer entries for the day one has currently selected into.
3. Selection should most likely originate on the active, ongoing timer on app open. This does not appear to be the case.

NghiaTranUIT:
### 📒 Description
This PR introduce some improvement in term of keyboard-based users, which are carefully mentioned in https://github.com/toggl/toggldesktop/issues/2926#issuecomment-511784646

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Support multiple selection on Time Entry list (Shift + Up / Down to select more)
- [x] Bulk action for deleting. If it's multiple selection -> Show warning. If it's single -> Old Logic
- [x] Bulk action for dragging.
- [x] Shift-Down -> Only select first Item if we're focusing on the Timer View. Otherwise, it's just like select next item (Multiple selection behavior)
- [x] If TE is running in Timer Bar and we're selecting any TE in TimeEntryList => Then we hide app -> Open app => The TimerBar will be focus (It's current logic) => Thus, we `de-select` all selection. I intentionally do it, because we can use keyboard to do action on Timer Bar (Edit, delete, stop) => It's make more sense, because we're focuing on the Timer, not TimeEntry List anymore. @IndrekV need your thought here 😄 
- [x] If the TE is running and there is no selection on TE list -> We can use Keyboard on the running TE (Edit, delete, stop)

### 👫 Relationships
Close #2926 

### 🔎 Review hints
- All acceptance criteria should be met https://github.com/toggl/toggldesktop/issues/2926#issuecomment-511784646

### DIFF
--- a/src/ui/osx/TogglDesktop/Feature/TimeEntry/TimeEntryDatasource.swift
+++ b/src/ui/osx/TogglDesktop/Feature/TimeEntry/TimeEntryDatasource.swift
@@ -423,9 +423,8 @@ extension TimeEntryDatasource {
     func collectionView(_ collectionView: NSCollectionView, pasteboardWriterForItemAt indexPath: IndexPath) -> NSPasteboardWriting? {
         guard let item = object(at: indexPath) else { return nil }
         guard !item.loadMore else { return nil }
-
         // Save indexpath
-        let data = NSKeyedArchiver.archivedData(withRootObject: indexPath)
+        let data = NSKeyedArchiver.archivedData(withRootObject: Array(collectionView.selectionIndexPaths))
         let pbItem = NSPasteboardItem()
         pbItem.setData(data, forType: NSPasteboard.PasteboardType.string)
         return pbItem
--- a/src/ui/osx/TogglDesktop/Feature/TimeEntry/TimeEntryDatasource.swift
+++ b/src/ui/osx/TogglDesktop/Feature/TimeEntry/TimeEntryDatasource.swift
@@ -98,6 +98,17 @@ class TimeEntryDatasource: NSObject {
     
     // MARK: Public
 
+    func selectFirstItem() {
+        // Check if there is a least 1 item
+        guard sections.first?.entries.first != nil else { return }
+        collectionView.window?.makeFirstResponder(collectionView)
+
+        // Deselect all because TE is multiple-selections
+        collectionView.deselectAll(self)
+        collectionView.selectItems(at: Set<IndexPath>(arrayLiteral: IndexPath(item: 0, section: 0)),
+                                   scrollPosition: .left)
+    }
+
     func process(_ timeEntries: [TimeEntryViewItem], showLoadMore: Bool) {
         isShowLoadMore = showLoadMore
 
--- a/src/ui/osx/TogglDesktop/Feature/TimeEntry/Views/TimeEntryCollectionView.h
+++ b/src/ui/osx/TogglDesktop/Feature/TimeEntry/Views/TimeEntryCollectionView.h
@@ -16,7 +16,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (strong, nonatomic) NSIndexPath *clickedIndexPath;
 
-- (TimeEntryCell *)getSelectedEntryCell;
+- (NSArray<TimeEntryCell *> *)getSelectedEntryCells;
 
 @end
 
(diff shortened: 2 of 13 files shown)


skel35:
My feedback after testing the branch:

- Shift-Up doesn't work for selection (Shift-Down works fine)
- Shift-Click works for selection, but doesn't work for deselection
- When collapsed group is selected and the user pressed Backspace -> only the first entry from the group gets deleted
- Dragging a group (collapsed or expanded) -> Only first entry gets moved

Out of scope:
- https://i.imgur.com/POPOgDU.png When expanded group is selected and the user pressed Enter -> Edit window for the first entry of the group is displayed, but it looks like it's the entire group

Unrelated:
- In Manual mode I'd expect Cmd-N to add a new time entry and not start the timer. Is it a bug?
- In Manual mode when entries list is focused and the user presses Esc -> I'd expect the focus switch to Add Time Entry button.

NghiaTranUIT:
```
- Shift-Up doesn't work for selection (Shift-Down works fine)
- Shift-Click works for selection, but doesn't work for deselection
- https://i.imgur.com/POPOgDU.png When expanded group is selected and the user pressed Enter -> Edit window for the first entry of the group is displayed, but it looks like it's the entire group
```
The last three commits are fixed above☝️ issues.
----------------------------
However, two extra requirements need library's update
```
- When collapsed group is selected and the user pressed Backspace -> only the first entry from the group gets deleted
- Dragging a group (collapsed or expanded) -> Only first entry gets moved
```
To elaborate, the Time Entry List doesn't know the instance of children TE if the Group TE is collapsed. For instance, if we have 5 children TE, the Library will sent 1 Group TE (with count = 5), and no children data. Therefore, if we want to drag and delete a group -> The library should be supported.

I tried to improve the library but it takes more effort and I suppose that this Bulk edit should implemented in separate tickets.  
----------------------------
```
- In Manual mode when entries list is focused and the user presses Esc -> I'd expect the focus switch to Add Time Entry button.
```
It's not correct behavior in term of macOS user. The ESC should deal with the focus. From what I understand, most of the time, ESC is related to "Cancel" current action or close the PopOver.

skel35:
@NghiaTranUIT Thanks, I'll create a ticket for bulk-edit. 
Regarding the last point about Manual mode, how the user can switch the focus to Add Time Entry button using a keyboard?

skel35:
@NghiaTranUIT Hm, I guess when Cmd-N for Manual Mode is fixed that would be enough for the keyboard-only flow. Then it's fine :)

NghiaTranUIT:
@skel35 please review https://github.com/toggl/toggldesktop/pull/3167, and it will fulfill your feedback 👍

skel35:
#2059 Issue for grouped entries bulk edit/delete/drag

skel35:
@NghiaTranUIT I think almost everything is fixed, very nice 👍 
I still see a problem when pressing Shift-Down a few times, then Shift-Up doesn't do deselection.
2xShift-Up-then-Shift-Down
The opposite works (Shift-Up, then Shift-Down)

NghiaTranUIT:
Hm, ok, it's different behavior than the original tickets. I'm looking on it now 👍

PR #3145, opened 2019-07-26
Description text field disappears when the TE is running

### LINKED ISSUE #3140: Running entry description cleared in edit popup
### 💻 Environment
Platform: macOS
OS Version: macOS 10.14
Toggl Version: master

### 🐞 Actual behavior
Description field is cleared when autocoplete items are shown

### 💯 Expected behavior
If user is editing description field (descrtiption field is in focus) do not clear or automatically update the contents.

### 🔨 Steps to reproduce
1. Start timer 
2. open edit real fast (cmd+e)
3. type something
4. the typed content is cleared

### 📦 Additional info

We had a similar issue with project field that was fixed with #3122 

mac-description-clear

NghiaTranUIT:
### 📒 Description
This PR introduce the fix when the description text field disappears in Editor View when the TE is running 

### Problems for this issues
The problem is actually how the Library works: Whenever we start a new TE -> The Library keep consistently updating the TE to Editor by notification (I believe that it's our shortcoming) => It results in the update of Editor -> The content, which the user is typing, will disappear ✈️ 

### Solution
- If the GUID of the current TimeEntry in Editor is `==` to the previous TE, and the Description TextField is focusing => Ignore the update
- Otherwise, it should be update as usual

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Update logic to prevent updating description text fields

### 👫 Relationships
Close #3140 

### 🔎 Review hints
- Start Empty TE then quickly opening the Editor and typying something in Description Text Field -> If the content doesn't disappear and it remains => 💯

### DIFF
--- a/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/EditorViewController.swift
+++ b/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/EditorViewController.swift
@@ -46,7 +46,7 @@ final class EditorViewController: NSViewController {
 
     var timeEntry: TimeEntryViewItem! {
         didSet {
-            fillData()
+            fillData(oldValue)
             registerUndoForAllFields()
 
             // Check if the Editor has update with new TimeEntry -> Reset focus to Description TextField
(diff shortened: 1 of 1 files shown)


IndrekV:
Works! Nice job!

PR #3155, opened 2019-07-29
Group entries with same tags but different order, closes #3133 (lib)

### LINKED ISSUE #3133: Entries with same tags but different order are not grouped
### 💻 Environment

Platform: Windows
OS Version: 10
Toggl Version: 7.4.422

### 🐞 Actual behavior
Two time entries with the same tags, but in a different order, are not grouped.

User has also reported that the tags change order by themselves at times, breaking grouping, but is not sure what causes this behaviour.

RoG4FrDrD5

### 💯 Expected behavior
Entries with the same tags should be grouped together, regardless of tag order.

Also, tags shouldn't change order by themselves, but this will probably be a non-issue when ☝️ is fixed.

### 🔨 Steps to reproduce

1. Create a time entry with three tags
2. Continue the time entry
3. In the continued entry, delete the last two tags and add them again in a different order

### 📦 Additional info
[Case](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/3155115/conversations/22980780522) - includes detailed video from user

skel35:
### 📒 Description
Sorts the list of tags before using them in the group name.

### 🕶️ Types of changes

### 👫 Relationships
Closes #3133

### 🔎 Review hints
See #3133

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -687,7 +687,7 @@ void Context::updateUI(const UIElements &what) {
                     ss << date_header << te->Description()
                        << te->PID() << te->TID()
                        << te->ProjectGUID()
-                       << te->Billable() << te->Tags();
+                       << te->Billable() << te->TagsHash();
                     std::string group_name = ss.str();
 
                     group_header_id[group_name] = i;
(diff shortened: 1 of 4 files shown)


IndrekV:
Looks good. A+ 👍 

I added a small refactoring so we don't create the same hash in multiple places in code.

Issue #3136, opened 2019-07-25
Segmentation fault on application exit


flops:
### 💻 Environment

Platform: Linux
OS Version: ArchLinux
Toggl Version: 7.4.476

### 🐞 Actual behavior

Segmentation fault on application exit

### 🔨 Steps to reproduce

1. Start application
2. Quit application via Toggl context menu

### 📦 Additional info

```
Application: TogglDesktop (TogglDesktop), signal: Segmentation fault
Using host libthread_db library "/usr/lib/libthread_db.so.1".
[Current thread is 1 (Thread 0x7f75abac0b40 (LWP 2557))]

Thread 27 (Thread 0x7f75595eb700 (LWP 2622)):
#0  0x00007f75bdae4415 in pthread_cond_wait@@GLIBC_2.3.2 () from /usr/lib/libpthread.so.0
#1  0x00007f75b8c8789c in ?? () from /usr/lib/libQt5WebEngineCore.so.5
#2  0x00007f75b8c88990 in ?? () from /usr/lib/libQt5WebEngineCore.so.5
#3  0x00007f75b8c88abf in ?? () from /usr/lib/libQt5WebEngineCore.so.5
#4  0x00007f75b8c917e6 in ?? () from /usr/lib/libQt5WebEngineCore.so.5
#5  0x00007f75b8c93171 in ?? () from /usr/lib/libQt5WebEngineCore.so.5
#6  0x00007f75b8c935a4 in ?? () from /usr/lib/libQt5WebEngineCore.so.5
#7  0x00007f75b8c89b35 in ?? () from /usr/lib/libQt5WebEngineCore.so.5
#8  0x00007f75bdade57f in start_thread () from /usr/lib/libpthread.so.0
#9  0x00007f75be1990e3 in clone () from /usr/lib/libc.so.6

Thread 26 (Thread 0x7f7559dec700 (LWP 2618)):
#0  0x00007f75be16636d in nanosleep () from /usr/lib/libc.so.6
#1  0x00007f75b45b1a04 in Poco::ThreadImpl::sleepImpl(long) () from /usr/lib/libPocoFoundation.so.62
#2  0x00007f75bfbb49e4 in Poco::Thread::sleep(long) () from /opt/lib/libTogglDesktopLibrary.so
#3  0x00007f75bfc7513c in toggl::WebSocketClient::runActivity() () from /opt/lib/libTogglDesktopLibrary.so
[...]

PR #3196, opened 2019-08-11
Fix CMake REGEX path variable from PR #3179

a-andreyev:
### 📒 Description

Fix CMake error `Error:string sub-command REGEX, mode REPLACE needs at least 6 arguments total to command` when building. It happens probably at some environments like archlinux AUR makepkg fakeroot. Add quotes to CMakeLists.txt to not lost the path argument.

### 🕶️ Types of changes

### 🤯 List of changes

`CMakeLists.txt`

### 👫 Relationships

PR #3179

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,9 +27,9 @@ if ("${TOGGL_VERSION}" STREQUAL "7.0.0")
             RESULT_VARIABLE RESULT_GIT
             OUTPUT_VARIABLE OUTPUT_GIT
             ERROR_VARIABLE ERROR_GIT)
-        string (REGEX REPLACE "^v" "" OUTPUT_GIT ${OUTPUT_GIT})
-        string (REGEX REPLACE "-([0-9]+)-.*" ".\\1" OUTPUT_GIT ${OUTPUT_GIT})
-        string (REGEX REPLACE "\n" ""  TOGGL_VERSION ${OUTPUT_GIT})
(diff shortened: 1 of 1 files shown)


MartinBriza:
Hmm, sorry for causing you trouble. I assume this is not the complete fix thought, it'll at least alleviate the issue when running cmake. 
With your fix, does the app show the right (that is, not 7.0.0 or an empty string) version in the About dialog?

a-andreyev:
Thank you for the fast reply @MartinBriza!
Hmm, you're right, I'm getting `7.0.0` currently :thinking: Will look for solution later, feel free to close my PR if I'm freezing with it :)

MartinBriza:
No, the PR is perfectly fine and it actually solves an issue for you, so I'm merging it. :) Thank you!
The issue with the version string for you is a different one - I assume the build is being run from a directory that's not a git repository, so it's impossible to determine what version you're currently building.

a-andreyev:
Got it, thanks! :) Looked at `CMakeFiles.txt` version string: I could probably ask package maintainer to pass something like `-DTOGGL_VERSION:STRING="${pkgver}"` as cmake argument during building [here](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=toggldesktop) so it won't be a problem in my case

MartinBriza:
Oh yeah, definitely, that would solve the issue in a much clearer way.
Also, while you're pinging him, you can tell him I fixed the crash he mentioned in the AUR comments ( #3136 ) :)

PR #3199, opened 2019-08-12
Adapt color for light/dark mode

### LINKED ISSUE #3182: Very hard to see dark project colours when using dark mode.
### 💻 Environment
MacOS

Platform: MacOS

### 📒 Description
When projects are created with dark colours they are very difficult to see in dark mode.

NghiaTranUIT:
### 📒 Description
This PR will try to get the visible color from the current Appearance mode of the system and the color.

### How it converts
As we don't want to manipulate the original color of the project, so it;s better to recognize if it's too dark or too light.
For instance,
- If the color is too dark (RGB <= 30) and OS is Dark Mode => Get visible color
- If the color is too light (RGB >= 225) and OS is Light Mode => Get visible color

The Visible color is calculated by reverting the color.

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Introduce basic algorithm to get the visible color from given color.
- [x] Set visible color in Project Text Field (Autocomplete, Entry Cell, Editor)
- [x] Remove old code

### 👫 Relationships
Close #3182 

### Prepare
- Set Black color for any project before testing and Set Darkmode OS => Make sure that it's difficult to recognize it

#### Test
- Check this PR, and the color is visible in Entry Cell, AutoComplete and Editor

### DIFF
--- a/src/ui/osx/TogglDesktop/Feature/TimeEntry/Views/DotImageView.swift
+++ b/src/ui/osx/TogglDesktop/Feature/TimeEntry/Views/DotImageView.swift
@@ -17,7 +17,8 @@ final class DotImageView: NSImageView {
     // MARK: Public
 
     @objc func fill(with tintColor: NSColor) {
-        let iconWithColor = icon.image(withTintColor: tintColor)
+        let visibleColor = tintColor.visibleColor()
+        let iconWithColor = icon.image(withTintColor: visibleColor)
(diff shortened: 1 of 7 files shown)


skel35:
@NghiaTranUIT what do you think about the comment above? I can actually push the changes since these are just about changing numbers, would like to know your opinion first.

PR #3203, opened 2019-08-13
Resolve conflicted TimeEntry in Timeline

### LINKED ISSUE #3187: Logic for "Resolve the conflicted Timeline Entry"
### Description
The UI is done at https://github.com/toggl/toggldesktop/pull/3050
This ticket is connecting with the last criteria in https://github.com/toggl/toggldesktop/issues/2892
```
Allow solving overlaps. Clicking on overlap shows a context menu with 3 choices:
- change first entry stop time
- change last entry start time
- create a new entry of overlap
(The texts should be improved, they should be shorter)
```

We need to elaborate more. Examples are welcome 👍 

### 💻 Environment
Platform: macOS

NghiaTranUIT:
### 📒 Description
This PR will introduce the logic the resolve the conflict of Timeline entry.

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Introduce new methods in library to override exact Start/End time (unix timestamp) of given GUID. The `Context::SetTimeEntryStop` and `Context::SetTimeEntryStart` doesn't work well since it only overrides the hour and minute (The second remains) 
- [x] Logic to resolve the conflicts 

### 👫 Relationships
Close #3187 

### 🔎 Review hints
Reproduce from https://github.com/toggl/toggldesktop/issues/3187#issuecomment-519121568

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -2935,6 +2935,34 @@ error Context::SetTimeEntryProject(
     return displayError(save(true));
 }
 
+error Context::GetTimeEntryFromGUID(const std::string GUID, TimeEntry** timeEntry) {
+    if (GUID.empty()) {
+        return displayError(std::string(__FUNCTION__) + ": Missing GUID");
+    }
+
+    TimeEntry *te = nullptr;
+
+    {
+        Poco::Mutex::ScopedLock lock(user_m_);
+        if (!user_) {
+            logger().warning("Cannot change stop time, user logged out");
+            return noError;
+        }
+        te = user_->related.TimeEntryByGUID(GUID);
+
+        if (!te) {
+            logger().warning("Time entry not found: " + GUID);
+            return noError;
+        }
+
+        if (isTimeEntryLocked(te)) {
+            return logAndDisplayUserTriedEditingLockedEntry();
+        }
+    }
+    *timeEntry = te;
+    return noError;
+}
+
 error Context::SetTimeEntryDate(
     const std::string GUID,
     const Poco::Int64 unix_timestamp) {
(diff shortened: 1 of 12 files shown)


src/context.cc
```diff
@@ -2935,6 +2935,34 @@ error Context::SetTimeEntryProject(
     return displayError(save(true));
 }
 
+error Context::GetTimeEntryFromGUID(const std::string GUID, TimeEntry** timeEntry) {
```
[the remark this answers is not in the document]
NghiaTranUIT: Fix and improve at https://github.com/toggl/toggldesktop/pull/3203/commits/dd09acf348dc11140c6e174b34cbebe5c81c0e09

NghiaTranUIT:
It's ready to review @skel35. Thank you in advance 👍

src/context.cc
```diff
@@ -2839,23 +2839,10 @@ error Context::DeleteTimeEntryByGUID(const std::string GUID) {
...
-    if (!te) {
-        logger().warning("Time entry not found: " + GUID);
-        return noError;
-    }
-
-    if (isTimeEntryLocked(te)) {
-        return logAndDisplayUserTriedEditingLockedEntry();
+    TimeEntry *te = nullptr;
+    error findError = GetTimeEntryFromGUID(GUID, &te);
+    if (findError != noError && !te) {
```
skel35: @NghiaTranUIT Can you explain why we add `&& !te` to the condition?
From my understanding, when there is some time entry but it is locked, we should not continue and should just return the error.

NghiaTranUIT: If we take a look the original implementation (in master ) 

https://github.com/toggl/toggldesktop/blob/4d662731dc10ccb91189e2a71918f141f8ae3009/src/context.cc#L2793-L2796

You can see that if `!user_` is null, the func will return `noError` immediately.

So, `!te` is make sure that the TimeEntry is not null, before proceeding.

skel35: Maybe you meant `|| !te` instead of `&& !te`?

NghiaTranUIT: wow, you're right 😮 My brain 🧠 is gone

src/context.cc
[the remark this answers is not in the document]
NghiaTranUIT: Thank for the point @MartinBriza. I revert all the change in library and use this boilerplate to make sure that it's locked properly.

Additionally, I realize we have a lot of misuse in the codebase too.
For instance: https://github.com/toggl/toggldesktop/blob/4d662731dc10ccb91189e2a71918f141f8ae3009/src/context.cc#L3003-L3064

At line 3030, we're trying manipulating the `te` outside the scope of the lock.

NghiaTranUIT:
@skel35 @MartinBriza so it's ready for reviewing 👍

Issue #3012, opened 2019-06-07
lib: Almost every method argument is passed by value


MartinBriza:
Compound-type arguments are passed to methods by value, not by reference. This has the potential to create unnecessary CPU overhead and copying of said arguments to the methods.

PR #3206, opened 2019-08-14
Make passing arguments by value into references

### LINKED ISSUE #3074: Almost all of json parsing is done by value
The issue with passing objects by value instead of reference is pretty widespread in the library. Let's look into the issue more and try to pass as much of stuff (probably not just JSON) as reference as possible.

MartinBriza:
### 📒 Description
Passing compound variables (classes) by value makes them actually copy the passed data instead of just passing a simple reference/pointer (without copying any data).
This PR changes most of value passing in the library into passing by reference.

### 🕶️ Types of changes

### 👫 Relationships
Closes #3074 #3012 

### 🔎 Review hints
Check if general functionality is preserved. This PR should also be reviewed by thorough code reading.

### DIFF
--- a/src/formatter.cc
+++ b/src/formatter.cc
@@ -28,15 +28,15 @@
 
 namespace toggl {
 
-const std::string Format::Classic = std::string("classic");
-const std::string Format::Improved = std::string("improved");
-const std::string Format::Decimal = std::string("decimal");
+const std::string &Format::Classic = std::string("classic");
+const std::string &Format::Improved = std::string("improved");
+const std::string &Format::Decimal = std::string("decimal");
 
 std::string Formatter::TimeOfDayFormat = std::string("");
 std::string Formatter::DurationFormat = Format::Improved;
 
 std::string Formatter::togglTimeOfDayToPocoFormat(
-    const std::string toggl_format) {
+    const std::string &toggl_format) {
     if ("h:mm A" == toggl_format) {
         return "%h:%M %A";
     }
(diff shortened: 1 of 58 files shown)


MartinBriza:
The concept of mutable and immutable variables doesn't really work that way in C++.
By passing the strings by value, you only increase the performance cost of function calls. There is no benefit regarding thread safety - the program has to access the memory of the original string at some point at least once. And yes, there's the chance some other thread will modify the memory location of the string but this is not something passing by value would solve - that's what mutexes and other synchronization primitives are for.

NghiaTranUIT:
@MartinBriza Can you review this branch from master? Your master is pretty old and it has some macOS-related bug, which fixed in last master. Then I could continue checking the logic in macOS side 👍

MartinBriza:
Huh? This branch still merges against `master` just fine. You can just rebase it for merging.

src/formatter.cc
```diff
@@ -28,15 +28,15 @@
 
 namespace toggl {
 
-const std::string Format::Classic = std::string("classic");
-const std::string Format::Improved = std::string("improved");
-const std::string Format::Decimal = std::string("decimal");
+const std::string &Format::Classic = std::string("classic");
+const std::string &Format::Improved = std::string("improved");
+const std::string &Format::Decimal = std::string("decimal");
```
skel35: does it make sense to use references here?

MartinBriza: Nice catch! While not breaking anything, it doesn't really make sense, you're right. I'll get rid of it.

src/websocket_client.cc
```diff
@@ -195,14 +195,14 @@ error WebSocketClient::receiveWebSocketMessage(std::string *message) {
...
-    } catch(const std::string& ex) {
+    } catch(const std::string & ex) {
         return error(ex);
     }
     *message = json;
     return noError;
 }
 
-const std::string kPong("{\"type\": \"pong\"}");
+const std::string &kPong("{\"type\": \"pong\"}");
```
skel35: same here

IndrekV:
I'm very curious to see numbers after this is ready and merged. How the performance or memory allocation is compared between the old solution and this update.

IndrekV:
I tested general functionality on Mac and Linux and everything works. We now have historic moment of 3 approvals for a PR. Well done @MartinBriza and the reviewers!

PR #3213, opened 2019-08-15
Tab to focus on "Add tag"

### LINKED ISSUE #3210: [BUG] Couldn't tab to Tag Text Field in Editor
### Description
Some users report that they couldn't focus on Tag Text Field when pressing the Tab button in Editor unless they sets "Focus on all control" in Keyboard Preference.

We should change the Tag button -> TextField -> So the user could focus by default setting.

### 💻 Environment
Platform: macOS
Toggl Version: Latest stable 7.4.484

### 🐞 Actual behavior
- Couldn't focus on Tag Text field if the Keyboard Setting is "Textbox and List only"

### 💯 Expected behavior
- Should be able to focus regardless of keyboard settings

### 🔨 Steps to reproduce
1. Select "Textbox and List only" in Keyboard Preference -> Shortcut tab
2. Open Editor and try to focus
3. Couldn't focus on Tag Text Field

### LINKED ISSUE #3212: Tabbing for Mac app, keyboard preference
### 💻 Environment

Mac OS

### 📒 Description
As requested by Nghia - when tabbing on the description, project etc inputs, is the skipping of the tags tab intended?

"Technically, the Tag Input isn’t a TextField, so by default, when we tap, the Tag Input doesn’t focus until we change the setting in Keyboard Preference.

I will improve it, and ensure that it should work regardless of keyboard’s option"

NghiaTranUIT:
### 📒 Description
The current design of "Add tag" looks like the TextField, but we couldn't tab to focus on it by default Keyboard's focus option. (Keyboard Preference -> Shortcut -> Text boxes and List only)

This PR will fix it, by replacing the NSButton by NSTextField, which works by default.

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Replace Add tag button with TextField
- [x] Improve to remain the current logic

### 👫 Relationships
Close #3212 
Close #3210 

#### Text boxs and List only
1. Select "Text-boxes and List only" in Keyboard Preference 
2. Able to tab to "Add button" and navigate around those text field
3. Old logic remains 

#### All controls
1. Select "All controls" in Keyboard Preference 
2. Able to tab to "Add button" and navigate around those text field
3. Old logic remains 

### Limitation
If we select "Text-boxes and List only", it's impossible to focus to Tag Token View (which is a NSButton) unless we select "All controls". It's the default behavior of macOS.

I tried to use TextField as a Tag Token View, but it has a problem that we couldn't hide the blind cursor when it's focused (see GIF)

### DIFF
--- a/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/AddTagButton.swift
+++ b/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/AddTagButton.swift
@@ -7,7 +7,6 @@
 //
 
 import Cocoa
-import Carbon.HIToolbox
 
 protocol AddTagButtonDelegate: class {
     func shouldOpenTagAutoComplete(with text: String)
@@ -18,9 +17,7 @@ final class AddTagButton: NSTextField {
     // MARK: Variables
 
     weak var keyboardDelegate: AddTagButtonDelegate?
-    private let ignoreKeys = [kVK_Tab,kVK_Space] // Tab and space
-
-    var cursor: NSCursor? {
+    private var cursor: NSCursor? {
         didSet {
             resetCursorRects()
         }
@@ -28,6 +25,7 @@ final class AddTagButton: NSTextField {
 
     override func awakeFromNib() {
         super.awakeFromNib()
+        delegate = self
         cursor = .pointingHand
     }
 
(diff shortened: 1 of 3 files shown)


skel35:
When there are many tags and `...` tag appears, I can't tab into it, not can I skip it.
ManyTags_Tab

PR #3223, opened 2019-08-16
Disable return key in AutoCompleteTextField (mac)

### LINKED ISSUE #3222: Beep sound when selecting the AutoComplete Item by hitting Enter
### 💻 Environment
Platform: macOS
Toggl Version: master

### 🐞 Actual behavior
The beep sound appears when hitting Enter key

### 💯 Expected behavior
- No beep sound

### 🔨 Steps to reproduce

1.  Open Editor
2. Select Description in AutoComplete by enter key
3. The sound beep occurs

NghiaTranUIT:
### 📒 Description
Disable the beep sound when returning the string in Editor view

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Disable the beep sound of return key

### 👫 Relationships
Close #3222

### 🔎 Review hints
Like #3222

### DIFF
--- a/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/AutoComplete/Input/AutoCompleteTextField.swift
+++ b/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/AutoComplete/Input/AutoCompleteTextField.swift
@@ -189,6 +189,10 @@ class AutoCompleteTextField: UndoTextField, NSTextFieldDelegate, AutoCompleteVie
         autoCompleteDelegate?.autoCompleteDidTapOnCreateButton(self)
     }
 
+    func shouldClose() {
+        closeSuggestion()
+    }
+
     override func performKeyEquivalent(with event: NSEvent) -> Bool {
         // Prevent beep sound
         if currentEditor() != nil {
--- a/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/AutoComplete/Input/AutoCompleteTextField.swift
+++ b/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/AutoComplete/Input/AutoCompleteTextField.swift
@@ -188,6 +188,19 @@ class AutoCompleteTextField: UndoTextField, NSTextFieldDelegate, AutoCompleteVie
     func didTapOnCreateButton() {
         autoCompleteDelegate?.autoCompleteDidTapOnCreateButton(self)
     }
+
+    override func performKeyEquivalent(with event: NSEvent) -> Bool {
+        // Prevent beep sound on return key
+        if let key = Key(rawValue: Int(event.keyCode)) {
+            switch key {
+                case .enter:
+                return true
+            default:
+                break
+            }
+        }
+        return super.performKeyEquivalent(with: event)
+    }
 }
 
 // MARK: Private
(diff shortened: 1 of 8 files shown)


skel35:
The beep did get fixed, but I've also found a beep when tabbing from the Tags autocomplete, perhaps you could fix it as well?
Beep_TabFromTags

src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/AutoComplete/Input/AutoCompleteTextField.swift
```diff
@@ -188,6 +188,19 @@ class AutoCompleteTextField: UndoTextField, NSTextFieldDelegate, AutoCompleteVie
     func didTapOnCreateButton() {
         autoCompleteDelegate?.autoCompleteDidTapOnCreateButton(self)
     }
+
+    override func performKeyEquivalent(with event: NSEvent) -> Bool {
+        // Prevent beep sound on return key
+        if let key = Key(rawValue: Int(event.keyCode)) {
```
NghiaTranUIT: Look like I have to find better solution to prevent a beep sound. 

Basically, if I return true if the key is tab (like what I did for the Enter key), it leads to weird behavior when I could't tab to move to next view in some controls (even it isn't a AutoComplete TextField). I investigated it, and it turns out that whenever the EditorController receives the key, it will pass this event to all sub-views. If we return true, it will skip the action.

NghiaTranUIT:
@skel35 it's ready for reviewing now 👍 

Technically, I add extra steps to prevent handling the "Tab action" from other controls. It works well now

skel35:
Cannot Tab from "Create new tag" button
Tags_Tabbing

src/ui/osx/TogglDesktop/TimeEntryListViewController.m
```diff
@@ -710,4 +714,21 @@ - (void)deselectAllTimeEntryNotification
 	[self.collectionView deselectAll:self];
 }
 
+- (void)windowDidBecomeKeyNotification:(NSNotification *)notification
+{
+	// Don't focus on Timer Bar if the Editor is presented
+	if (self.timeEntrypopover.isShown)
```
NghiaTranUIT: So, I fixed @skel35  issue by https://github.com/toggl/toggldesktop/pull/3223/commits/2c306ce440458d5e09a952aef5d94b2f760e86c8.

Basically, the logic "Always focus on Timer Bar when the app become key", which we've introduced recently causes a lot of problem with Time Entry Editor. For instance, when we close the AutoComplete (is separated NSWindow), the app will become Key, and make a app focus on Timer Bar -> Causes lost Focus, and couldn't Tab to move to next controls.

Therefore, I change a logic a bit:
- Shouldn't focus on the Timer Bar if the TE Editor is presented -> It make more sense, because the user's editing the TE.
- Only Focus on Timer Bar if the Editor is not presented, and come from background.

By following this new logic. I resolve the bug.

skel35:
@NghiaTranUIT Found one more thing, sorry 😄 
`Esc` key does not work when the focus is on empty tags TextField. Works fine on all other text fields.
image

skel35:
Checked out `master` and the bug is present there as well. We'll merge this PR and open a new issue for the `Esc` with Tags text field.

PR #3237, opened 2019-08-21
Timeline: Activity

### LINKED ISSUE #3018: Timeline UI - activities list
### 💻 Environment
Platform: macOS

### 📒 Description
List of data collected with the timeline

- Vertical list with items that represent 15 minutes of data.
- Every hour has 4 vertical items
- Items are always blue
- The size and color depend on the activity during this 15 minute period. Data is given as time in seconds. The less it is the lighter is the blue color.
- Hovering items will show all info about the data tracked in this 15-minute timeslot.

Both lists are connected and are scrollable together. (#3017 entries list)

NghiaTranUIT:
### 📒 Description
This PR introduces the Activity timeline activity.

### Problems
~@IndrekV please help me how to get the Timeline Event in library. For now, I couldn't get any thing from our current implementation at~ https://github.com/toggl/toggldesktop/blob/ee8d8ebfb5b7b31b51d2b558516ced87d3926fc5/src/gui.cc#L508

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Get Activity, Event and Sub-Event from Library
- [x] Render on Timeline
- [x] Support Hover view on Activity: Render event and sub-events in stackview
- [x] Calculate the suitable shade blue color of activity: depend on the total duration of events.
- [x] Empty label state
- [x] Color from light to dark, depend on the total duration of sub-events
- [x] Find the way to convert the duration (in second) from Timeline Library to string 
- [x] Test on Real Activity data

### 👫 Relationships
Closes #3018 

### 🔎 Review hints
1. Start the app and Toggl The Timeline Activity
2. Dont' stop the app, just play around in Safari, Xcode, or whatever you like.
3. Open the Database at /Users/nghiatran/Library/Application Support/Kopsik/development/kopsik.db, and check the timeline_events table to see the events are recored 
4. Go back to the app, switch different day and go back to current day
5. The Timeline Activity should be updated and render correctly 💯 

### Known-issues
- There is no way to update the timeline event in realtime. We have to switch to previous day and go back to get the update data.

### DIFF
--- a/src/ui/osx/TogglDesktop/test2/TimelineDisplayCommand.m
+++ b/src/ui/osx/TogglDesktop/test2/TimelineDisplayCommand.m
@@ -31,6 +31,14 @@
 		{
 			self.fileName = @"";
 		}
+		if (view->DurationString)
+		{
+			self.durationStr = [NSString stringWithUTF8String:view->DurationString];
+		}
+		else
+		{
+			self.durationStr = @"";
+		}
 		self.duration = view->Duration;
 		self.isHeader = view->Header;
 
@@ -65,6 +73,14 @@
 		else
 		{
 			self.startedTimeString = @"";
+		}
+		if (view->EndTimeString)
+		{
+			self.endedTimeString = [NSString stringWithUTF8String:view->EndTimeString];
+		}
+		else
+		{
+			self.endedTimeString = @"";
 		}
 
 		// Events
--- a/src/ui/osx/TogglDesktop/test2/TimelineDisplayCommand.m
+++ b/src/ui/osx/TogglDesktop/test2/TimelineDisplayCommand.m
@@ -140,7 +140,7 @@
 		self.start = startDay;
 		self.end = endDay;
 		self.timeEntries = [[timeEntries reverseObjectEnumerator] allObjects];
-		self.activities = [activities copy];
+		self.activities = [[activities reverseObjectEnumerator] allObjects];
 	}
 	return self;
 }
--- a/src/ui/osx/TogglDesktop/test2/TimelineDisplayCommand.m
+++ b/src/ui/osx/TogglDesktop/test2/TimelineDisplayCommand.m
@@ -10,19 +10,6 @@
 #import "TimeEntryViewItem.h"
 
 @implementation TimelineDisplayEvent
-- (instancetype)initWithTitle:(NSString *)title fileName:(NSString *)filename duration:(NSTimeInterval)duration isHeader:(BOOL)isHeader
-{
-	self = [super init];
-	if (self)
-	{
-		self.title = title;
-		self.fileName = filename;
-		self.duration = duration;
-		self.isHeader = isHeader;
-	}
-	return self;
-}
-
 - (instancetype)initWithView:(TogglTimelineEventView *)view
 {
 	self = [super init];
@@ -46,6 +33,17 @@
 		}
 		self.duration = view->Duration;
 		self.isHeader = view->Header;
+
+		// Load app subevents
+		NSMutableArray<TimelineDisplayEvent *> *subEvents = [@[] mutableCopy];
+		TogglTimelineEventView *sub_it = view->Event;
+		while (sub_it)
+		{
+			TimelineDisplayEvent *event = [[TimelineDisplayEvent alloc] initWithView:sub_it];
+			[subEvents addObject:event];
+			sub_it = sub_it->Next;
+		}
+		self.subEvents = [subEvents copy];
 	}
 	return self;
 }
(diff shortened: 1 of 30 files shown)


src/ui/osx/TogglDesktop/test2/TimelineDisplayCommand.m
```diff
@@ -31,11 +109,22 @@ - (instancetype)initWithOpen:(BOOL)open
 			it_entry = it_entry->Next;
 		}
 
+		// Get activity
+		NSMutableArray<TimelineDisplayActivity *> *activities = [@[] mutableCopy];
```
NghiaTranUIT: Here is all activity in library will be converted to objc models.

NghiaTranUIT:
It seems to work now if I remove the `!event->Chunked()` in

https://github.com/toggl/toggldesktop/blob/eb7a1e930558d74d72eca555b886ef3bc9778e8a/src/user.cc#L1431

@IndrekV do we the reason behind the scene why we ignore the chunk event?

Screen Shot 2019-08-29 at 14 18 18

NghiaTranUIT:
Screen Shot 2019-08-29 at 16 24 56
Screen Shot 2019-08-29 at 16 24 52

Basically, it's ready for review 👍

NghiaTranUIT:
Separated ticket for known-issues at https://github.com/toggl/toggldesktop/issues/3269

IndrekV:
Is it possible to make the timeline activity popup hoverable. Currently it is not possible to see the full text of items that are longer and cutoff. If the blue item loses mouseover the popup is closed. It would make sense to be able to hover the popup and then maybe have a tooltip to see the full text or extend the popup to be more wide to show more content. (this has to have maximum width at the text can be pretty long)

See this screenshot of cutoff text
image

NghiaTranUIT:
Sure, I will make it growable depend on the size of text (with maximum width)

IndrekV:
I'm currently seeing no updates of data. Even if I change date to previous and back. 
time is 42 and it only shows a few minutes of data.

Screenshot 2019-08-30 at 12 43 17

Or do I have to wait for the 15 minutes to be over to see the whole 15 minute chunk of data. Anyway the few minutes of data looks strange.

NghiaTranUIT:
## Changelogs
- [x] Change the behavior of Timeline Activity and TimeEntry Popover when hovering => Popover will be remained when the mouse goes outside => Able to see the content
- [x] Dynamically resize the Timeline Activity Popover depend on inner content. Max width at 500px
- [x] Don't remove the Timeline_event before 9 days. The the past, all **uploaded** events are removing from database as soon as the app starts. => Keeps it to present the data.
- [x] Re-render the TimelineUI when there is new Timeline Event. Fixed #3269

## GIF

NghiaTranUIT:
It's ready to review 👍

IndrekV:
Great job so far @NghiaTranUIT 
Timeline seems to be rendering ok and updating also. One issue that I found was that the slot title is not updated. It shows `2:30PM - 2:31PM` although it has already tracked more than 1 minute of activity. It seems to always show that the time slot is 1 minute when it should be 15minutes.

See Screenshot for reference:
Screenshot 2019-09-06 at 14 42 05

NghiaTranUIT:
Fixed Timeline UI Improvements - Scroll to right now  https://github.com/toggl/toggldesktop/issues/3287 with https://github.com/toggl/toggldesktop/pull/3237/commits/eafccc579a94ca1d492d5896aa6604dcb5142e8f and https://github.com/toggl/toggldesktop/pull/3237/commits/4d9c2b8b998a29c22ec59791ab0625163aeabeb4

NghiaTranUIT:
@IndrekV From the current design, the Start / End time title is always 15 mins chunk time. IMO, it make no sense if my time event is small, but the title keep show 15mins

I change the logic a bit to present that StartTime - EndTime is actually the **total duration** of sub-events (see the screenshot)

Screen Shot 2019-09-09 at 14 34 08

Screen Shot 2019-09-09 at 14 41 42

How do you think?

IndrekV:
It should always show the size of the time chunk (15 minutes). This is so because it indicates the period when the time was tracked. With the current solution for example `18:45 -18:52`It seems that the activity was tracked during this 7 minute period but actually it is tracked sometime during the time chunk `18:45 - 19:00`.

I agree that it is good to see the total duration but in some other place not the chunk title.

Maybe add a row under the time chunk title saying `total duration 03:33 min` or something similar.

NghiaTranUIT:
I've changed the logic. It's ready for final review @IndrekV  👍

IndrekV:
Great Job @NghiaTranUIT !

PR #3252, opened 2019-08-23
Set the default size of the window (win)

### LINKED ISSUE #3251: Specify default window size
### 💻 Environment
Platform: Windows
Toggl Version: 7.4.527

### 🐞 Actual behavior
The default window size is set to 60% of the screen width and height (the default WPF logic).

### 💯 Expected behavior
The default window size should be 510x330. It has previously been a default size, specified by in the `LoginView` as a minimum size (470x330 + 40px top bar).

### 🔨 Steps to reproduce
1. Uninstall the app and clean all local data.
2. Install 7.4.527 and run.

### 📦 Additional info
First, the library started returning default window size values as `Int64.MinValue` instead of zero in some situations. This caused crashes on Windows and was fixed by validation of window sizes #3171.
But previously, when the size was being set to zero, it resulted in the size equal to the minimum size of the window. #3171 changed this logic and caused the size to be set to default WPF size, which is 60% of the screen size and looks really bad for our application.
image

My suggestion is to reset the size to 470x330 every time the valid size values can't be restored from the app. I'm against setting the size to 0x0 since it will look bad when the user has already had an app installed and the default screen is a Timer view. In this case, the main app window will shrink and look almost like a mini-timer.

skel35:
### 📒 Description
Always set a proper default window size if there is no valid saved window size in the database.

### 🕶️ Types of changes

### 👫 Relationships
Closes #3251 

### 🔎 Review hints
See #3251

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/utilities/Utils.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/utilities/Utils.cs
@@ -41,7 +41,9 @@ public static class Utils
         }
         else
         {
-            Toggl.Debug("Failed to retrieve window location and size");
+            mainWindow.Width = 330;
+            mainWindow.Height = 510;
+            Toggl.Debug("Failed to retrieve window location and size. Setting the default size.");
         }
 
         // First try to shift the window onto the bounding box of visible screens


src/ui/windows/TogglDesktop/TogglDesktop/utilities/Utils.cs
```diff
@@ -41,7 +41,9 @@ public static void LoadWindowLocation(Window mainWindow, EditViewPopup editPopup
         }
         else
         {
-            Toggl.Debug("Failed to retrieve window location and size");
+            mainWindow.Width = 330;
```
NghiaTranUIT: The default height and width look good to me.

Do you think we should reset the top left of the window to zero too? So in worst case, the x y is out of screen, and it could reset to the origin position.

PR #3256, opened 2019-08-26
Load More in Timeline

### LINKED ISSUE #3190: Load more in Timeline view
### 💻 Environment
Platform: macOS

### 📒 Description
In Timeline view when moving a few days backward there is a situation when the UI says `No entries here... Go ahead and track some time` while in reality there is time tracked for that days, the entries just haven't been loaded. 
We should perform Load More action when the Load More button is visible in the List view and when we have moved back to the day for which there are no entries loaded.

NghiaTranUIT:
### 📒 Description
This PR introduce the fix to load "Load More" when navigating around in Timeline

### 🕶️ Types of changes
 **Improve** (non-breaking change which fixes an issue)

### 🤯 List of changes
- [x] Call LoadMore in Library when changing date in Timeline.

### 👫 Relationships
Close #3190 

### 🔎 Review hints
1. Start the app with account has bunch of old TE
2. Make sure the TE List has Load More button
3. Switch to Timeline and change to the day has LoadMore
4. The Timeline could get old TimeEntry -> 💯 
5. Back to TE List -> The Load More btn is disappear -> 💯

### DIFF
--- a/src/ui/osx/TogglDesktop/Feature/TimeEntry/TimeEntryDatasource.swift
+++ b/src/ui/osx/TogglDesktop/Feature/TimeEntry/TimeEntryDatasource.swift
@@ -98,6 +98,17 @@ class TimeEntryDatasource: NSObject {
     
     // MARK: Public
 
+    @objc func loadMoreTimeEntryIfNeed(at date: Date) {
+        guard isShowLoadMore,
+            let lastDate = lastDateInSections() else { return }
+
+        // If the date of timeline is before the last day and the Load More btn is showing
+        // We should load more item
+        if date < lastDate {
+            DesktopLibraryBridge.shared().loadMoreTimeEntry()
+        }
+    }
+
     func selectFirstItem() {
         // Check if there is a least 1 item
         guard sections.first?.entries.first != nil else { return }
@@ -271,6 +282,11 @@ extension TimeEntryDatasource {
         collectionView.register(NSNib(nibNamed: Constants.TimeHeaderNibName, bundle: nil),
                                 forSupplementaryViewOfKind:NSCollectionView.elementKindSectionHeader,
                                 withIdentifier: Constants.TimeHeaderID)
+    }
+
+    fileprivate func lastDateInSections() -> Date? {
+        let lastSection = sections.reversed().first { !$0.isLoadMore }
+        return lastSection?.entries.last?.ended
     }
 }
 
--- a/src/ui/osx/TogglDesktop/Feature/TimeEntry/TimeEntryDatasource.swift
+++ b/src/ui/osx/TogglDesktop/Feature/TimeEntry/TimeEntryDatasource.swift
@@ -286,7 +286,7 @@ extension TimeEntryDatasource {
 
     fileprivate func lastDateInSections() -> Date? {
         let lastSection = sections.reversed().first { !$0.isLoadMore }
-        return lastSection?.entries.last?.ended
+        return lastSection?.entries.last?.started
     }
 }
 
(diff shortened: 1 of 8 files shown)


NghiaTranUIT:
## Problem
- I've checked the Library, and from what I know, that we don't have any knowledge about what the day for Load More button is. 
- All we do in library is `Always` add the Load More button in the end of TE List at the **launch**.

## Solution
- Call the "Load More" whenever we change the date in Timeline.

@IndrekV do we have better approach 🤔 ?

IndrekV:
Load more button loads older entries. Up to 1 month old I think, so excuting it on every date change is not needed. It is only needed to execute load more in scenario where user has not pressed "Load more" in the listing view and goes to timeline view and moves to the past to the point where there is a day that does not have entries loaded. At that point we should load the older entries.

NghiaTranUIT:
I suppose that we shouldn't do it in macOS side. since the LoadMore has safety guard here 

https://github.com/toggl/toggldesktop/blob/b11236614b92445a5388b72fe8f819860e0f86f8/src/context.cc#L4478-L4481

So, even if we call it's multiple time, the Load More only works **one** time.

IndrekV:
The safeguard is correct you actually only need to do "Load more" once per app lifetime. As the days go by and you don't close the app you will have all data available already loaded.

Basically, you can never see data past the 30 days on the desktop app.

NghiaTranUIT:
I've updated the logic on macOS side, since it's straightforward. From now, the Timeline will only call "Load More" if the Timeline's current date is earlier than the last day of Time Entry 👍

src/ui/osx/TogglDesktop/Feature/TimeEntry/TimeEntryDatasource.swift
```diff
@@ -272,6 +283,11 @@ extension TimeEntryDatasource {
                                 forSupplementaryViewOfKind:NSCollectionView.elementKindSectionHeader,
                                 withIdentifier: Constants.TimeHeaderID)
     }
+
+    fileprivate func lastDateInSections() -> Date? {
+        let lastSection = sections.reversed().first { !$0.isLoadMore }
```
skel35: can you use `last` instead of `reversed().first`?

src/ui/osx/TogglDesktop/Feature/TimeEntry/TimeEntryDatasource.swift
```diff
@@ -272,6 +283,11 @@ extension TimeEntryDatasource {
...
+        return lastSection?.entries.last?.ended
```
skel35: the day of time entry is assigned by the start time AFAIK, perhaps using `started` instead of `ended` is better here?

src/ui/osx/TogglDesktop/Feature/TimeEntry/TimeEntryDatasource.swift
NghiaTranUIT: Actually, I deliberately choose `reversed().first {where: {}}` since it only costs two if-comparison steps at maximum. 

Additionally, The O(n) of `reversed()` is 1, since it doesn't reverse entire the array. it's just a reversed iterator.
Ref: https://developer.apple.com/documentation/swift/array/1690025-reversed

---------
If we use `last {where: {}}`, it will iterates from the beginning and find the item at end of the list => O(n) = n, which is unnecessary.

src/ui/osx/TogglDesktop/Feature/TimeEntry/TimeEntryDatasource.swift
NghiaTranUIT: It's good point. I've just updated 👍

PR #3271, opened 2019-08-30
DnD for notification and Idle popup 

### LINKED ISSUE #3268: Respect notification center settings and "Do not disturb" when triggering notifications
### 💻 Environment
Platform: macOS

### 📒 Description
We should always take system notifications setting into account when triggering user notifications and sounds. We handled sounds with #3082 

Now we need to address all the user notifications we are showing
All the places we use `UserNotificationCenter`

- [ ]  tracking reminder
- [ ]  pomodoro timer
- [ ]  pomodoro break timer
- [ ]  autotracker

NghiaTranUIT:
### 📒 Description
This PR will fix the bug that the idle popup doesn't respect the Do not Disturb setting.

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Detect the status of DnD setting
- [x] Prevent the Idle Popup appears when DnD is on
- [x] Present it again if DnD is off

### 👫 Relationships
Closes #3268

#### Notifications on DnD
1. Moving the mouse to top-right of screen -> Turn on DnD
2. Start any TE
3. Wait for Pomodoro or Reminder notification 
4. There is no one appear -> 💯 
#### Idle Popup on DnD
1. Turn on DnD
2. Start any TE, and set Idle time in Setting is 1 min
3. Leave the app after 1 mins
4. After 1 mins, there is no popup 
5. If we turn OFF DnD and come back to the app
6. Idle Dnd must come up -> 💯 
7. User can select those idle options as usual -> 💯 

#### Notification and idle when DnD is off
- Both notifications and idle should works as normal if the DnD is off.

### DIFF
--- a/src/ui/osx/TogglDesktop/test2/AppDelegate.m
+++ b/src/ui/osx/TogglDesktop/test2/AppDelegate.m
@@ -1430,7 +1430,6 @@ const NSString *appName = @"osx_native_app";
 	NSAssert([NSThread isMainThread], @"Rendering stuff should happen on main thread");
 
 	[self.idleNotificationWindowController displayIdleEvent:idleEvent];
-	[NSApp activateIgnoringOtherApps:YES];
 }
 
 - (PLCrashReporter *)configuredCrashReporter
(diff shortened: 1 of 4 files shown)


IndrekV:
👍

PR #3280, opened 2019-09-02
Sign up with Google on Windows

### LINKED ISSUE #3064: Add "Sign up with Google" button
### 💻 Environment
Platform: Windows

### 📒 Description
Add "Sign up with Google" button below the Sign up" button.

"Select Country" should be a required field and "I agree to terms of service and privacy policy" should be checked. E-mail and password fields are not used.

### ⭐️ Why do you want this?
We should allow our users to sign up with Google directly from the desktop app.

skel35:
### 📒 Description
Adds Sign up with Google button in the Windows app

### 🕶️ Types of changes

### 🤯 List of changes
- Added Sign up with Google to the library
- Added Sign up with Google button in the Windows app
- Added proper handling of expired token errors

### 👫 Relationships
Closes #3064.

### 🔎 Review hints
You can delete one of your existing Toggl accounts in case you don't have a Google account not already tied to a Toggl one.

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -5460,9 +5460,12 @@ error Context::signupGoogle(
         ws["initial_pricing_plan"] = 0;
         user["workspace"] = ws;
 
+        std::stringstream ss;
+        ss << "/api/v9/signup?app_name=" << TogglClient::Config.AppName;
+
         HTTPSRequest req;
         req.host = urls::API();
-        req.relative_url = "/api/v9/signup?app_name=kopsik";
+        req.relative_url = ss.str();
         req.payload = Json::StyledWriter().write(user);
 
         HTTPSResponse resp = toggl_client->Post(req);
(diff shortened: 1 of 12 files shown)


NghiaTranUIT:
Find this bug when testing this PR https://github.com/toggl/toggldesktop/issues/3283

src/context.cc
```diff
@@ -2273,6 +2273,29 @@ error Context::Signup(
     return Login(email, password);
 }
 
+error Context::GoogleSignup(
+    const std::string access_token,
+    const uint64_t country_id) {
+
+    TogglClient client(UI());
+    std::string json("");
```
NghiaTranUIT: Since we don't use JSON in this func. I suggest we could get rid of it.

skel35: We use it on the next line as the third parameter to `signupGoogle` method.
I'd say let's leave it like that as it's quite readable and we're using it like this in other places.

PR #3289, opened 2019-09-06
Process Tag in background

### LINKED ISSUE #2787: App crash if there are enormous tags
### 📖 Description
There are certain users who are encountering the crash on TogglDesktop if there are enormous tags in their account.
([conversation](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/2098566/conversations/20567021731))

### 💻 Environment
Platform: macOS
Toggl Version: 7.4.336

### 🐞 Actual behavior
- The app crashed unexpectedly if there are tons of tags

### 💯 Expected behavior
- The app should be fine regardless of number of tags.

### 🔨 Steps to reproduce
1. Login to account, which has tons of tags.
2. The app 💥

NghiaTranUIT:
### 📒 Description
This PR will introduce the fix for the UI Freeze if there is a huge number of tags.

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Process the tag in background

### 👫 Relationships
Close #2787

### 🔎 Review hints
Replace `void GUI::DisplayTags(const std::vector<view::Generic> list)` with the following code
```
void GUI::DisplayTags(const std::vector<view::Generic> list) {
    logger().debug("DisplayTags");

//    TogglGenericView *first = generic_to_view_item_list(list);
    TogglGenericView *first = nullptr;
    for (int i=0; i < 10000; i++) {
        for (std::vector<toggl::view::Generic>::const_iterator
                it = list.begin();
                it != list.end();
                it++) {
            TogglGenericView *item = generic_to_view_item(*it);
            item->Next = first;
            first = item;
        }
    }

    on_display_tags_(first);
    view_item_clear(first);
}
```
It will 10x your current tags. Make sure you have 5-6, so the result is 50k or 100k.

- Start the app, and search through the tag
- Observe if the app is freeze when opening the TE -> If it's smooth -> 💯 
- Searching could cause the freeze a bit but it's reasonable.
- Check RAM. If it's around 100-200Mb -> Good to go 💯 

Screen Shot 2019-09-06 at 18 50 21
Screen Shot 2019-09-06 at 18 50 55

### DIFF
--- a/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/AutoComplete/Storage/TagStorage.swift
+++ b/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/AutoComplete/Storage/TagStorage.swift
@@ -25,10 +25,11 @@ extension Notification.Name {
     // MARK: Public
 
     func update(with viewItems: [ViewItem]) {
-        self.tags = build(from: viewItems)
-        print("==== Tag data source \(self.tags.map {$0.name})")
-        NotificationCenter.default.post(name: .TagStorageChangedNotification,
-                                        object: tags)
+        build(from: viewItems) {[weak self] tags in
+            guard let strongSelf = self else { return }
+            strongSelf.tags = tags
+            NotificationCenter.default.post(name: .TagStorageChangedNotification, object: tags)
+        }
     }
 
     func filter(with text: String) -> [Tag] {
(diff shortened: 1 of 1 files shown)


IndrekV:
Nice, no freezing UI and the background tags loading is done before the user manages to start adding tags. For the users it just feels fast.

PR #3296, opened 2019-09-09
Improve solution for bad color contrast in dark mode

skel35:
### 📒 Description
Improve solution for bad color contrast in dark mode using the notion of contrast ratio and relative color luminance.

### 🕶️ Types of changes

### 🤯 List of changes
1. Checks contrast ratio based on color's luminance.
2. If contrast ratio is very bad (as in black color on black background) it inverts it.
3. If contrast ratio is somewhat bad (as in dark blue on black background) it makes the color lighter.


skel35:
Color contrast ratio calculations based on this article https://medium.muz.li/the-science-of-color-contrast-an-expert-designers-guide-33e84c41d156 (contains links to Web Content Accessibility Guidelines at w3.org)

PR #3317, opened 2019-09-16
Cancel current Entry selection on Timer Bar

### LINKED ISSUE #3250: Add remove icon to project row in timer
### 💻 Environment
Platform: macOS

### 📒 Description
Add icon to remove the selected project from timer. For example, if user selects project and actually does not want to track against it. Currently, they would have to start the timer and then edit the project.

Sample of the button in windows.

image

NghiaTranUIT:
### 📒 Description
This PR will add a little Cancel btn in Timer bar when we select the TimeEntry from AutoComplete in TimerBar, which enable to remove it. 

This behavior likes Windows version.

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Add Cancel icon from window
- [x] Logic for cancel the current selection

### 👫 Relationships
Closes #3250 

### 🔎 Review hints
1. Type in Timer bar and select any TE by mouse in the AutoComplete View
2. TimerBar will populate the TE, and the Cancel btn appears -> 💯 
3. Able to click to reset the input -> 💯 
4. If start by keyboard, or continue any TE, the Cancel icon should be hidden -> 💯


IndrekV:
The clear button is meant to clear only the project as it is not possible to clear it the other way. The description should not be cleared as it is a text field and can be edited with ease. 

Also, the button is in front of the project row indicating that it works for this data. Clearing the description will create confusion.

IndrekV:
Cancel button should not touch the description but clear everything else (project, client, tags, billable)

Also, timer should stay in the editing state and not go back to the single-text input view.

NghiaTranUIT:
All done. Please do second round @IndrekV 🍺

PR #3258, opened 2019-08-27
[Need Help][WIP] Locally store Unsynced in database

NghiaTranUIT:
### 📒 Description
According the proposal in https://github.com/toggl/toggldesktop/issues/3128#issuecomment-525168994.

This PR tries to store unsynced state of TimeEntry in database whenever the TE is set "Unsynced" from the Syncer in Library.

It could resolve the case that when launching the app, the Unsynced state is remained and fulfill the requirement on https://github.com/toggl/toggldesktop/issues/3128#issuecomment-525180989 

### Need help
@IndrekV I couldn't find suitable way to force the TE save in database when `BaseModel::SetUnsynced()` is trigger from the Syncer. 

I need some help 🙌

### 🕶️ Types of changes
 **Bug fix** (non-breaking change which fixes an issue)

### 🤯 List of changes
- [x] Add unsynced to database (migration)
- [x] Receive and save in database when the TE is saved.
- [ ] When TE is set as "Unsynced" -> Store to database

### 👫 Relationships
Close #3128

### 🔎 Review hints
- Steps in #3128

PR #3328, opened 2019-09-18
Use UIModifiedAt and UpdatedAt comparison to detect unsynced items

### LINKED ISSUE #3128: Unsynced entry is not correctly displayed when app is relaunched
### 💻 Environment

Platform: macOS
OS Version: macOS 10.14.3
Toggl Version: 7.4.471

### 🐞 Actual behavior
Unsynced entry is shown as a synced entry if the app is relaunched.

### 💯 Expected behavior
Unsynced entry should be displayed as unsynced.

### 🔨 Steps to reproduce
1. Go offline
2. Log a time entry
3. Now It is correctly shown as unsynced (red warning sign on the bottom right)
4. Quit Toggl Desktop
5. Start Toggl Desktop
6. Now the red warning sign is not shown

### 📦 Additional info
image

IndrekV:
### 📒 Description
We use `UIModifiedAt` and `UpdatedAt` to detect if time entry has been synced.

We currently use `UIModifiedAt` to keep the timestamp of the last UI changes made locally

We currently use `UpdatedAt` to keep the timestamp of the moment data was saved to the server. 

If sync was success these should be same as we update the `UIModifiedAt` to be same as the `UpdatedAt`. If they are different this means that UI has been changed later and the changes have not been synced to the servers.

Marking the time entry `Dirty` means that it needs to be synced. Doing this right after loading the data from the database will make it sync faster

### 🕶️ Types of changes

### 👫 Relationships
Alternative solution to #3258 

Closes #3128

### 🔎 Review hints
- Go offline
- Create time entries
- See that they are marked unsynced
- Stay offline and restart app
- Unsynced items are marked as unsynced right away after app start

### DIFF
--- a/src/time_entry.cc
+++ b/src/time_entry.cc
@@ -457,8 +457,8 @@ void TimeEntry::LoadFromJSON(Json::Value data) {
     SetBillable(data["billable"].asBool());
     SetDurOnly(data["duronly"].asBool());
     SetUpdatedAtString(data["at"].asString());
+    SetUIModifiedAt(Formatter::Parse8601(data["at"].asString()));
 
-    SetUIModifiedAt(0);
     ClearUnsynced();
 }
 
--- a/src/database.cc
+++ b/src/database.cc
@@ -1969,7 +1969,7 @@ error Database::loadTimeEntriesFromSQLStatement(
                 } else {
                     model->SetValidationError(rs[19].convert<std::string>());
                 }
-                if (model->UIModifiedAt() != model->UpdatedAt()) {
+                if (model->UIModifiedAt() > model->UpdatedAt()) {
                     model->SetUnsynced();
                     model->SetDirty();
                 } else {
--- a/src/database.cc
+++ b/src/database.cc
@@ -1969,7 +1969,13 @@ error Database::loadTimeEntriesFromSQLStatement(
                 } else {
                     model->SetValidationError(rs[19].convert<std::string>());
                 }
-                model->ClearDirty();
+                if (model->UIModifiedAt() != model->UpdatedAt()) {
+                    model->SetUnsynced();
+                    model->SetDirty();
+                } else {
+                    model->ClearDirty();
+                }
+
                 list->push_back(model);
                 more = rs.moveNext();
             }


NghiaTranUIT:
The idea of this PR is good, but I unexpectedly encountered some issues:

### 1. At the first launch on this PR, all TimeEntry are marked as Unsync (Red icon in right-bottom).
After that, the list is synced and the Unsync icons are gone. I couldn't reproduce it anymore. 

I suppose that this line of code cause a problem:
https://github.com/toggl/toggldesktop/blob/2c22dd1a39610742762372004024a13dda721330/src/database.cc#L1972-L1973

As the first launch, the `UIModifiedAt` is not equal with `UpdatedAt`, so all my TEs were marked as unsync, which is incorrect.

### 2. The TE is marked as unsync when we do following steps:
1.  Start any TE
2. Stop
3. Open. Editor and set the date to yesterday or older.
4. Observer that this TE is successfully synced (No red icon)
5. Stop and open the app again
6. This TE is marked as unsynced, which is not right.

IndrekV:
**1.** The issue was that newly created time entry can have UIModifiedAt set as `0`. I've updated the if clause to do a valid check

**2.** I'm not able to replicate this. Maybe the change was really not synced before the app was closed or the last commit also fixed this issue

Anyhow Thank you for the thorough testing and feedback @NghiaTranUIT 

It's ready for another round of review!

NghiaTranUIT:
For the second issues, I realize that this behavior is correct. The reason is that I stopped the app quickly (TE wasn't sync properly), then the next launch, this TE was marked as Unsynce, and disappear after couple seconds. Thus, it's correct 👍 

I tested for old and new users and it works as we expected 🌮

src/time_entry.cc
```diff
@@ -457,8 +457,8 @@ void TimeEntry::LoadFromJSON(Json::Value data) {
     SetBillable(data["billable"].asBool());
     SetDurOnly(data["duronly"].asBool());
     SetUpdatedAtString(data["at"].asString());
+    SetUIModifiedAt(Formatter::Parse8601(data["at"].asString()));
```
skel35: @IndrekV This change broke library logic. If `ui_modified_at` is not  `0` it is marked as unsynced and `BaseModel::NeedsPUT()` returns `true`.

IndrekV: Good call, I'll look for a fix.

skel35: I've submitted a PR for this #3346, you can check out if it looks good

IndrekV: this should fix it b5a5f15b5053094d20293aa16d33adf94111bd42

PR #3338, opened 2019-09-20
[LibOverhaul 03] Update JsonCpp

MartinBriza:
### 📒 Description
To state the obvious, JsonCpp is a third party library we're using to parse Json in the library.
We were using quite an outdated version so I updated it to the latest release (1.9.1).
This also means we had to change the way we're using the library (the classes we used to use are now deprecated) so I added a helper class so we don't have to put in a lot of boilerplate code.

### 🕶️ Types of changes

### 🤯 List of changes
- JsonCpp dependency has been updated

### 🔎 Review hints
- As with most library overhaul PRs, warnings should disappear and the app should continue to compile on all platforms.


NghiaTranUIT:
Basically, this PR works well on all platforms. I double-checked all primary features and it's as we expected.

However, I encountered the crash when trying logging out. The log is pointing to JSON Parser. Therefore, I suppose we should take a look at it.

### How to reproduce
1. Check out this PR, and login
2. Start any TE and wait 10s
3. Stop and quickly log-out 
4. The crash happens 💥 

- Full stack trace: https://pastebin.com/tyLcKQpQ
- Current thread stack trace: https://pastebin.com/SBHR6ryq

### Screenshots
Screen Shot 2019-09-26 at 14 13 43

MartinBriza:
I spent a good while testing this today and it crashes on random locations. I'm starting to suspect I discovered another race condition in the library....

PR #3339, opened 2019-09-20
[LibOverhaul 04] Initialize all class members

MartinBriza:
### 📒 Description
Not explicitely initializing class member variables can lead to undefined behavior. Therefore I added explicit initialization to `nullptr` to the ones that were not initialized.

### 🕶️ Types of changes

### 🔎 Review hints


MartinBriza:
I have just rebased this PR against master and omitted the dependency on #3338.

If the build passes, I'll merge it as per Nghia's review. There's also virtually no chance this would break anything.

PR #3340, opened 2019-09-20
[LibOverhaul 05] Use the prefix increment operator

MartinBriza:
### 📒 Description
Because of the (perfectly logical) way how C++ works, there's a (very) slight peformance benefit of using the prefix increment operator (`++it`) compared to the postfix variant (`it++`) with iterators. `cpplint` complains about this so I wrote a patch to get rid of the warning.

### 🕶️ Types of changes

### 🔎 Review hints


NghiaTranUIT:
This PR is branched off from https://github.com/toggl/toggldesktop/pull/3338, and this parent has an issues, which need to be address (https://github.com/toggl/toggldesktop/pull/3338#issuecomment-535377952), so I approve and merge it later when it's done.

PR #3341, opened 2019-09-20
[LibOverhaul 06] Mark override methods

MartinBriza:
### 📒 Description
It's better to mark overriden virtual methods in child classes to avoid mixups when changing the underlying API or to just be sure that we're working with a virtual method in the child class.

### 🕶️ Types of changes

### 🔎 Review hints

### DIFF
--- a/src/autotracker.h
+++ b/src/autotracker.h
@@ -36,9 +36,9 @@ class TOGGL_INTERNAL_EXPORT AutotrackerRule : public BaseModel {
     void SetTID(const Poco::UInt64 value);
 
     // Override BaseModel
-    std::string String() const;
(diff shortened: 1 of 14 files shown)


NghiaTranUIT:
The change of this PR is https://github.com/toggl/toggldesktop/pull/3341/commits/eacb1cf5a8cb086f5e40ac709d33bbdb420286b9 and it works well in macOS / Windows by playing around the app.

However, this PR is children of the [JSON PR](https://github.com/toggl/toggldesktop/pull/3338#issuecomment-535377952), which has 1 issue. Therefore, I will approve but not merge until the host is properly addressed. 👍

PR #3363, opened 2019-09-25
Run Store app on startup

### LINKED ISSUE #3347: Run on Windows startup doesn't work for Store version
### 💻 Environment
Platform: Windows (Store version)
Toggl Version: 7.4.530

### 🐞 Actual behavior
Checking "Run on Windows startup" doesn't work for the Store version.

### 💯 Expected behavior
Checking "Run on Windows startup" should work.

skel35:
### 📒 Description
Makes "Run the app on Windows login" work for Microsoft Store version of the app.

### 🕶️ Types of changes

### 👫 Relationships
Closes #3347.

### 🔎 Review hints

- Uninstall regular Toggl Desktop app to make sure you run a Store version of the app
- In Visual Studio Solution Explorer right-click `TogglDesktop.Package` -> Set as StartUp Project
- Run a Store app (`StoreDebug` or `StoreRelease` config) from Visual Studio. It installs the app on your PC when you run it
- Verify that "Preferences->Run the app on Windows login" is unchecked by default and "Toggl Desktop" entry is Disabled in Task Manager->Startup
- Restart the PC and verify that Toggl Desktop app is not running
- Check "Preferences->Run the app on Windows login"
- Restart the PC and verify that the Store version of Toggl Desktop app is running
- Go To Task Manager->Startup and Disable "Toggl Desktop"
- Open Preferences again and verify that "Run the app on Windows login" checkbox is hidden.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop.Package/Package.appxmanifest
+++ b/src/ui/windows/TogglDesktop/TogglDesktop.Package/Package.appxmanifest
@@ -49,7 +49,7 @@
       <Extensions>
         <desktop:Extension
           Category="windows.startupTask"
-          Executable="TogglDesktop.exe"
+          Executable="TogglDesktop\TogglDesktop.exe"
           EntryPoint="Windows.FullTrustApplication">
           <desktop:StartupTask
               TaskId="TogglDesktopStartupTask"
--- a/src/ui/windows/TogglDesktop/TogglDesktop.Package/Package.appxmanifest
+++ b/src/ui/windows/TogglDesktop/TogglDesktop.Package/Package.appxmanifest
@@ -53,7 +53,7 @@
           EntryPoint="Windows.FullTrustApplication">
           <desktop:StartupTask
               TaskId="TogglDesktopStartupTask"
-              Enabled="true"
+              Enabled="false"
               DisplayName="Toggl Desktop" />
         </desktop:Extension>
       </Extensions>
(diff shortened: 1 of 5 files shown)


skel35:
Added a WIP label only for improving the Windows 10 SDK references. The idea is to use a NuGet package or something else in order to not tie to a specific version of Windows 10 SDK.

The actual functionality of this PR is done (not WIP)

skel35:
@NghiaTranUIT This is because Windows API makes it impossible to enable it from within the app after the user manually disabled it in the Task Manager.

skel35:
@NghiaTranUIT If the option would be disabled but not hidden, users might ask questions about why it is disabled. Simply hiding the option seems more reasonable to me.

NghiaTranUIT:
It make sense 👍 I approve this PR.

When we fix the Windows 10 SDK issues. WE could merge it 🌮

skel35:
The results of my investigation into referencing Windows 10 SDK:
- The app will work for the user as long as the user's Windows version has the API that is being called. It is easy to do a version check before calling a certain API or to use APIs that are supported by the minimum target version.
- You can add any Windows 10 SDK references from the range `[Minimum Target version; Target version]`. The recommended way is to reference the `Target version` in order to be able to use the API from the latest possible version. Developers should have installed the SDK with the referenced version in order to build such an app.
- It's better to disable all Windows 10 API usages (incl. `using` statements) from the regular app compilation to make sure there's no need to install any Windows 10 SDK for developers who don't want to build a desktop.

For now, I'll keep the reference to .16299 version as we don't use newer APIs anyway. I'll bump the version whenever we have a need for it (e.g. if CI doesn't have SDK this old).

skel35:
My only commit after the review is reordering `#if` directive to make the app compilable without Windows 10 SDKs, so I feel it doesn't need further reviews and I'm merging this to `master`.

Issue #3362, opened 2019-09-25
Improvements for Continue buttons UX in the time entries list


IndrekV:
### 💻 Environment

Platform: macOS

### 📒 Description
Two issues that have been reported by our users.

- [ ] Continue button is hidden by default. 
We made the functionality the same as in the web for this. This logic might work ok in web, but it's not common in mac native apps. So let's just change it so that the continue button is always visible.

- [ ] Increase the size of the clickable area for continue button. 
Before we had a bigger button, we had a play button with a circle around it and it was easier to click it. With new design, we removed the circle and made the button smaller. We will not bring back the circle but we will increase the click area for easier click experience.

PR #3364, opened 2019-09-26
Improve the Continue btn in TimeEntry Cell

NghiaTranUIT:
### 📒 Description
This PR introduces some improvements to the Continue button to make it's easier to interact 

We made the functionality the same as in the web for this. This logic might work ok in web, but it's not common in mac native apps. So let's just change it so that the continue button is always visible.

I update the tracking option of NSTrackingArea, so we are able to see the Continue button when hovering on it (regardless of the state of the app). Don't need to show all Continue buttons.

### 🕶️ Types of changes
 **Improvements** (non-breaking change which fixes an issue)

### 🤯 List of changes
- [x] Change Tracking Option to Always to make sure the user could hover to see the Continue button even the app is inactive.
- [x] Fix the size of the Continue button and expand 20px for width and height.

### 👫 Relationships
Closes https://github.com/toggl/toggldesktop/issues/3362

#### Able to see the Continue button when the app is inactive
1. Open TogglDesktop 
2. Focus on the different app (make sure Toggl is in the same Desktop)
3. Hover mouse on TimeEntry Cell
4. Able to see the Continue btn -> 💯 
5. One-click to start the task -> 💯 

#### Able to tap on the button
1. Click on edge of button and we're able to start the TimeEntry -> 💯 

Screen Shot 2019-09-26 at 13 15 22


IndrekV:
Everything works. I agree with the solution of not making all continue buttons visible all the time but adding the logic so hovering always shows the continue icon.

Great Job and crazy fast execution! @NghiaTranUIT 👍

PR #3387, opened 2019-10-02
Improve status code 429 handling

IndrekV:
### 📒 Description
We show a human-readable error message when 429 happens. 

### 🕶️ Types of changes
**Bug fix** (non-breaking change which fixes an issue)

### 👫 Relationships

Closes #2245
Closes #3379 
Closes #3372

### 🔎 Review hints
 - Create a rate limit situation (pressing the keyboard shortcut for "new" million times)
 - Or simulate 429 some other way
 - Wait to see the error message about delayed sync

PR #3388, opened 2019-10-02
Alternative solution for handling 429 error message

NghiaTranUIT:
## Description
Simple solution to handle 429 error message in all platforms.

## Changelogs
- [x] Treat 429 response as a normal error message 
- [x] Present the error by `DisplayError` func as usual 

## Steps to review

See https://github.com/toggl/toggldesktop/pull/3387

### DIFF
--- a/src/gui.cc
+++ b/src/gui.cc
@@ -150,7 +150,7 @@ error GUI::DisplayError(const error &err) {
         if (kBackendIsDownError == err) {
             DisplayOnlineState(kOnlineStateBackendDown);
         }
-        else if (kRateLimit != err) {
+        else {
             DisplayOnlineState(kOnlineStateNoNetwork);
         }
         return err;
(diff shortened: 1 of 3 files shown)


IndrekV:
If it works the same and catches 429 and shows the error I'm all for this simpler solution. Can someone verify this works. I don't have proxy app setup on my mahcine.

src/gui.cc
```diff
@@ -149,7 +149,8 @@ error GUI::DisplayError(const error &err) {
         ss << "You are offline (" << err << ")";
         if (kBackendIsDownError == err) {
             DisplayOnlineState(kOnlineStateBackendDown);
-        } else {
+        }
+        else if (kRateLimit != err) {
```
skel35: `IsNetworkingError` for 429 returns `false` as far as I understood this, so `err` should never be `kRateLimit` at this point and we don't need this check. Am I correct? @NghiaTranUIT

PR #3396, opened 2019-10-04
Only format the TogglDeskop source (mac)

### LINKED ISSUE #3395: [BUG] Xcode-formatter bumps the size of Pod's library
### Problem
In master, I observer that the compile time increase significantly after 20 times stop / start the debug build.

The size of GTMAppAuth increase to 50Mb 💥 

Screen Shot 2019-10-04 at 16 41 41

After the investigation, It turns out that `Xcode-formatter` try to format all Pod's libraries and cause this issues.

### 💻 Environment
Platform: mac

### 🐞 Actual behavior
- The size of GTMAppAuth increase everytime we run the app in Debug mode (Xcode)

### 💯 Expected behavior
- Shouldn't run theXcode-formatter
- Shouldn't increase the size

### 🔨 Steps to reproduce

1. Start / Stop 20 times
2. Observe the size of GTMAppAuth

NghiaTranUIT:
### 📒 Description
This PR will change the direction of xcode-formatter.

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Only format the TogglDesktop's source code

### 👫 Relationships
Closes #3395 

### 🔎 Review hints
1. Start / Stop in Debug mode 20 times
2. Check the size of those files in `./src/ui/osx/Pods/GTMSessionFetcher/Source/*`
3. If the size is not change -> 💯 
4. Check the build log -> Make sure xcode-formatter doesn't format those Pod's libraries

Screen Shot 2019-10-04 at 16 45 19

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -210,7 +210,7 @@ fmt_lib: third_party/google-astyle/build/google-astyle
 	third_party/google-astyle/build/google-astyle -n src/ui/windows/TogglDesktop/TogglDesktopDLLInteropTest//*.cs
 
 fmt_ui:
-	./third_party/Xcode-formatter/CodeFormatter/scripts/formatAllSources.sh src/ui/osx/
+	./third_party/Xcode-formatter/CodeFormatter/scripts/formatAllSources.sh src/ui/osx/TogglDesktop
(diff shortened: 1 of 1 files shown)


IndrekV:
I'm not sure why formatter was increasing the file size from a few kb to 100 mb but it is fixed now. I also noticed that building was superslow when code formatting these pods

Issue #3095, opened 2019-07-18
Google Login claims invalid password in Linux Mint


MartinBriza:
### 💻 Environment
Platform: Linux
OS Version: Linux Mint Tessa (19.1)
Toggl Version: toggldesktop_7.4.459_amd64.deb

### 🐞 Actual behavior
When logging in with a Google account, the app claims the email or password is wrong.

### 📦 Additional info
There's probably a window with more information for a split second.
https://app.intercom.io/a/apps/ayixs927/conversations/22667832385

lise-toggl:
[linking again](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/3155115/conversations/22667832385) so intercom will track this

IndrekV:
Unable to reproduce the login issue. Maybe user has to re-enable the Google login from webapp.

MartinBriza:
This seems like a problem that is resolved by clearing the cache and cookies in the embedded browser we have inside our Linux app. I had an issue with the dialog showing for a split second too and it has disappeared after that fix. I'm adding this issue to the Google Login PR that contains the commit in question.

MartinBriza:
I'm convinced this issue is now fixed. I have tested the app on Linux Mint Tessa with Cinnamon and it works as expected.

PR #3405, opened 2019-10-08
Implement Google Login for the Linux app

### LINKED ISSUE #3279: Sign up with Google on Linux
### 💻 Environment
Platform: Linux

### 📒 Description
Add "Sign up with Google" button below the Sign up" button.
To be done after the library part is implemented in #3064.
Obtaining the Google access token is exactly the same as with Google login.

MartinBriza:
### 📒 Description
This PR enables Google Login in the Linux app.
It also fixes some issues in the OAuth code, namely:
 - The dialog no longer closes if you press Enter/Return while writing your email or password (this drove me insane while testing this)
 - The OAuth dialog no longer remembers your account with repeated Login/Signup attempts

### 🕶️ Types of changes

### 🤯 List of changes 
 - Linux app now supports Google Login
 - Google Login/Signup dialog on Linux is now more user-friendly

### 👫 Relationships
Closes #3279 #3095

### 🔎 Review hints
Probably similar as with other platforms: Check if Google Login works the same as before and the Google Signup functionality is consistent.


src/ui/linux/TogglDesktop/loginwidget.cpp
```diff
@@ -46,6 +46,20 @@ void LoginWidget::displayError(
     enableAllControls(true);
 }
 
+void LoginWidget::setGoogleSignup(bool visible) {
+    ui->email->setEnabled(!visible);
+    ui->password->setEnabled(!visible);
+    ui->password->clear();
+    if (visible) {
+        ui->email->setText("You're signing up with your Google Account");
```
NghiaTranUIT: Look like this LOC add the Text into the Email textbox. Thus, I couldn't select it after failed authentication.

MartinBriza:
The very signing up happens only after you click the `Sign up` button. It felt more natural to let the user log into their google Account first and THEN select their country and agree to the terms. This is intended behavior (except for the part where the "You're signing up.." string stays there even when trying to log in).

src/ui/linux/TogglDesktop/loginwidget.cpp
MartinBriza: This is now fixed by hiding the message when returning to the login screen.

MartinBriza: I changed the logic to ask for country and agree to the terms first and then enable opening the OAuth dialog. Resolving conversation.

NghiaTranUIT:
Thank for quick address @MartinBriza. Reviewing in second round now 🌮

NghiaTranUIT:
Improvement ticket for the Sign-up with Google at https://github.com/toggl/toggldesktop/issues/3425

PR #3426, opened 2019-10-11
Fix race condition in loadUserTimeEntryFromJSON

### LINKED ISSUE #2186: Abort due to contraint violation
/case/481342

The last couple of days I've had an error on Toggl Desktop for Windows. Something like this: ';Abort due to constraint violation';. I then can';t stop/start the time until I close Toggl and re-open it. I'd attach the log, but the ';add screenshot'; button won';t let me do that.

Not sure if related to 2185

### LINKED ISSUE #3374: UNIQUE constraint failed error
### 💻 Environment

Platform: macOS

OS Version: macOS 10.14.6

Toggl Version: 7.4.1012, 7.4.484

### 🐞 Actual behavior

Error message: "Constraint violation: UNIQUE constraint failed: time_entries.uid, time_entries.id"

The application then (usually) hangs and has to be force-quit.

I've also seen the same error message as issue #3372 -- no idea if there's a relation.

### 💯 Expected behavior

No error messages, syncing between desktop, phone, and web

### 🔨 Steps to reproduce

1. Open Toggl Desktop
2. Start tracking time
3. See errors

Note: I often start an entry on my phone and stop it on desktop (or vice versa). Don't know if this contributes. Entries seem to be working via the Android app and web.

### 📦 Additional info

Have tried both clearing the cache and reinstalling the desktop app. It downloaded 7.4.484, and the error showed up there as well. The "restart app to upgrade" button isn't working now, either.

This was submitted via the feedback option in the Help menu, but wanted to create an official report here.

Screen Shot 2019-09-30 at 9 31 24 AM

skel35:
### 📒 Description
The method `User::loadUserTimeEntryFromJSON` is being called from multiple threads:
- from the thread performing syncing
- from the thread handling WebSocket messages
- from the thread performing Load More
- from the thread performing post-login user data loading

The `Constraint violation` issue can happen when `related.TimeEntries` vector contains time entries with the same ID. 
The way it can happen is when two threads call `User::loadUserTimeEntryFromJSON` at the same time with JSON data of a brand new entry. This way `TimeEntry *model = related.TimeEntryByID(id);` will return `nullptr` for both of the threads and two entries with the same ID would get added to `related.TimeEntries`. Locking the code between getting the time entry by ID and adding a time entry with ID should fix the issue.
This scenario makes sense considering the users reporting these issues usually mention they often create entries with a web app or a mobile app.

### 🕶️ Types of changes

### 👫 Relationships
Closes #3374
Closes #2186

### 🔎 Review hints
I wasn't able to find proper steps to reproduce. See Description on my explanation on how the bug actually happens.
- Review the code itself
- Test the following:
1. time entry creation
2. time entry editing
3. syncing while starting/editing entries on web or mobile
4. clear cache and sync
5. offline mode

### DIFF
--- a/src/user.cc
+++ b/src/user.cc
@@ -1064,23 +1064,29 @@ void User::loadUserTimeEntryFromJSON(
         return;
     }
 
-    TimeEntry *model = related.TimeEntryByID(id);
+    TimeEntry* model;
+    {
+        Poco::Mutex::ScopedLock lock(loadTimeEntries_m_);
+        model = related.TimeEntryByID(id);
 
-    if (!model) {
-        model = related.TimeEntryByGUID(data["guid"].asString());
-    }
-
-    if (!data["server_deleted_at"].asString().empty()) {
-        if (model) {
-            model->MarkAsDeletedOnServer();
+        if (!model) {
+            model = related.TimeEntryByGUID(data["guid"].asString());
         }
-        return;
+
+        if (!data["server_deleted_at"].asString().empty()) {
+            if (model) {
+                model->MarkAsDeletedOnServer();
+            }
+            return;
+        }
+
+        if (!model) {
+            model = new TimeEntry();
+            model->SetID(id);
+            related.pushBackTimeEntry(model);
+        }
     }
 
-    if (!model) {
-        model = new TimeEntry();
-        related.pushBackTimeEntry(model);
-    }
     if (alive) {
         alive->insert(id);
     }
--- a/src/user.cc
+++ b/src/user.cc
@@ -1052,6 +1052,33 @@ void User::loadUserProjectFromJSON(
     model->LoadFromJSON(data);
 }
 
+bool User::SetTimeEntryID(
+    Poco::UInt64 id,
+    TimeEntry* timeEntry) {
+
+    poco_check_ptr(timeEntry);
+
+    {
+        Poco::Mutex::ScopedLock lock(loadTimeEntries_m_);
+        auto otherTimeEntry = related.TimeEntryByID(id);
+        if (otherTimeEntry) {
+            // this means that somehow we already have a time entry with the ID
+            // that was just returned from a response to time entry creation request
+            logger().error("There is already a newer version of this entry");
+
+            // clearing the GUID to make sure there's no GUID conflict
+            timeEntry->SetGUID("");
+
+            // deleting the duplicate entry
+            // this entry has no ID so the corresponding server entry will not be deleted
+            timeEntry->Delete();
+            return false;
+        }
+        timeEntry->SetID(id);
+        return true;
+    }
+}
+
 void User::loadUserTimeEntryFromJSON(
     Json::Value data,
     std::set<Poco::UInt64> *alive) {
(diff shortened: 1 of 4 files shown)


src/user.cc
```diff
@@ -1064,23 +1064,29 @@ void User::loadUserTimeEntryFromJSON(
...
         }
-        return;
-    }
 
-    if (!model) {
-        model = new TimeEntry();
-        related.pushBackTimeEntry(model);
+        if (!model) {
+            model = new TimeEntry();
+            model->SetID(id);
```
skel35: I've added this one line. `ID` was previously set in `LoadFromJSON()`, but now we want to make sure all entries in `related` have correct IDs before we get out of scope of the lock, so that `related.TimeEntryByID()` returns correct results.

skel35:
Removed the WIP label.
This PR is now ready for review.

src/user.cc
NghiaTranUIT: The code it self is straightforward and this context is locked properly. 

However, I'm not able to reproduce the Constrain issues itself, so I'm not guarantee that this fix could completely fix it or now.

NghiaTranUIT:
Here is my initial investigation report:

### Why does SQL Constraint error take place? 
Screen Shot 2019-10-22 at 17 31 24
Our database scheme enforces that the `id` and `guid` must be unique. Thus, if we try to insert an TimeEntry, which has `id` matches with the existing TE in the database -> The exception would be thrown. 

Please look back to our `database.cc` where we do `insert` or `update` on `time_entries` table.
https://github.com/toggl/toggldesktop/blob/29b60c16a5d067970efb632d4109a63026969f8f/src/database.cc#L2033-L2035
- This ☝️ func is in charge of saving or updating the TE to the local database.

https://github.com/toggl/toggldesktop/blob/29b60c16a5d067970efb632d4109a63026969f8f/src/database.cc#L2061
- This ☝️ if statement check whether we should do `insert` or `update`.
=> Therefore, it's possible that the race-condition take place here.

- Let check where we update the ID of TimeEntry is:
https://github.com/toggl/toggldesktop/blob/29b60c16a5d067970efb632d4109a63026969f8f/src/time_entry.cc#L414
from https://github.com/toggl/toggldesktop/blob/29b60c16a5d067970efb632d4109a63026969f8f/src/time_entry.cc#L402

- It turns out that this `void TimeEntry::LoadFromJSON(Json::Value data)` is actually executed in many threads and this `id` is not `protected` by any proper locks.
 
Screen Shot 2019-10-22 at 17 39 44

=> My conclusion that by execute `SetID()` from many concurrent threads could be a bottom of the problem.

### Suggestion
1. @skel35 Thank for the first attempt to protect when we setID.
2. Figure it out the solution to protect the `_id` in TimeEntry class.

skel35:
My conclusions after having analyzed all usages of `TimeEntry::LoadFromJSON()`:
1. `BaseModel::LoadFromDataString()` - is only called by an unused `BatchUpdateResult` code. I've created an issue to remove that code #3664. It is out of scope of the current issue.
2. `User::loadUserTimeEntryFromJSON()` - added proper lock protection when calling `TimeEntry::SetID()`.
3. `Context::pushEntries()` - added proper lock protection when calling `TimeEntry::SetID()`. Handled a case when we're trying to add an entry with an already existing ID (very unlikely theoretically possible due to a weird race condition). Added error throwing in the case when the server changes the ID, as it is a backend behavior that we do not expect.

I have moved `TimeEntry::SetID()` out of `LoadFromJSON()` so that it's always protected by a lock.

The PR is now ready for review.

PR #3441, opened 2019-10-21
Remove Tag by backspace

### LINKED ISSUE #3429: Add option to remove tags while editing TE with keyboard (backspace)
### 💻 Environment

Platform:  MacOS

### 📒 Description

Allow users to remove tags on existing TE by tabbing to tags field and pressing backspace. 

### ⭐️ Why do you want this?

A user is saying the previous version of the app allowed them to remove tags by pressing backspace (while editing an existing TE). They would like this functionality added back. 

[User](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/534741/conversations/24103673112)

### LINKED ISSUE #3440: Jump issues on duration text field in Manual mode
### 💻 Environment

Platform:  MacOS
OS Version:  
Toggl Version:  7.4.1023

### 🐞 Actual behavior

The duration field clears entry when tabbing through the fields to add new time entry. This is in manual mode. 

### 💯 Expected behavior

The duration should stay in the text field and not be cleared during entry. 

### 🔨 Steps to reproduce

Video here: https://toggl.slack.com/archives/C029WCX0T/p1571400237078600

### 📦 Additional info

[User Report](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/534741/conversations/24160322470)

NghiaTranUIT:
### 📒 Description
This PR introduce a minor change that users could delete selected tag by Backspace / Delete btn.
The focus ring also focus on the first Tag Token if possible.

This PR also tries to fix the Jump radically by determine of the reload is from the Syncer or not.

### What is the current problem on how we render the Editor data?
Currently, whenever we save the Editor (by updating the name, project, tags, ...), there are two triggers from the Library to **render the Editor**:
1. The first library notification is from the trigger func.
Ex: https://github.com/toggl/toggldesktop/blob/1781091766ad4fe261113496d0e51311c7b1d0eb/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/EditorViewController.swift#L604
2.The second notification is from the Syncer (When the TE is successfully synced to the Server)
https://github.com/toggl/toggldesktop/blob/1781091766ad4fe261113496d0e51311c7b1d0eb/src/context.cc#L4413

#### Problem
- Two render Editor notification are identical: There is no way to know which one is from User Action, which one is from the Syncer.
- Causes a lot of "Jump" issues we've encountered recently. For instance, Lost focus on text field and Losing the text, which is typing from the user

### Solution
- Introduce the bool to know the source of the notification

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Detect keys and delete the selected tag
- [x] Focus on the first Tags
- [x] Improve the Library to know the reload is from Syncer or not.
- [x] Ignore the Update on Editor if the reload is from the Syncer.

### 👫 Relationships
Closes #3429

### 🔎 Review hints
1. Make sure we enable Keyboard Focus on All Control In Keyboard Preference.
2. Open Editor for an entry with many tags
3. Focus on any Tag any hit Delete/Backspace
4. The Tag should be removed and the focus should select the first Tag -> 💯

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -307,7 +307,7 @@ error Context::StartEvents() {
     return noError;
 }
 
-error Context::save(const bool push_changes, const bool is_from_syncer) {
+error Context::save(const bool push_changes) {
     logger().debug("save");
     try {
         std::vector<ModelChange> changes;
@@ -323,7 +323,6 @@ error Context::save(const bool push_changes, const bool is_from_syncer) {
         UIElements render;
         render.display_unsynced_items = true;
         render.display_timer_state = true;
-        render.is_from_syncer = is_from_syncer;
         render.ApplyChanges(time_entry_editor_guid_, changes);
         updateUI(render);
 
@@ -844,8 +843,7 @@ void Context::updateUI(const UIElements &what) {
         UI()->DisplayTimeEntryEditor(
             what.open_time_entry_editor,
             editor_time_entry_view,
-            what.time_entry_editor_field,
-            what.is_from_syncer);
+            what.time_entry_editor_field);
     }
 
     if (what.display_time_entries) {
@@ -2278,7 +2276,7 @@ error Context::GoogleSignup(
 }
 
 error Context::AsyncGoogleSignup(const std::string &access_token,
-                                 const uint64_t country_id) {
+    const uint64_t country_id) {
     std::thread backgroundThread([&](std::string access_token, uint64_t country_id) {
         return this->GoogleSignup(access_token, country_id);
     }, access_token, country_id);
@@ -4453,7 +4451,7 @@ void Context::syncerActivity() {
                     logger().error("Error pushing OBM action: " + err);
                 }
 
-                displayError(save(false, false));
+                displayError(save(false));
             }
 
         }
@@ -4480,7 +4478,7 @@ void Context::syncerActivity() {
                     logger().error("Error pushing OBM action: " + err);
                 }
 
-                displayError(save(false, true));
+                displayError(save(false));
             }
         }
     }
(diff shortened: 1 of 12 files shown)


src/context.cc
```diff
@@ -4451,7 +4453,7 @@ void Context::syncerActivity() {
                     logger().error("Error pushing OBM action: " + err);
                 }
 
-                displayError(save(false));
+                displayError(save(false, false));
```
NghiaTranUIT: I deliberately select `false` here, because it happens when the user hit "Sync" button -> Should force rendering the content again

src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/EditorViewController.swift
```diff
@@ -46,6 +46,11 @@ final class EditorViewController: NSViewController {
 
     var timeEntry: TimeEntryViewItem! {
         didSet {
+            // If the TimeEntry is from the Syncer
+            // We shouldn't render since it causes the lost focus and the text is changed unexpectedly
+            if let timeEntry = timeEntry, timeEntry.isFromSyncer {
```
NghiaTranUIT: If the Render Editor notification is from the Syncer, we completely ignore it.

NghiaTranUIT:
Confirmed that the fix in Library from this PR also fixes the "Jump issues" in all Text Fields. (#3440 is fixed too)

NghiaTranUIT:
Confirmed that this PR will fix #3457, since it's a same source

NghiaTranUIT:
1. It's possible to separate the Remove Tag with the Jump issues. However, whenever we delete, we couldn't focus on the next Tag Token due to the jump 😰

Thus, the users have to hit the tab button multiple time to focus on the next tag again if they would delete again.

2. That make sense, but I have to manually ignore the Syncer Update for every controls if it is editing. Let me revert it

NghiaTranUIT:
All done @skel35. I removed the Syncer from library and do normal check in the Editor to make sure there is no jump after deleing the tag 👍

PR #3479, opened 2019-10-29
GitHub Actions workflow for macOS

MartinBriza:
### 📒 Description
Now, this is a very basic GitHub Actions workflow. It basically just builds the app and that's it. It's just to verify everything builds fine on macOS at this moment (which is pretty useful, too).

**IMPORTANT** I had to disable signing because I had issues using the provided signing key on both my machine and in the virtual environment on GitHub. Anyway, I think this is for the good because outside developers won't have to tweak the build settings in XCode each time they want to try building the app for themselves. For us it probably doesn't change much, we can override this in any case.

Meanwhile I'm working on making signing work in a more sane way on my MacBook in the meantime but the builds are just taking too much time. Signing and packaging will be done in a separate PR in the future. 

### 🕶️ Types of changes

### 🤯 List of changes
- macOS app is built as a part of CI

### 🔎 Review hints
If there's a CI/CD badge at the bottom of this and it's green, that's probably it, no more testing needed really.

It shouldn't matter if this gets before or after the other Actions PR I opened a while ago.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,6 +11,19 @@ on:
       - '**'
 
 jobs:
+  macos:
+    runs-on: macOS-latest
+    steps:
+    - uses: actions/checkout@v1
+      with:
+        fetch-depth: 1
+    - name: Build the dependencies
+      run: |
+        make deps
+    - name: Build the app
+      run: |
+        make
+
   linux-basic:
     runs-on: ubuntu-latest
     if: github.event_name != 'create'
--- a/src/ui/osx/TogglDesktop/TogglDesktop.xcodeproj/project.pbxproj
+++ b/src/ui/osx/TogglDesktop/TogglDesktop.xcodeproj/project.pbxproj
@@ -1318,7 +1318,6 @@
 				ORGANIZATIONNAME = Alari;
 				TargetAttributes = {
 					69FC17F017E6534400B96425 = {
-						DevelopmentTeam = B227VTMZ94;
 						LastSwiftMigration = 1020;
 						ProvisioningStyle = Automatic;
 						SystemCapabilities = {
@@ -1856,11 +1855,11 @@
 				CLANG_CXX_LIBRARY = "compiler-default";
 				CLANG_ENABLE_MODULES = YES;
 				CODE_SIGN_ENTITLEMENTS = TogglDesktop/TogglDesktop.entitlements;
-				CODE_SIGN_IDENTITY = "Mac Developer";
+				CODE_SIGN_IDENTITY = "-";
 				CODE_SIGN_STYLE = Automatic;
 				COMBINE_HIDPI_IMAGES = YES;
-				DEVELOPMENT_TEAM = B227VTMZ94;
-				ENABLE_HARDENED_RUNTIME = YES;
+				DEVELOPMENT_TEAM = "";
+				ENABLE_HARDENED_RUNTIME = NO;
 				FRAMEWORK_SEARCH_PATHS = (
 					"$(inherited)",
 					"$(SRCROOT)",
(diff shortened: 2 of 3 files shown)


MartinBriza:
Aha, no CI/CD badge bere because I'm opening a PR from a fork. Well, the last build is here: https://github.com/MartinBriza/toggldesktop/commit/164879b60386fadd7b1345a10c750a2be3cfab13/checks?check_suite_id=286649387

.github/workflows/main.yml
```diff
@@ -11,6 +11,19 @@ on:
       - '**'
 
 jobs:
+  macos:
+    runs-on: macOS-latest
```
NghiaTranUIT: I suggest that `macOS-10.14` would be better because it ensures that Github Action will run on Xcode 10.3 in macOS 10.14. On the other hand, the latest means 10.15 and Xcode 10.11

Ref: https://help.github.com/en/github/automating-your-workflow-with-github-actions/software-in-virtual-environments-for-github-actions#installed-sdks

src/ui/osx/TogglDesktop/TogglDesktop.xcodeproj/project.pbxproj
```diff
@@ -1856,11 +1855,11 @@
 				CLANG_CXX_LIBRARY = "compiler-default";
 				CLANG_ENABLE_MODULES = YES;
 				CODE_SIGN_ENTITLEMENTS = TogglDesktop/TogglDesktop.entitlements;
-				CODE_SIGN_IDENTITY = "Mac Developer";
+				CODE_SIGN_IDENTITY = "-";
 				CODE_SIGN_STYLE = Automatic;
 				COMBINE_HIDPI_IMAGES = YES;
-				DEVELOPMENT_TEAM = B227VTMZ94;
-				ENABLE_HARDENED_RUNTIME = YES;
```
NghiaTranUIT: I suggest the we shouldn't disable the `ENABLE_HARDENED_RUNTIME` because it's a crucial step to submit for Apple Notarization. If we set NO, we couldn't submit.

NghiaTranUIT: It turns out that if we remove code signing in `DEVELOPMENT_TEAM` and `CODE_SIGN_IDENTITY`, the app won't build until we disable `ENABLE_HARDENED_RUNTIME`. Thus, we have to find another way to disable the codesigning.

.github/workflows/main.yml
```diff
@@ -11,6 +11,19 @@ on:
...
+    steps:
+    - uses: actions/checkout@v1
+      with:
+        fetch-depth: 1
+    - name: Build the dependencies
+      run: |
+        make deps
+    - name: Build the app
+      run: |
+        make
```
NghiaTranUIT: In order to keep the code-signing config and `ENABLE_HARDENED_RUNTIME` in the project, we could improve the Makefile:
- If it's Github Action (`GITHUB_ACTION` env is visible), we could disable code-signing in `xcodebuild_command`
- If not, just run as normal, so we we mini-bot still works as usual.

For instance,
```makefile
ifdef GITHUB_ACTION
xcodebuild_command=xcodebuild \
				  -scheme TogglDesktop \
				  -workspace src/ui/osx/TogglDesktop.xcworkspace  \
				  -configuration Debug CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO
else
xcodebuild_command=xcodebuild \
				  -scheme TogglDesktop \
				  -workspace src/ui/osx/TogglDesktop.xcworkspace  \
				  -configuration Debug
endif
```
Ref: https://help.github.com/en/github/automating-your-workflow-with-github-actions/virtual-environments-for-github-actions#default-environment-variables

MartinBriza: This doesn't solve anything though because we need notarization and signing in GitHub actions too.

NghiaTranUIT: Unfortunately, If we directly disable the DEVELOPMENT_TEAM and ENABLE_HARDENED_RUNTIME in the Xcode Project, I couldn't build the app in the local machine anymore, even the buildbot is also affected.

In general, my suggestion would be
- Keep the DEVELOPMENT_TEAM and ENABLE_HARDENED_RUNTIME in Xcode Project
- Improve the makefile in `default` route: Disable codesigning in xcodebuild if it's Github action => So the Github action could build for testing without dealing with the codesign.
- For the Notarization & Codesign in Github action: We could improve later in `codesign.sh` and `notarize.sh` from the branding repo.

## Result
- Able to build in local machine
- Able to run for testing in Github Action (No Codesiging)
- Able to do notarization + Codesign when we support in the Branding repo.

Please correct me if I'm wrong, or please share us your proposal 🌮

MartinBriza:
Ok, this PR is now obsoleted by #3554

PR #3495, opened 2019-11-05
Run our test suite on Linux in GitHub Actions

MartinBriza:
### 📒 Description
This change enables running our test suite on each commit.
Test failures won't make the builds fail at this moment (because sometimes the tests simply just fail).
Only failures on `master` will get reported to Slack.

I also wrote a testing script that could possibly work on other platforms without a lot of changes so we should be able to use it in the future when we make the tests work elsewhere.

### 🕶️ Types of changes

### 🔎 Review hints
Look over at https://github.com/toggl/toggldesktop/runs/288812523 where the last result from this branch is stored. You could also check the `#desktop-bots` channel to see the spam from when I was testing the code with an intentionally broken test suite.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,6 +26,14 @@ jobs:
     - name: Build the Linux binary
       run: |
         bash ./dist/linux/package.sh build
+    - name: Run App test
+      run: |
+        pushd build-tarball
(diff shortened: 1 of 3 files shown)


MartinBriza:
:D Well, I wanted to use something that'll eventually work on all platforms and the XML coming from GTest is simple enough. For more serious parsing I'd go for a more serious tool, too.

PR #3502, opened 2019-11-11
Reduce workspace/preferences and preferences/desktop fetches (lib)

### LINKED ISSUE #3417: Reduce numerous workspace/preferences fetches
### 💻 Environment
Platform: Library

### 📒 Description
We fetch `/workspaces/preferences` too many times. It is actually used to detect if the user has any workspaces left. Currently, it seems that it is triggered every time the user gets an error from the backend. We should narrow it down to certain error that indicates this scenario is possible. The error is probably something like `403 Forbidden`

skel35:
### 📒 Description
Reduce workspace/preferences and preferences/desktop fetches

### 🕶️ Types of changes

### 🤯 List of changes
- Perform workspace/preferences fetches only on locked time entry errors, login, and full sync.
- Perform preferences/desktop fetches only on login and full sync.
- `TogglClient` is now passed as a const reference instead of a pointer. I had to fix a few methods due to conflicts with #3344, then went on and fixed all the other ones.

### 👫 Relationships
Closes #3417 

### 🔎 Review hints
1. Test that `preferences/desktop` fetch is performed on full sync and on login:
- Open the desktop app
- Open the web app, go to Profile settings -> Group similar time entries -> Check or uncheck the checkbox
- Trigger sync manually in the desktop app / Log out and log back into the desktop app
- Make sure the time entry grouping setting was applied

2. Test that `workspace/preferences` fetch is performed on full sync and on login:
- Open the desktop app
- Open the web app, lock time entries (https://support.toggl.com/en/articles/2206898-locking-time-entries)
- Trigger sync manually in the desktop app / Log out and log back into the desktop app
- Make sure the entries are displayed as locked

3. Test that `workspace/preferences` fetch is performed when the locked time entry error is received.
- Open the desktop app
- Open the web app, lock time entries (https://support.toggl.com/en/articles/2206898-locking-time-entries)
- Try editing an entry that is supposed to be locked
- The entries are displayed as locked without the need to manually trigger the sync.
Unfortunately, this leads to a bad state, caused by #3513 and #3512. I suggest fixing these issues is out of scope of this PR as they can happen now as well.

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -79,6 +79,7 @@ Context::Context(const std::string &app_name, const std::string &app_version)
 , update_check_disabled_(UPDATE_CHECK_DISABLED)
 , trigger_sync_(false)
 , trigger_push_(false)
+, trigger_full_sync_(false)
 , quit_(false)
 , ui_updater_(this, &Context::uiUpdaterActivity)
 , reminder_(this, &Context::reminderActivity)
@@ -1072,6 +1073,7 @@ void Context::scheduleSync() {
 }
 
 void Context::FullSync() {
+    trigger_full_sync_ = true;
     user_->SetSince(0);
     Sync();
 }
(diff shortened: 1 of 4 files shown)


skel35:
Rebased and resolved conflicts.
This PR is ready for review.

skel35:
Closing in favor of #4318

PR #3509, opened 2019-11-13
Implement Feedback window UI

### LINKED ISSUE #3447: Send Feedback window design update
### 💻 Environment
Platform: Windows

### 📒 Description
Update the Send Feedback window view according to the new design

skel35:
### 📒 Description
Implement Feedback window UI according to mockups.

### 🕶️ Types of changes

### 🤯 List of changes
- Implement Feedback window UI according to mockups
- Implement an immediate attached file size validation
- Implement field validation logic (validate on Send click and LostFocus, switch focus to the field when the validation error occurs)
- Switched to .NET 4.6.1
- First step towards MVVM and RFP: created a FeedbackViewModel, started using ReactiveUI

### 👫 Relationships
Closes #3447.

### 🔎 Review hints
- Verify that Send Feedback functionality works.
- Verify that UI is according to mockups, both Light and Dark mode.

Things not yet implemented:
- Error around TextBox/ComboBox as in the mockup (issue #3510)
- MessageBox UI (issue #3511)

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/Converters/BytesToStringConverter.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/Converters/BytesToStringConverter.cs
@@ -1,6 +1,8 @@
 using System;
 using System.Globalization;
 using System.Windows.Data;
+using static System.Convert;
+using static System.Math;
 
 namespace TogglDesktop.Converters
 {
@@ -17,11 +19,13 @@ namespace TogglDesktop.Converters
         {
             string[] suf = { "B", "KB", "MB", "GB", "TB", "PB", "EB" };
             if (byteCount == 0)
-                return "0" + suf[0];
-            var bytes = Math.Abs(byteCount);
-            var place = System.Convert.ToInt32(Math.Floor(Math.Log(bytes, 1024)));
-            var num = Math.Round(bytes / Math.Pow(1024, place), 1);
-            return (Math.Sign(byteCount) * num).ToString() + " " + suf[place];
+            {
+                return $"0 {suf[0]}";
+            }
+            var bytes = Abs(byteCount);
+            var place = ToInt32(Floor(Log(bytes, 1024)));
+            var num = Round(bytes / Pow(1024, place), 1);
+            return $"{(Sign(byteCount) * num)} {suf[place]}";
         }
 
         public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/Converters/BytesToStringConverter.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/Converters/BytesToStringConverter.cs
@@ -1,0 +1,32 @@
+using System;
+using System.Globalization;
+using System.Windows.Data;
+
+namespace TogglDesktop.Converters
+{
+    public class BytesToStringConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            var byteCount = (long)value;
+            var humanReadableBytes = BytesToString(byteCount);
+            return humanReadableBytes;
+        }
+
+        private static string BytesToString(long byteCount)
+        {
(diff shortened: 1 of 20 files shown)


src/ui/windows/TogglDesktop/TogglDesktop/ui/Converters/BytesToStringConverter.cs
```diff
@@ -0,0 +1,32 @@
...
+            var humanReadableBytes = BytesToString(byteCount);
+            return humanReadableBytes;
+        }
+
+        private static string BytesToString(long byteCount)
+        {
+            string[] suf = { "B", "KB", "MB", "GB", "TB", "PB", "EB" };
+            if (byteCount == 0)
+                return "0" + suf[0];
+            var bytes = Math.Abs(byteCount);
```
kornelijepetak: Why not add a `using static System.Math` and simplify it all to:
```cs
var bytes = Abs(byteCount);
var place = Convert.ToInt32(Floor(Log(bytes, 1024)));
var num = Round(bytes / Pow(1024, place), 1);
return $"{Sign(byteCount) * num} {suf[place]}";
```
(+string interpolation)

skel35: Completely agree 👍  (except `Convert.ToInt32` thing as it creates ambiguity with the `Convert` method above and causes compilation errors).

NghiaTranUIT:
Overall, the PR is good, but it has some flaws:

### The Error label presents unexpectedly
1. Open the Feedback
2. Click on the Topic Dropdown
3. Hover on those options
4. The error alert appears, which is incorrect because I haven't selected any thing yet.



#### Suggestions
I suggest that we should remove the validation when losing the focus and only validate the input when the user click on Send btn.

Or we can find the way to make the error behavior is better

### Close the Success Panel cause all Windows app disappears
1. Open the feedback, select the topic and sent the content
2. Hit Send
3. The Success Panel and click OK
4. All Windows (Feedback, Success Panel and the Main app) are disappear. I have to open again from the taskbar.



#### Suggestion
I would suggest to keep the Main Windows as usual. The action of closing the Feedback or Success Panel shouldn't affect on the main Windows.

skel35:
Thank you for the feedback @NghiaTranUIT!
I've addressed the mentioned issues:
- Fixed the validation error appearing on the combobox hover
- Made the keyboard focus switch happen only when clicking Send button
- Made the main window focus happen when the Feedback window is closing (this is actually a WPF bug that I had to workaround)
- Removed some code behind and moved more control to the viewmodel

The PR is ready for another review.

src/ui/windows/TogglDesktop/TogglDesktop/ui/windows/AboutWindow.xaml
```diff
@@ -25,7 +26,7 @@
     <i:Interaction.Behaviors>
         <behaviors:HideWindowOnClosingBehavior />
```
Abdulbaqi-Alshareef: Why hiding window instead of just closing it?

skel35: It's just the way it had been previously implemented. Won't speculate on why, but I can say that with our app being rather small keeping those extra views in memory doesn't hurt our performance that much. To make it possible to close child windows instead of hiding we need to take care of even subscription/unsubscription, state, and the logic which is being executed even when the child windows are hidden.
This behavior class was introduced by refactoring the existing procedural code-behind logic into more declarative code. This also makes it easier to get rid of "hiding instead of closing" in the future.

Abdulbaqi-Alshareef: Closing window needs to take care of "Finalizing" on closing but it makes code cleaner and predictable, and in the opposite side hiding window needs to "resetting state" which done here by re-creating new  viewmodel and rebinding controls and maybe forget to reset some objects or controls could cause unpredictable behavior, this approach also makes the code harder to read.

I understand that its just the way it had been previously implemented, but I can't see any benefit of using it in this window specifically.

PR #3530, opened 2019-11-20
Fix new Timeline APIs for Window and Linux

NghiaTranUIT:
### 📒 Description
This PR fixes 

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Add Start/End time param into `toggl_start` func in Windows + Linux
- [x] Fix the testcase
- [x] Update new TogglApi.cs
- [x] Fix couldn't convert `char_t *` to `wchar_t *` in library

### 🔎 Review hints
- Able to run and build the app in Windows / macOS and Linux

### DIFF
--- a/src/gui.cc
+++ b/src/gui.cc
@@ -541,17 +541,20 @@ void GUI::DisplayTimeline(
             bool item_present = false;
             TogglTimelineEventView *event_app = first_event;
             while (event_app) {
-                if (compare_string(event_app->Filename, copy_string(event.Filename().c_str())) == 0) {
+                char_t *filename = copy_string(event.Filename().c_str());
+                if (compare_string(event_app->Filename, filename) == 0) {
                     timeline_event_view_update_duration(event_app, event_app->Duration + event.Duration());
                     app_present = true;
                     item_present = false;
                     ev = reinterpret_cast<TogglTimelineEventView *>(event_app->Event);
                     while (ev) {
-                        if (compare_string(ev->Title, copy_string(event.Title().c_str())) == 0) {
+                        char_t *title = copy_string(event.Title().c_str());
+                        if (compare_string(ev->Title, title) == 0) {
                             timeline_event_view_update_duration(ev, ev->Duration + event.Duration());
                             item_present = true;
                         }
                         ev = reinterpret_cast<TogglTimelineEventView *>(ev->Next);
+                        free(title);
                     }
 
                     if (!item_present) {
@@ -562,6 +565,7 @@ void GUI::DisplayTimeline(
                     }
                 }
                 event_app = reinterpret_cast<TogglTimelineEventView *>(event_app->Next);
+                free(filename);
             }
 
             if (!app_present) {
@@ -618,8 +622,10 @@ void GUI::DisplayTimeline(
     }
 
     std::string formatted_date = Formatter::FormatDateHeader(TimelineDateAt());
-    on_display_timeline_(open, copy_string(formatted_date.c_str()), first_chunk, first_entry, start_day, end_day);
+    char_t *date = copy_string(formatted_date.c_str());
+    on_display_timeline_(open, date, first_chunk, first_entry, start_day, end_day);
     delete first_entry;
+    free(date);
 }
 
 TogglTimelineEventView* GUI::SortList(TogglTimelineEventView *head) {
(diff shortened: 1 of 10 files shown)


src/gui.cc
```diff
@@ -541,13 +541,13 @@ void GUI::DisplayTimeline(
             bool item_present = false;
             TogglTimelineEventView *event_app = first_event;
             while (event_app) {
-                if (compare_string(event_app->Filename, event.Filename().c_str()) == 0) {
+                if (compare_string(event_app->Filename, copy_string(event.Filename().c_str())) == 0) {
```
MartinBriza: `copy_string` here is completely unnecessary (just `c_str` is fine) and you don't free the result anywhere, causing a considerable memory leak. Same for the usages in the same file below.

NghiaTranUIT: Unfortunately, c_str() causes the error in Windows. As I remember that it was
```
Error | C2664 | 'int compare_string(const wchar_t *,const wchar_t *)': cannot convert argument 2 from 'const _Elem *' to 'const wchar_t *'        with        [            _Elem=char        ] | TogglDesktopDLL | C:\Users\nghiatran\Desktop\toggldesktop\src\gui.cc | 544 |  
```

Screen Shot 2019-11-20 at 20 38 56

It's reason why I try using `copy_string`.

NghiaTranUIT: 👍 for the point regarding the leaking on c_string. I will fix it now

MartinBriza: Using `copy_string` is still wrong in this case unless you use `free` to the returned pointer afterwards (which introduces an unnecessary performance overhead).
The way to right way to fix this on the Windows side is to use the same code we have in `copy_string` to convert `std::string` to `std::wstring`:
```
char_t *copy_string(const std::string &s) {
#if defined(_WIN32) || defined(WIN32)
    std::wstring ws;
    Poco::UnicodeConverter::toUTF16(s, ws);
    return wcsdup(ws.c_str());
#else
    return strdup(s.c_str());
#endif
}
```
The part between the `#if` and `#else` is what you need to use to convert the string to a wide string.

NghiaTranUIT: Thank for the hint. I will try this approach 🌮

MartinBriza: It actually seems `to_char_t` in `toggl_api_private.cc` does the same thing but I'm not entirely sure about that. 
It also seems to be written in a very dangerous way (but it probably still works most of the time)... This is a different issue and I'll open a report for it.

NghiaTranUIT: I choose the safe way to release the `char_t *` after using the `copy_string` rather than fixing the the `copy_string` body since it could affect entire the library. Please help me to review on the second round @MartinBriza 👍

src/gui.cc
```diff
@@ -541,17 +541,20 @@ void GUI::DisplayTimeline(
             bool item_present = false;
             TogglTimelineEventView *event_app = first_event;
             while (event_app) {
-                if (compare_string(event_app->Filename, copy_string(event.Filename().c_str())) == 0) {
+                char_t *filename = copy_string(event.Filename().c_str());
```
MartinBriza: While this fixes the memory leak, it adds an unnecessary `strdup`/`free` pair of calls, increasing system load. I outlined how to fix this properly in the comments at the previous commit.

NghiaTranUIT: I understand, but as I mention that I would choose the safe approach this PR.

I suggest that we should move the actually fix to different ticket? since the fix is on `copy_string()` and it impacts the majority of the codebase as well as it's out of scope of this PR too.

If it's reasonable, we could merge then I will open the ticket 👍

MartinBriza: There's nothing wrong with `copy_string` itself. It's just unnecessary to use it here as it only increases the performance impact of this method.

I suggested the usage of `to_char_t`, did you try it?

skel35: @MartinBriza As you mentioned previously, `to_char_t` is written in a dangerous way, and I can confirm this as I was writing it only for tests and, as I noticed later, missed the memory leak. So if we want to use it in production we should fix it first.

NghiaTranUIT:
All done. Thank @MartinBriza for detailed feedback on C++ world 🌏

NghiaTranUIT:
Thank guys for cooperatively reviewing the PR. All good and it's going to merge now 👍

PR #3543, opened 2019-11-21
"Record shortcut" UI update

### LINKED ISSUE #3525: "Record shortcut" UI update
### 💻 Environment
Platform: Windows

### 📒 Description
- [x] Use HotKeyBox from MahApps.Metro instead of current custom solution
- [x] Use NHotkey.Wpf for registering shortcuts functionality instead of current custom solution
- [x] Update "Record shortcut" UI to match new design update

skel35:
### 📒 Description
Implementation of Record shortcut UI update using HotKeyBox from MahApps.Metro.

Note: Merge after #3526.

### 🕶️ Types of changes

### 🤯 List of changes
- Use HotKeyBox from MahApps.Metro instead of current custom solution
- Use NHotkey.Wpf for registering shortcuts functionality instead of current custom solution
- Update "Record shortcut" UI to match new design update
- Implement nicer validation behavior

### 👫 Relationships
Closes #3525.

### 🔎 Review hints

- Check that setting and removing global shortcuts works.
- Check that it's not possible to set an already existing shortcut.
- Check that global shortcuts work when application is hidden in the tray
- When the shortcut was assigned in Toggl Desktop and the user has quit the application, then the same shortcut can be used in another application (to check that the global shortcut is freed up).


skel35:
Thanks @NghiaTranUIT ! I've discovered the reason for the bug and fixed it, please have a look once again 🌮

PR #3549, opened 2019-11-21
Fix tagging so the main build workflow actually runs on a new release

MartinBriza:
### 📒 Description
For reasons that are not entirely clear to me using just `GITHUB_TOKEN` doesn't quite cut it for the workflow to be able to create a release that will in turn be registered by the other (`main.yml`) workflow we have in place to actually compose the releases.
Before we have a better solution (like triggering one workflow from another), the way to go is using a `TAGGING_SECRET` in Secrets containing a github token that can create tags in this repo...

### 🕶️ Types of changes

### 🔎 Review hints
Pretty hard to review without actually merging a PR so let's just hope everything will be fine after merging this.

### DIFF
--- a/.github/workflows/pr_tagger.yml
+++ b/.github/workflows/pr_tagger.yml
@@ -10,4 +10,4 @@ jobs:
     - uses: actions/checkout@master
     - name: Bump version and push tag
       run: |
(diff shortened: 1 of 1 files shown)


MartinBriza:
I see there was #3519 before but the problem in that was a completely different one - the Workflow was trying to create a release for a tag that has been deleted while it was running, that's why it failed.

PR #3550, opened 2019-11-22
Better tooltip for tray icon

### LINKED ISSUE #3533: Better tooltip for tray icon
### 💻 Environment

Platform: Windows

### 📒 Description
Make the tooltip text multi-line for a better overview

skel35:
### 📒 Description
Implements nicer-looking tray icon tooltip instead of a single one-line string.
image

### 🕶️ Types of changes

### 👫 Relationships
Closes #3533

### 🔎 Review hints
Test how tooltip works in various scenarios (tracking, non-tracking, empty description, no project, project+no client, project+task+client)

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/TogglDesktop.csproj
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/TogglDesktop.csproj
@@ -4,6 +4,7 @@
   <PropertyGroup>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
     <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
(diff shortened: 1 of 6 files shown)


IndrekV:
Are tags and billable also visible in this tooltip?

skel35:
no, just description/project/task/client for now

skel35:
The tooltip is not hoverable because it disappears as soon as you move cursor from the tray icon, so the tags icon has limited value (just shows whether at least 1 tag exists or not).
If we're adding billable icon, then it makes sense to add the tags icon as well. I could do it when I'm implementing the new UI of the mini-timer, so that I can brush up the code a bit as well, if you think it's a good idea @IndrekV

MartinBriza:
I like how such a small piece of code changes the functionality of the tray tooltip so much! This is definitely a great addition and a huge improvement over the previous tooltip so I'm merging this.
I'll open an issue to add the icons Indrek mentioned.

PR #3553, opened 2019-11-25
Make sure syncing is performed after coming back from sleep (lib)

### LINKED ISSUE #3537: App syncing issues after coming back from sleep
### 💻 Environment

Platform: Windows

### 🐞 Actual behavior
The app does not sync

### 💯 Expected behavior
The app should sync and get the current state

### 🔨 Steps to reproduce

1. Put the computer to sleep
2. Start timer from some other Toggl App
3. Wake computer
4. Open Toggl Desktop and see if it sync the running timer

### 📦 Additional info
Reported in Go Desktop Week

Might be connected to #3049

skel35:
### 📒 Description
The library has `isPostponed` method to determine whether the function has been scheduled to run again in the future so that it doesn't have to run now.
It assumes that if the scheduled function is running before the timestamp for which it was scheduled, it means it is scheduled to run again in the future and so, cancels the current function.
In reality, POCO Timer doesn't guarantee to run exactly after the timestamp with microseconds precision. According to my tests, scheduled function often runs up to 30 ms before the timestamp, for which it was scheduled.
Therefore, I've added a 100 ms precision threshold when comparing timestamps. I find it high enough to get rid of these rounding errors (the biggest I've experienced in 10 tries was 33.5 ms) and low enough so that the user can hardly perform some operation several times during the 100 ms interval.

### 🕶️ Types of changes

### 👫 Relationships
Closes #3537.

### 🔎 Review hints
See STR in #3537.

### DIFF
--- a/src/const.h
+++ b/src/const.h
@@ -27,6 +27,7 @@
 #define kMinimumAllowedYear 2006
 #define kMaximumAllowedYear 2030
 #define kMaximumDescriptionLength 3000
+#define kTimeComparisonEpsilonMicroSeconds 100000 // 100 ms
 
 #define kLostPasswordURL "https://toggl.com/forgot-password?desktop=true"
 #define kGeneralSupportURL "https://support.toggl.com/toggl-on-my-desktop/"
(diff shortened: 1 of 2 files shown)


MartinBriza:
Hm, this is kinda challenging for me to test because my tablet doesn't really have a sleep state in the traditional sense. With "modern sleep", it actually keeps running and syncing stuff in the background including network connection... which doesn't cause #3537 to happen on it.

I'll try to come up with a different way to test this but in the meantime, somebody with a "normal" computer or a VM should probably take a look at this too

PR #3578, opened 2019-12-03
Add in-app message fetching to library

### LINKED ISSUE #3412: Add in-app messaging
### 💻 Environment
Platform: library

### 📒 Description
We'll add an in-app messaging system so if needed we can send information straight to the users.

**UI**
The information will be shown as a separate message. Details about mac design #3571

**The logic**
We'll have a JSON file online and the app will fetch it from time to time to see if the version number has changed and it needs to show the info that was fetched. 

The json file will be kept in `toggl/toggldesktop` repo gh page.

We'll keep the message version number in the local database. Also, we'll keep the message DateTime so we don't show too old messages if the user has just installed the app. Messaged from the same day or day before can be shown. Also, we use the app version to verify the user is on the version of the app that needs to see the message.

We should also take into account that we might want to show links in the message. Either has a link in a separate value in which case it will have something like `Read more` at the end of the message

or 

Have the possibility to use HTML inside the message text to show links.

```
{
"id":1,
"datetime": timestamp,
"appversion":"7.4.1022",
"type": "same",
"days": "0"
"title": "title text here",
"text": "New version of Desktop available",
"button": "Get it now!"
"url": "https://toggl.com"
}
```

Breakdown of the message json:

`id` - message id is used to determine if user has seen the message. If id saved in local db is smaller then show the message

`type` - can be `0`/`1`/`2`
this is how app version is compared. 

```
- 0 (same) - The message is shown only to a specific version of the app.
- 1 (older) - The message is shown to apps older than the specified version
- 2 (newer) - The message is shown to apps newer than the specified version
```
`days` - how many days this message is active. 0 means forever

`appversion` - so we can narrow down the people seeing the message. Maybe there is an issue with a certain version then we only show the message to them. Value can also be `<=7.4.1022` in which case all older version get the message

`datetime` - we check the datetime so we don't show too old messages. Maybe only stuff sent today or yesterday. This needs more clarification.

`title` - title text for the message
`text` - message content
`button` - text on the button (empty if no button needs to be shown)
`url` - url to open when user clicks on the button

IndrekV:
### 📒 Description
Adds logic to fetch in-app messages from json files

### 🕶️ Types of changes

### 🤯 List of changes

- [x] Added field to settings database table 3e560da
- [x] Save and fetch value form database table 232e96a
- [x] Connections between lib and ui 594c940
- [x] Added message fetching logic 60377c3
- [x] Trigger message fetch on app start 60377c3
- [ ] Periodic message fetching (every4 hrs) _[not done]_

### 👫 Relationships

Closes #3412

### 🔎 Review hints

Currently, I used a temporary location for JSON file to test the logic.

Message parameters should be checked correctly:
- id in database is older -> show message
- 3 scenarios of app version validation
- days and datetime logic

### DIFF
--- a/src/gui.cc
+++ b/src/gui.cc
@@ -573,6 +573,29 @@ void GUI::DisplayUpdateDownloadState(
     free(version_string);
 }
 
+void GUI::DisplayMessage(const std::string &title,
+                         const std::string &text,
+                         const std::string &button,
+                         const std::string &url) {
+    logger().debug("DisplayMessage: " + title);
+
+    char_t *tmp_title = copy_string(title);
+    char_t *tmp_text = copy_string(text);
+    char_t *tmp_button = copy_string(button);
+    char_t *tmp_url = copy_string(url);
+    on_display_message_(
+        tmp_title,
+        tmp_text,
+        tmp_button,
+        tmp_url);
+
+    free(tmp_title);
+    free(tmp_text);
+    free(tmp_button);
+    free(tmp_url);
+}
+
+
 void GUI::DisplaySettings(const bool open,
                           const bool record_timeline,
                           const Settings &settings,
--- a/src/context.cc
+++ b/src/context.cc
@@ -1984,6 +1984,17 @@ int64_t Context::GetWindowEditSizeWidth() {
     return value;
 }
 
+void Context::SetMessageSeen(
+    const int64_t value) {
+    displayError(db()->SetSettingsMessageSeen(value));
+}
+
+int64_t Context::GetMessageSeen() {
+    Poco::Int64 value(0);
+    displayError(db()->GetMessageSeen(&value));
+    return value;
+}
+
 void Context::SetKeyStart(
     const std::string &value) {
     displayError(db()->SetKeyStart(value));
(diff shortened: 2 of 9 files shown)


src/gui.cc
```diff
@@ -573,6 +573,29 @@ void GUI::DisplayUpdateDownloadState(
...
+        tmp_text,
+        tmp_button,
+        tmp_url);
+
+    free(tmp_title);
+    free(tmp_text);
+    free(tmp_button);
+    free(tmp_url);
+}
+
```
IndrekV: @MartinBriza @NghiaTranUIT I'm sure this is not the best solution. Looking for feedback how to improve this part.

NghiaTranUIT: I'm good with it and have no better solution to avoid this boilerplate code.

NghiaTranUIT:
@IndrekV Do you would to implement the `Periodic message fetching (every4 hrs) [not done]` in different ticket? because it would be great if we merge into master, then I could fully integrate and test in the UI quickly

NghiaTranUIT:
Merged to master for the InApp message UI. Feel free to open new PR for the interval checker 👍

PR #3584, opened 2019-12-05
Improve in-app message logic 

### LINKED ISSUE #3583: In-app message improvements
### 💻 Environment
Platform: lib

### 📒 Description

- [x] Add periodic message check (every 4 hrs)
- [x] Add analytics `show (appstart/periodic), click (close/cta)`
- [x] Change the date logic. Drop `datetime` and `days` and go for from and to

`from`- is start day from which the message is active
`to` -  is the end date after which this message is not active anymore

connected to #3412

IndrekV:
### 📒 Description
Adds periodic fetch and user analytics, improves message logic and format.

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Add periodic message check (every 4 hrs)
- [x] Add analytics `show (appstart/periodic), click (close/cta)`
- [x] Change the date logic. Drop `datetime` and `days` and go for from and to

`from`- is start day from which the message is active
`to` -  is the end date after which this message is not active anymore

### 👫 Relationships

Closes #3583

### 🔎 Review hints

**Prep**
- Move `fetchMessage(0)` out of the `production` env check (line 300 in `context.cc`)
- All production checks or `CanDisplayMessage` checks should be commented out so `fetchMessage` executes until the end

**Periodic sync**

- Do **Prep**
- Update the `kCheckInAppMessageIntervalSeconds` (inside `const.h`) to `2` (2seconds)
- Move `startPeriodicInAppMessageCheck();` to be the first row of `fechMessage` function.
- Set breakpoint in `fetchMessage` function and see if it triggers every 2 seconds

**Analytics**

all production checks should be commented out so `fetchMessage` executes until the analytics and triggers analytics request

**New message format**

- Do **Prep**
The from datestamp is 2 yesterday and to timestamp is Dec 10

Change the now date to create situations where it is invalid. (outside `from` and `to` range)

### DIFF
--- a/src/toggl_api.cc
+++ b/src/toggl_api.cc
@@ -1428,3 +1428,11 @@ void track_edit_size(void *context,
     }
     app(context)->TrackEditSize(width, height);
 }
+
+void toggl_iam_click(void *context,
+                     const uint64_t type) {
+    if (!context) {
+        return;
+    }
+    app(context)->TrackInAppMessage(type);
+}
--- a/src/analytics.cc
+++ b/src/analytics.cc
@@ -56,6 +56,30 @@ void Analytics::TrackOSDetails(const std::string &client_id) {
     Track(client_id, "osdetails", ss.str());
 }
 
+void Analytics::TrackInAppMessage(const std::string &client_id,
+                                  const std::string &id,
+                                  const Poco::Int64 type) {
+    std::stringstream ss;
+    ss << "message/" << id << "/";
+
+    if (type == 0) {
+        // in-app message shown on app start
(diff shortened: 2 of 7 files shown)


src/toggl_api.cc
```diff
@@ -1428,3 +1428,11 @@ void track_edit_size(void *context,
     }
     app(context)->TrackEditSize(width, height);
 }
+
+void toggl_iam_click(void *context,
```
skel35: looks like this method wasn't added to the header file, so it can't be called from the clients.

PR #3636, opened 2019-12-18
Timeline Timezone issues

### LINKED ISSUE #3635: [BUG] Timeline: Timezone issues
### Description
There is an issues on Timeline that Time Entries and Activities are not fetched from the Lib properly due to a timezone.

For instance, here is the Timeline in Timezone +7 (Vietnam)
1. There are several Time Entries at 00:00 -> 7:00AM 
Screen Shot 2019-12-18 at 15 08 19

2. However, Timeline doesn't render properly
- The start of selected day is 7AM (Look at the Scroll Bar) => Incorrect ❌ 
- Missing some TimeEntries from 0:00 -> 7AM ❌ 

Screen Shot 2019-12-18 at 15 08 39

We should fix the timezone issues.

### 💻 Environment
Platform: macOS
Toggl Version: Latest Timeline build

### 🐞 Actual behavior
- Start of the current day is off. It should be at 00:00 in current time zone
- Missing some TimeEntries in 00:00 -> Current Timezone

### 💯 Expected behavior
- Timeline should render properly. From 00:00 to midnight at current time zone
- All Time Entries in TimeEntry tab should be rendered in Timeline Tab.

### 🔨 Steps to reproduce
1. Create some TE from 00:00 -> Your current time zone. For instance, if you're at +2, please create some TEs from 0:00 -> 2AM.
2. Switch to Timeline mode
3. Observer the start time of current date and the number of TEs

NghiaTranUIT:
### 📒 Description
This PR will fix the timezone issues, which is described in #3635

### 🕶️ Types of changes

### 🤯 List of changes
- Fix the timezone in library when calculating the Start - End time in order to fetch the TimeEntry and Event properly
- Fix the timeline convert func in Timeline Flow Layout

### 👫 Relationships
Close #3635
Fix "Change the start Time of TE make it moves to yesterday" in https://github.com/toggl-open-source/toggldesktop/issues/3380#issuecomment-538887226

### 🔎 Review hints
- Reproduce the Steps in https://github.com/toggl-open-source/toggldesktop/issues/3635#issue-539525212 and make sure that it works
- Observe that the TimeEntry and Event date (start/end) should be rendered properly in the Timeline View
- Make sure the start of the current day is at 00 (current timezone)
- Make sure that we can see the TimeEntry & event from 00:00 -> current Timezone. In the latest build, they are vanished.

Screen Shot 2019-12-18 at 15 09 24

### DIFF
--- a/src/gui.cc
+++ b/src/gui.cc
@@ -490,10 +490,11 @@ void GUI::DisplayTimeline(
         TimelineDateAt().year(),
         TimelineDateAt().month(),
         TimelineDateAt().day());
+    int tzd = datetime.tzd();
 
     // Get all entires in this day (no chunk, no overlap)
     TogglTimeEntryView *first_entry = nullptr;
-    time_t start_day = datetime.timestamp().epochTime();
+    time_t start_day = datetime.timestamp().epochTime() - tzd;
     time_t end_day = start_day + 86400; // one day
     for (unsigned int i = 0; i < entries_list.size(); i++) {
         view::TimeEntry te = entries_list.at(i);
@@ -510,7 +511,7 @@ void GUI::DisplayTimeline(
     while (datetime.year() == TimelineDateAt().year()
             && datetime.month() == TimelineDateAt().month()
             && datetime.day() == TimelineDateAt().day()) {
-        time_t epoch_time = datetime.timestamp().epochTime();
+        time_t epoch_time = datetime.timestamp().epochTime() - tzd;
         time_t epoch_time_end = epoch_time + 900;
 
         // Create new chunk
--- a/src/ui/osx/TogglDesktop/Feature/Timeline/Models/TimelineData.swift
+++ b/src/ui/osx/TogglDesktop/Feature/Timeline/Models/TimelineData.swift
@@ -145,7 +145,7 @@ extension TimelineData {
             times.append(current)
             current += span
         }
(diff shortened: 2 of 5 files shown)


src/gui.cc
```diff
@@ -490,10 +490,11 @@ void GUI::DisplayTimeline(
         TimelineDateAt().year(),
         TimelineDateAt().month(),
         TimelineDateAt().day());
+    int tzd = datetime.tzd();
 
     // Get all entires in this day (no chunk, no overlap)
     TogglTimeEntryView *first_entry = nullptr;
-    time_t start_day = datetime.timestamp().epochTime();
+    time_t start_day = datetime.timestamp().epochTime() - tzd;
```
NghiaTranUIT: We have to minus the Timezone different since the EpochTime is at GTM 0, but we need to fetch all TimeEntry from the midnight of current time zone (+7 for instance).

If we have any better solution, please comment below 👍

NghiaTranUIT:
That's good point. I'm going to remove the last hour in the Timeline View

NghiaTranUIT:
All done @IndrekV 

Screen Shot 2019-12-18 at 16 09 25

src/gui.cc
IndrekV: No better solution comes to mind. Let's go with this one.

PR #3648, opened 2019-12-23
Prevent update Duration for a running TimeEntry if the text isn't changed from users

### LINKED ISSUE #3613: Clicking in the editable duration field results in field overwriting running timer when TE is stopped
### 💻 Environment

Platform:   MacOS  10.14.6
OS Version:    10.14.6
Toggl Version:   7.4.1041

### 🐞 Actual behavior

If you  are clicking in the  editable duration field, and then stop the timer, the time reverts to what was in the editable field and not the running  timer.

### 💯 Expected behavior

We should use the running timer time when the TE is stopped. 

### 🔨 Steps to reproduce

Not reproducible on MacOS Catalina 

### 📦 Additional info

[Slack](https://toggl.slack.com/archives/C029WCX0T/p1575957691302600)
[Case reference](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/534741/conversations/23939691503)
Case also includes videos.

NghiaTranUIT:
### 📒 Description
This PR fix a bug, which is described in https://github.com/toggl-open-source/toggldesktop/issues/3613.

### Root of problem?
- If the TE is running and we open an Editor then focusing in. If we stop the Running TimeEntry, it would cause the Duration Text field end of editing => Make the change on Library

https://github.com/toggl-open-source/toggldesktop/blob/366133a851733440947b13a4c5c8d04cc06ef855/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/EditorViewController.swift#L150-L154

Therefore, we need to know whether the duration should be saved or not.

### Attempts?
**Determine whenever or not a TextField is focus when clicking or Tab** that seems a trivial task, but it is one of different tasks since AppKit doesn't officially provide any methods.

The following is few attempts:

- [textDidBeginEditing(_:)](https://developer.apple.com/documentation/appkit/nstextfield/1399426-textdidbeginediting) doesn't call if we just select or focus on this text field. It only is executed if we start typing
- (BOOL)becomeFirstResponder is not reliable since it only executed at the first time and it doesn't call after that anymore.

### Why is this approach?
This solution is simple enough to determine that if the duration value is actually changed by the user's action or not by an introduction of EditableTextField class

Ref: https://stackoverflow.com/questions/34840914/detecting-when-nstextfield-becomes-active/34841087 

### 🕶️ Types of changes
**Bug fix** (non-breaking change which fixes an issue

### 🤯 List of changes
- Add EditableTextField and get the state whether the user edits the text
- Ignore the duration update when the TE is stopped and it's not the user' action

### 👫 Relationships
Close #3613

#### Resolve the user's case
- Follow steps in a video at https://toggl.slack.com/archives/C029WCX0T/p1575957691302600
- Make sure the duration doesn't updated if we stop the TE

#### Able to change duration in the running TE
1. Start any TE and open the Editor
2. Change the duration 
3. Make sure the duration is changed because it's a user action

#### Able to change duration in the non-running TE
1. Open Editor
2. Change the duration 
3. Make sure the duration is changed because it's a user action

### DIFF
--- a/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/EditorViewController.swift
+++ b/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/EditorViewController.swift
@@ -508,7 +508,7 @@ extension EditorViewController: NSTextFieldDelegate {
         guard let textField = obj.object as? NSTextField else { return }
 
         // Duration
-        if textField == durationTextField {
+        if textField == durationTextField && durationTextField.isTextChanged {
             durationTextFieldOnChange(self)
             return
         }
--- a/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/EditorViewController.swift
+++ b/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/EditorViewController.swift
@@ -148,7 +148,7 @@ final class EditorViewController: NSViewController {
     }
 
     @IBAction func durationTextFieldOnChange(_ sender: Any) {
-        guard durationTextField.stringValue != timeEntry.duration else { return }
+        guard durationTextField.stringValue != timeEntry.duration && durationTextField.isTextChanged else { return }
         DesktopLibraryBridge.shared().updateTimeEntry(withDuration: durationTextField.stringValue,
                                                       guid: timeEntry.guid)
     }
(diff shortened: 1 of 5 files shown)


src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/EditorViewController.swift
```diff
@@ -508,7 +508,7 @@ extension EditorViewController: NSTextFieldDelegate {
         guard let textField = obj.object as? NSTextField else { return }
 
         // Duration
-        if textField == durationTextField {
+        if textField == durationTextField && durationTextField.isTextChanged {
```
skel35: I suggest removing this check as we already perform a check in the method being called here + without this check `controlTextDidEndEditing` method looks more simple and straightforward.

skel35:
@NghiaTranUIT Let's merge this PR, the fix is good 👍

PR #3649, opened 2019-12-23
Remove old logic and make it focus on Description Editor by default (…

### LINKED ISSUE #3234: Add time for idle time, active entry fields
### 💻 Environment

MacOS

### 📒 Description
A few users have reported that in the past, when they had idle time detection on, when the popup appeared and they went to add time for that, it'd automatically have the description field highlighted for entry completion, so wouldn't need to click the field. 

Since the latest update, this appears to have changed, and now requires the user to click the respective field.

So the request is to highlight the field and allow instant input of information to speed up user workflows.

### ⭐️ Why do you want this?
To speed up input of entries and efficiency 

[example user case](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/2097915/conversations/23333544729)

Thanks,

NghiaTranUIT:
### 📒 Description
This PR would improve the logic to focus on Description whenever we open Time Entry Editor.

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Remove old focus code
- [x] Make sure it focus on the Description whenever the Editor open

### 👫 Relationships
Closes #3234

### The editor focus on description if we create TimeEntry from Idle Notification
1. Set Idle detection as 1 minutes
2. Start any TE and keep idle for 1 min to make sure the Idle notification appears
3. Select "Add Idle time as a new Time Entry"
4. New TE is running and the Editor is opening 
5. Make sure the description is focus, so the user could start typing without "Click to focus"

### DIFF
--- a/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/EditorViewController.swift
+++ b/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/EditorViewController.swift
@@ -54,11 +54,7 @@ final class EditorViewController: NSViewController {
         didSet {
             fillData(oldValue)
             registerUndoForAllFields()
-
-            // Check if the Editor has update with new TimeEntry -> Reset focus to Description TextField
-            // If not, just keep focus on current textfield
-            let defaultFocus = checkShouldFocusByDefault(for: oldValue, newValue: timeEntry)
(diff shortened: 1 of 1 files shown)


skel35:
@NghiaTranUIT please check out if this PR can still be relevant, otherwise close it

NghiaTranUIT:
Just test and it doesn't fix the problem anymore. Close it due to out of date

Issue #3244, opened 2019-08-22
[Library] Update Description, Project and Tag in one func


NghiaTranUIT:
# 📒 Description
It's clearly that as soon as we call any `toggl_time_entry_...` to update Description, Project, or Tag. It will shoot the notification on main thread to update the UI.

However, in some case, when we update with AutoCompleteItem (Description + Project + tag), it will notify 3 times, which is redundant. (See this PR for more detail https://github.com/toggl/toggldesktop/pull/3243)

### ⭐️ Why do you want this?
- Reduce the overheat when rendering too many time just by doing single action (Update TE from AutoCompleteItem)

### Proposal
1. Add an additional parameter
```c++
const bool_t shouldUpdateUI
```
into save() func
https://github.com/toggl/toggldesktop/blob/c0e7870da78abe19e33bfb4e5050a41ffec7d397/src/context.cc#L310
to determine if we need to call the Render or not.
2. Introduce the new func to update description, project and tag with by re-using existing method with no update
3. Call save and the UI
=> Only one notification

### 💻 Environment
Platform: lib

### Accpetance Criteria 
- [ ] One notification update when selecting the AutoComplete Item in Editor
- [ ] Other update methods in library should work as usual

skel35:
Would like to add to this:
Currently, the logic behind selecting the autocomplete in Editor differs a little bit between platforms.
Differences are, for example, in whether autocomplete overwrites the project if one is set, whether `billable` is copied from autocomplete item, etc.
We can take `EditView.xaml.cs` `descriptionAutoComplete_OnConfirmCompletion()` from Windows as a basis, as it is a mature implementation which handles corner cases.

Having such a method in the library will allow achieving uniform behavior across platforms, so I definitely agree this should be done.

IndrekV:
I'm all for moving things to library so we can do it great once an use it everywhere. Moving forward one step, it would be great to move the whole autocomplete filtering to library as we currently have 3 implementations of filtering for the same logic. I think that it might be difficult to do as the implementations and rendering is too different on platforms. All feedback on this is welcome.

skel35:
Lib and macOS parts are done, now Windows and Linux need to start using the new method as well. Tweaked the labels to reflect the current state of the task.

PR #3653, opened 2019-12-24
Update Description, Project and Tag in one func

NghiaTranUIT:
### 📒 Description
This PR improves how we update the TE from TimeEntryAutoComplete by reducing some unnecessary UI Update from the library from 4 -> 2 times:
- 1 call is for updating the UI from AutoComplete Item
- 1 call is from the Syncer

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Introduce update_ui in save func to only update the UI if need
- [x] Add 1 func to update TE from autocomplete item (Description, project, task, tags)
- [x] Use this func to update on mac build

### 👫 Relationships
- macOS part of [Library] Update Description, Project and Tag in one func #3244

#### All Features are remains
- Play around and make sure we're able to update the TE as usual

#### There are only 2 UI Updates from Library
1. Open Editor, and search some TE in Description Text Field
2. Select any TE AutoComplete Item in Description
3. Observer that there are only 2 UI Update from the library by looking at the Console
```
❌ TimeEntryListViewController displayTimeEntryList
```

### DIFF
--- a/src/context.h
+++ b/src/context.h
@@ -389,13 +389,13 @@ class TOGGL_INTERNAL_EXPORT Context : public TimelineDatasource {
         const std::string &project_guid);
 
     error updateTimeEntry(
-                          const std::string &GUID,
-                          const std::string &description,
-                          const Poco::UInt64 task_id,
-                          const Poco::UInt64 project_id,
-                          const std::string &project_guid,
-                          const std::string &tags,
-                          const bool billable);
+        const std::string &GUID,
+        const std::string &description,
+        const Poco::UInt64 task_id,
+        const Poco::UInt64 project_id,
+        const std::string &project_guid,
+        const std::string &tags,
+        const bool billable);
 
     error SetTimeEntryDate(
         const std::string &GUID,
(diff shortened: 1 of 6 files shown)


NghiaTranUIT:
That's total make sense @tanel 👍 Thank for the keyword "Flag argument"

I will address it 💻

NghiaTranUIT:
Ready for reviewing @skel35 👍

### How it's addressed
> This is a FlagArgument anti-pattern https://martinfowler.com/bliki/FlagArgument.html. I know we have a few of those in our codebase, but I think it's bad and we shouldn't introduce more.

I completely remove the introduction of `bool update` param in `save()` function

> The logic is moved into the library, but we still perform the same actions (like finding the entry by GUID) multiple times. If we perform them once it will improve the performance.

- Split and reuse the logic when updating description, project, tag, and billable
- Introduce big function to update all things with one single update and one time to find TimeEntry from GUID
```c++
error Context::updateTimeEntry(
                       const std::string &GUID,
                       const std::string &description,
                       const Poco::UInt64 task_id,
                       const Poco::UInt64 project_id,
                       const std::string &project_guid,
                       const std::string &tags,
                     const bool billable) 
```

> (minor) I believe this is a bug that the macOS app does not update the billable flag, so I think updating it should be included in this method as well.

Update billable param in the same function

src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/DesktopLibraryBridge.m
```diff
@@ -192,15 +192,15 @@ - (void) registerUndoDeleteItem:(TimeEntryViewItem *)item
...
-	[self updateTimeEntryWithTags:autocomplete.tags guid:timeEntry.GUID];
+    const char *tags = [[autocomplete.tags componentsJoinedByString:@"\t"] UTF8String];
+    toggl_set_time_entry(ctx,
+                         [timeEntry.GUID UTF8String],
+                         [autocomplete.Description UTF8String],
+                         autocomplete.TaskID,
+                         autocomplete.ProjectID,
+                         0,
+                         tags,
+                         timeEntry.billable);
```
skel35: `billable` should come from `autocomplete` and not from `timeEntry`.
We need this to make sure autocompleting billable time entries sets the billable flag correctly. I believe the fact that we didn't do it before was a bug on macOS.

NghiaTranUIT:
Thank you for your feedback @skel35. It's done for reviewing 👍

src/context.h
```diff
@@ -388,6 +388,15 @@ class TOGGL_INTERNAL_EXPORT Context : public TimelineDatasource {
         const Poco::UInt64 project_id,
         const std::string &project_guid);
 
+    error updateTimeEntry(
```
skel35: this is a public method so it should be capitalized

NghiaTranUIT: Thanks for pointing out 👍

PR #3659, opened 2019-12-27
Fix how we trigger Pomodoro if we update Start / Duration in current running Time Entry

### LINKED ISSUE #2778: Toggl Desktop MacOS 7.4.330 - Pomodoro/Manual mode issue
### 💻 Environment

Platform: macOS
OS Version: MacOS Mojave 10.14.2.
Toggl Version: 7.4.330 (stable)

### 🐞 Actual behavior
Similar to #2773 users are not able to create TEs in manual mode or edit a TE while it's running to adjust the duration so that it's longer than the Pomodoro interval. 

Intercom case: 20602902845

### 💯 Expected behavior
The Pomodoro timer should not be triggered by manual edits or time entries created in manual mode. 

### 🔨 Steps to reproduce

1. Pomodoro timer = ON (20min intervals)
2. Start timer
3. While the timer is running change the start time to 45 minutes ago

Instead of changing the start time only, because Pomodoro is ON the end time will change as well to be exactly 20 minutes after the new start time.

NghiaTranUIT:
### 📒 Description
This PR would improve how we trigger the Pomodoro if the user updates a Start or Duration on a current Running Time Entry.

### Current problem
Describe in https://github.com/toggl-open-source/toggldesktop/issues/2778

### Solution
- Skip Pomodoro if the total time (after updating from the user) is greater than Pomodoro Time
- Pomodoro is still working if the total time (after updating from the user) is lesser than Pomodoro Time.

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Add logic to check if we should skip Pomodoro or not when the user updating the "Start / Duration" on the current running Time Entry

### 👫 Relationships
Closes #2778

#### 1. Don't show Pomodoro notification if we change Start Time of the current Running TimeEntry later than Pomodoro Time
1. Set Pomodoro Time in Preference is 10 minutes
2. Start Time Entry A and open Editor on the current running 
3. Update the Start time back to 1 hour ago. Ex: from 16:30 to 15:30
4. Make sure the Pomodoro notification doesn't trigger
5. Make sure the running TimeEntry is still running 
6. Click Stop button after 10 mins
7. Make sure the start and duration time of this TE is correct.

#### 2. Don't show Pomodoro notification if we change Duration of the current Running TimeEntry later than Pomodoro Time
1. Repeat the 1st review instruction and update the Duration instead of the Start Time.
2. Verify that all expectations are met.

#### 3. Show Pomodoro notification if we change Start Time of the current Running TimeEntry less than Pomodoro Time
1. Set Pomodoro Time in Preference is 10 minutes
2. Start Time Entry A and open Editor on the current running 
3. Update the Start time back to 5 min ago. Ex: from 16:30 to 16:25
4. Make sure the Pomodoro notification triggers after 5 min (duration of this TE is 10 mins)
5. Make sure the running TimeEntry is stopped 
7. Make sure the start and duration time of this TE is correct.

#### 4. Show Pomodoro notification if we change Duration of the current Running TimeEntry less than Pomodoro Time
1. Repeat the 3rd review instruction and update the Duration instead of the Start Time.
2. Verify that all expectations are met.

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -4554,6 +4554,13 @@ void Context::displayPomodoro() {
                 return;
             }
         }
+
+        // At this state, the total duration > pomodoro time
+        // but we should skip if the user manually update start / duration time
+        if (current_te->ManualUpdateFromUser()) {
+            return;
+        }
+
         const Poco::Int64 pomodoroDuration = settings_.pomodoro_minutes * 60;
         wid = current_te->WID();
         Stop(true);
--- a/src/context.cc
+++ b/src/context.cc
@@ -3148,6 +3148,12 @@ error Context::SetTimeEntryDuration(
     }
 
     te->SetDurationUserInput(duration);
+
+    // Skip checking Pomodoro if it's current running TE and the total time is > pomodoro
+    if (checkIfSkipPomodoro(te)) {
+        te->SetSkipPomodoro(true);
+    }
+
     return displayError(save(true));
 }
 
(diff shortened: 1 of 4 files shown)


NghiaTranUIT:
Ready to review 👍

PR #3663, opened 2020-01-02
Add a lightweight VS solution without Poco projects

### LINKED ISSUE #3662: Add a lightweight VS solution without Poco projects
### 💻 Environment
Platform: Windows

### 📒 Description
Add a lightweight VS solution that does not contain Poco projects.

### Benefits:
- The solution can be cleaned without cleaning the Poco artifacts which always stay the same. Poco artifacts should be built just once. This is the most important one for me as I'm wasting 7-8 minutes waiting for the Poco build which can happen multiple times a week.
- The solution is a lot smaller so working Visual Studio becomes faster.
- It's the first step towards getting rid of building Poco altogether.

### Downsides:
- Before working with the smaller solution for the first time or after `git clean -fx`, the full solution has to be built at least once or the Poco binaries should be copied over.
- Both solutions have to be updated when we're adding/removing projects. This happens rarely and in the future, we're going to remove the "full" solution.

skel35:
### 📒 Description
Add a lightweight VS solution without Poco projects

### 🕶️ Types of changes
- **Improvement** (non-breaking change which adds functionality) -->

### 🤯 List of changes
- Added a lightweight VS solution without Poco projects
- Removed the direct reference to `sqlite3` to keep the dependency only on Poco artifacts (`database.cc` change from @MartinBriza 's PR: https://github.com/toggl-open-source/toggldesktop/pull/3618/files).

### 👫 Relationships
Closes #3662

### 🔎 Review hints

Method 1:
1. build `TogglDesktop.sln`
2. copy `bin`, `bin64`, `lib`, `lib64` folders from `third_party/poco` to some other place outside the repo.
3. `git clean -xdf`
4. copy `bin`, `bin64`, `lib`, `lib64` folders back to `third_party/poco`.
5. open `TogglDesktop.NoPoco.sln`
6. should be able to work with it normally. Try Rebuild solution -> should work properly, should take seconds instead of minutes.

Method 2:
1. `git clean -xdf`
2. build `TogglDesktop.sln`
3. open `TogglDesktop.NoPoco.sln`
4. should be able to work with it normally.

### DIFF
--- a/src/database.cc
+++ b/src/database.cc
@@ -51,7 +51,7 @@ Database::Database(const std::string &db_path)
     session_ = new Poco::Data::Session("SQLite", db_path);
 
     {
-        int is_sqlite_threadsafe = sqlite3_threadsafe();
+        int is_sqlite_threadsafe = Poco::Data::SQLite::Utility::isThreadSafe();
 
         std::stringstream ss;
         ss << "sqlite3_threadsafe()=" << is_sqlite_threadsafe;
--- a/src/lib/windows/TogglDesktopDLL/TogglDesktopDLL.vcxproj
+++ b/src/lib/windows/TogglDesktopDLL/TogglDesktopDLL.vcxproj
@@ -222,8 +222,8 @@
     <Link>
       <SubSystem>Windows</SubSystem>
       <GenerateDebugInformation>true</GenerateDebugInformation>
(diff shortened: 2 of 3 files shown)


skel35:
Merging as it was approved before the code freeze and is not risky for macOS release.

PR #3691, opened 2020-01-08
Login view improvements

### LINKED ISSUE #3616: Autocomplete country selection in signup
### 💻 Environment
Platform: Windows

### 📒 Description
Implement a proper country autocomplete in the Signup view vinstead of a standard WPF combobox.

### LINKED ISSUE #3690: Login view new UI: fix minor issues
### 💻 Environment
Platform: Windows

### 📒 Description
- When the user starts typing the email, the validation gets fired immediately and the user sees an annoying red border with an error message. The validation should fire first only after the user has moved keyboard focus from the email field.
- When the user starts the app logged in, then logs out => the email field is not focused.
- There is a 1px jump of the text when hovering over the Login/Signup button.

skel35:
### 📒 Description
Login view improvements

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Enabled autocomplete on country selection
- [x] Got rid of 1px jump when hovering over Login/Signup button
- [x] Updated button colors (hover/pressed background/border) based on the new design specs.
- [x] Fixed issue when email validation got activated during initial typing
- [x] Fixed issue when email field did not get focused after logout

### 👫 Relationships
Closes #3616
Closes #3690

### 🔎 Review hints
1. Verify that #3616 is implemented
2. Verify that #3690 is implemented.
3. Verify that button hover/pressed background/border colors are corresponding to the design in Zeplin

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop.sln.DotSettings
+++ b/src/ui/windows/TogglDesktop/TogglDesktop.sln.DotSettings
@@ -1,0 +1,2 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=Toggl/@EntryIndexedValue">True</s:Boolean></wpf:ResourceDictionary>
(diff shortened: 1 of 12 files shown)


NghiaTranUIT:
IMO, I see that the Input Validation (Red Border) sometime is little strange (GIF)



I haven't inputed, just jump around, but I got Red Border.

----------------------------
I suggest that we could remove the Auto Validation. Just let it as it is and ONLY validate the input after I click on Login or Sign Up button (Same with macOS).

Do you have any thought @IndrekV @skel35 🤔 ?

NghiaTranUIT:
It's okay to keep the current Validation for this PR and address it later 👍

skel35:
I agree validation shouldn't be fired when just clicking through fields or switching to sign up, that's a mistake.
I think it makes sense that it gets fired after the user typed the email and switched from the email field, but before pressing the Login/Signup button.

skel35:
I'll go with applying validation only after the first Login/Signup button click.
I'd like to give more instant feedback to the user, but applying validation on "LostFocus" seems too intrusive as the email field loses focus when the user switches from Login to Signup.
I'd appreciate any suggestions of precise logic of when to fire the validation so that the user receives more rapid feedback, although I think it's not much of an issue as it would be if we had complex password rules, for example.

NghiaTranUIT:
I will merge when the GA is finished 🙌

skel35:
No need to wait for GA as Linux tests just got stuck and this PR has no changes in Linux app or library.

NghiaTranUIT:
Yay. That's right 👍

PR #3692, opened 2020-01-09
Fix missing QRect include error

a-andreyev:
### 📒 Description

Getting error [similar to this](https://aur.archlinux.org/packages/toggldesktop/#comment-723453) when building `v7.5.3` version with Qt 5.14.0.

### 🕶️ Types of changes

Looks like including `QRect` to the linux ui headers helps. 

### 🤯 List of changes

`src/ui/linux/TogglDesktop/toggl.h`


MartinBriza:
Strange that this happens but thank you for looking into it and submitting the fix!
Nothing to review here, really, merging right away.

PR #3713, opened 2020-01-10
Add HTTP support (lib)

MartinBriza:
### 📒 Description
This PR adds HTTP support in addition to HTTPS. I'm running the backend code locally and this will make life a lot easier while testing.
The gist is we now look at the URI scheme from the URIs/URLs returned from `urls.cc` and if it is `http`, we create a `Poco::Net::HTTPClientSession` instead of `Poco::Net::HTTPSClientSession` (which is pretty easy due to polymorphism).
I also renamed all `toggl::HTTPS<x>` classes to `toggl::HTTP<x>` because having the `S` in there would be a lie.

### 🕶️ Types of changes

### 🔎 Review hints
Test if you can still connect to staging and production as expected. You can also test connecting to a HTTP Toggl server but I think our main API is not accessible over it. Basically just focus on what wold change for the user (and that should be nothing: nothing should change).

### DIFF
--- a/src/analytics.cc
+++ b/src/analytics.cc
@@ -154,12 +154,12 @@ const std::string GoogleAnalyticsEvent::relativeURL() {
 }
 
 void GoogleAnalyticsEvent::runTask() {
-    HTTPSRequest req;
+    HTTPRequest req;
     req.host = "https://ssl.google-analytics.com";
     req.relative_url = relativeURL();
 
-    HTTPSClient client;
-    HTTPSResponse resp = client.Get(req);
+    HTTPClient client;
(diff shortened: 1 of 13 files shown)


MartinBriza:
Huh, what are these errors on Windows? O.o

NghiaTranUIT:
Let me review on macOS side 👍

skel35:
Please don't merge it until Windows build is not fixed ⚠️

IndrekV:
@skel35 Is the Windows build fixed?

skel35:
@IndrekV No, I haven't looked into it

PR #3729, opened 2020-01-15
Add more Menu Context in Timeline

### LINKED ISSUE #3724: Add context menu items for time entries in timeline view
### 💻 Environment
Platform: macOS

### 📒 Description
For more easy time entry editing we should add right mouse click context menu items.

items in the context menu:
```
- Continue this entry
- Start entry from the end of this entry
- Delete
```

In case there are also overlap fixing items in list they will be in top and then other items follow

NghiaTranUIT:
### 📒 Description
This PR introduces some menu items for all Timeline TimeEntry.

### 🕶️ Types of changes
**New feature** (non-breaking change which adds functionality

### 🤯 List of changes
- [x] Add new menu items to Timeline menu
- [x] Normal Timeline TimeEntry and overlapped TimeEntry will have menu context
- [x] Logic for new menu

### 👫 Relationships
Closes #3724

#### New menu logic is working normal TimeEntry and Overlapped TimeEntry
1. Right click on normal TimeEntry or Overlapped TimeEntry
2. "Continue" -> Verify that the running TimeEntry is the same 
3. "Start" -> Verify that there is new TE at the end of the selected TE -> Editor is presented
4. "Delete" -> Verify that we could delete the selected TE (or show confirm popup)

#### Overlap menu is disable in normal TimeEntry
1. Right click on normal TimeEntry (no overlap)
2. Verify that the both "Overlap menus" are disabled

Screen Shot 2020-01-15 at 15 19 27

#### Overlap menu is enable in normal TimeEntry
1. Right click on overlapped TimeEntry
2. Verify that the both "Overlap menus" are enabled

Screen Shot 2020-01-15 at 15 19 45

### DIFF
--- a/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/DesktopLibraryBridge.m
+++ b/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/DesktopLibraryBridge.m
@@ -238,8 +238,8 @@ void *ctx;
 - (NSString *)starNewTimeEntryAtStarted:(NSTimeInterval)started ended:(NSTimeInterval)ended
 {
 	char *guid = toggl_start(ctx,
-							 @"".UTF8String,
-							 "0",
+							 "",
+							 "",
 							 0,
 							 0,
 							 0,
--- a/src/ui/osx/TogglDesktop/Feature/Timeline/Models/TimelineData.swift
+++ b/src/ui/osx/TogglDesktop/Feature/Timeline/Models/TimelineData.swift
@@ -144,7 +144,15 @@ class TimelineData {
     }
 
     func startNewFromEnd(_ timeEntry: TimelineTimeEntry) {
-        NotificationCenter.default.post(name: Notification.Name(kStarTimeEntryWithStartTime), object: timeEntry.timeEntry.ended)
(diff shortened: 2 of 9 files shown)


IndrekV:
Is it possible to not show the disabled menu item at all in case there is no overlap

NghiaTranUIT:
Yes, it's possible, but I'm not prefer this approach. If we completely hide two bottom menu item, they are possibly undiscoverable until they user right-clicks on the "Overlap item".

If we disable it, it's clear that the user is aware that there are those features.

IndrekV:
Valid point. It is good so the user can already learn that there is this possibility. Thank you for sharing this idea.

IndrekV:
The logic is:

If the view shows today - start the time entry
If the view shows a day in the past - create the entry (don't start) and show edit popup (this is the current functionality.

NghiaTranUIT:
All done 👍 

## Changelogs
- [x] Update the title of menu depend on the today TE or the past
- [x] Start an empty running TE if it's today TE
- [x] Create an empty TE and oped Editor on it if it's a TE in the past
- [x] If we start an empty running TE, which the start date is the future -> It will create an empty TE, which start time = now

Issue #3730, opened 2020-01-15
Extract Timer viewmodel


skel35:
### 💻 Environment
Platform: Windows

### 📒 Description
Reduce code duplication between timer and mini timer by extracting a common viewmodel.

Issue #3725, opened 2020-01-14
Windows app focus issues


skel35:
### 💻 Environment
Platform: Windows

### 📒 Description
- [x] The timer is now focused by default ~The timer should receive focus when the application window gains focus~
- [x] The duration field receives focus when the time entry is added via manual mode ~The edit view should receive focus when the time entry is added via manual mode~
- [x] After starting time entry from the mini timer, shortcuts stop working. This is caused by the mini timer losing keyboard focus (`Keyboard.FocusedElement` is `null`)

skel35:
The last issue is fixed in the new design as the Stop button grabs focus after starting the entry from the mini timer.
Fixed.

PR #3732, opened 2020-01-15
Mini timer design update

### LINKED ISSUE #2009: Make the size of the mini-timer less.
Hello. Please make the size of the mini timer is less, it is too big! Or, provide the opportunity to make it smaller (I use: Windows Desktop Toggle App). Thank you!

### LINKED ISSUE #3486: Mini timer design update
### 💻 Environment
Platform: Windows

### 📒 Description
Implement Mini timer view according to the new design.
Currently Mini timer shares the view code with the main window timer, in the new design the timers look different so their view code should be separated.

skel35:
### 📒 Description
Implementation of the new design of the mini timer.

### 🕶️ Types of changes

### 🤯 List of changes
 - [x] Updated design for the Mini timer
 - [x] Moved Mini timer window from custom `ChromelessWindow` to library's `MetroWindow`
- [x] Removed low-level resizing hack in favor of simple property setter
- [x] Extracted a few user controls that are shared between the timer and the mini timer

### 👫 Relationships
Closes #3486
Closes #2009

### 🔎 Review hints
- Verify that the mini timer design matches the one specified in Zeplin
- Test the mini timer functionality to make sure everything works

Existing issues:
- Mini timer loses keyboard focus when starting time entry. This issue exists on `master` as well. To be fixed in #3725.
- There's still a bit of code duplication between timer and mini timer. Previously the mini timer UI was identical to timer, so they used exact same user control. Now they have several differences. As there was no separation between UI layout and UI logic, it will take a lot of time to properly get rid of code duplication. Created #3730 to fix this issue.


skel35:
Thanks for review @NghiaTranUIT !
Regarding the second point: the height of the mini timer should be bigger when selecting from autocomplete view, otherwise TextBox + project bar won't fit. This is corresponding to the Zeplin design, please check out this mockup from Zeplin:
image
Regarding the first point, you're right, it should be square on the first screenshot, I'll have a look into it.

NghiaTranUIT:
Understand the size of the mini bar, but I suggest that we should make sure the Red Box is square (width=height) 😄

skel35:
To me personally, it looks ok with height>width if it's only in a scenario when we have a project bar.
Please offer your suggestion in #project-win-redesign or in Zeplin. If the designer agrees with you I'd be happy to always keep it square 😄

skel35:
Fixed the issue with increased height in the running state. 
@NghiaTranUIT Ready for re-review.

PR #3738, opened 2020-01-16
Timeline upload issues

NghiaTranUIT:
### 📒 Description
This PR would fix all possible upload timeline issues in order to improve performance / reduce the stress on the server.

Issues:
- [x] Don't upload the Events, which are already uploaded 
- [ ] ~Double check why the chunk 15 mins isn't working anymore~ -> There is no 15 interval for Timeline uploading.

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Fix the way we get the event for uploading (ignore the uploaded=1)
- [x] Change the interval to 5 mins

### 👫 Relationships
No ticket yet

### Make sure the app only upload events, which are not uploaded
1. Install Proxyman.io
2. Start TogglDesktop in Xcode (only work with debug build since the Production build has SSL-pinning to prevent Mitm)
3. Enable Timeline Recorder 
4. Open local database 
5. Observer after the interval, there is `v8/timeline` requests and the body only contains the new event.

Screen Shot 2020-01-16 at 16 17 52

### DIFF
--- a/src/const.h
+++ b/src/const.h
@@ -25,7 +25,7 @@
 #define kTimelineChunkSeconds 900
 #define kEnterpriseInstall false
 #define kDebianPackage (TOGGL_BUILD_TYPE == std::string("deb"))
-#define kTimelineUploadIntervalSeconds 300
+#define kTimelineUploadIntervalSeconds 60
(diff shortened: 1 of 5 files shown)


skel35:
- Restored the previous logic of not uploading the `chunked` events.
- Brought the 1 minute interval back as it doesn't influence the actual uploads.

PR #3779, opened 2020-01-31
Clear auth info before logging

IndrekV:
### 📒 Description
Removed the auth info from logs

### 🔎 Review hints
No auth info should be present in the logs.

### DIFF
--- a/src/https_client.cc
+++ b/src/https_client.cc
@@ -410,6 +410,9 @@ HTTPSResponse HTTPSClient::makeHttpRequest(
         // Request gzip unless downloading files
         poco_req.set("Accept-Encoding", "gzip");
 
+        // Remove Auth info
+        poco_req.set("Authorization", "");
+
         // Log out request contents
         std::stringstream request_string;
         poco_req.write(request_string);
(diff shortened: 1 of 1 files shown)


src/https_client.cc
```diff
@@ -410,6 +410,9 @@ HTTPSResponse HTTPSClient::makeHttpRequest(
         // Request gzip unless downloading files
         poco_req.set("Accept-Encoding", "gzip");
```
skel35: I've also noticed this suspicious line of code #3780 as it doesn't seem to influence anything.
I did some testing without this line and I did some code history investigation. I've discovered only that it used to be before sending the request and was moved here in the commit 1edcb957eb9db03ce6dd3d0c4abc9e73bef61641. I tested most of the basic operations (create/stop TE, create/change project/client/tag) and also tested sending feedback, everything seems to work.

So my vote is to simply remove this line.

src/https_client.cc
```diff
@@ -410,6 +410,9 @@ HTTPSResponse HTTPSClient::makeHttpRequest(
...
 
+        // Remove Auth info
+        poco_req.set("Authorization", "");
```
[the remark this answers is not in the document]
IndrekV: I agree, this is a better solution.

src/https_client.cc
IndrekV: I created a separate issue to look into the gzip compression part #3780

IndrekV:
Ready for another round of review

PR #3791, opened 2020-02-05
[NEED HELP] Memory Leak in Timeline

NghiaTranUIT:
### 📒 Description
This PR would fix all leak as possible from the Timeline in the Library.

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Release all Timeline view

### 👫 Relationships
Closes 

### 🔎 Review hints
1. Check out the build on Xcode
2. Product -> Profile
3. Follow the guideline 
Screen_Shot_2020-02-05_at_16_15_59

4. Make sure there is no Leak on `copy_string()`

### Need Help
- Even though we clear all view lists, but the `copy_string()` is still leaking, which is lesser than before

### DIFF
--- a/src/toggl_api_private.cc
+++ b/src/toggl_api_private.cc
@@ -586,6 +586,13 @@ void timeline_chunk_view_clear(
         timeline_chunk_view_clear(next);
         chunk_view->Next = nullptr;
     }
+    if (chunk_view->Entry) {
+        TogglTimeEntryView *entry =
+            reinterpret_cast<TogglTimeEntryView *>(chunk_view->Entry);
+        time_entry_view_item_clear(entry);
+        chunk_view->Entry = nullptr;
+
+    }
     delete chunk_view;
 }
 
@@ -628,6 +635,12 @@ void timeline_event_view_clear(
             reinterpret_cast<TogglTimelineEventView *>(event_view->Next);
         timeline_event_view_clear(next);
         event_view->Next = nullptr;
+    }
+    if (event_view->Event) {
+        TogglTimelineEventView *event =
+            reinterpret_cast<TogglTimelineEventView *>(event_view->Event);
+        timeline_event_view_clear(event);
+        event_view->Event = nullptr;
     }
     delete event_view;
 }
--- a/src/toggl_api_private.cc
+++ b/src/toggl_api_private.cc
@@ -604,6 +604,10 @@ TogglTimelineEventView *timeline_event_view_init(
 
 void timeline_event_view_update_duration(TogglTimelineEventView *event_view, const int64_t duration) {
     event_view->Duration = duration;
+    if (event_view->DurationString) {
+        free(event_view->DurationString);
+        event_view->DurationString = nullptr;
+    }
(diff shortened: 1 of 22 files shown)


NghiaTranUIT:
## Problem
Look like `timeline_chunk_view_list_clear()` doesn't completely clear all events and sub-event, but the code look good and I couldn't find any errors.

1. Leak occurs when init the event
Screen Shot 2020-02-05 at 20 53 57

2. Not sure why `copy_string()` is leaking
Screen Shot 2020-02-05 at 20 53 50

src/toggl_api_private.cc
```diff
@@ -606,6 +607,14 @@ void timeline_event_view_update_duration(TogglTimelineEventView *event_view, con
     event_view->DurationString = copy_string(toggl::Formatter::FormatDuration(duration, toggl::Format::ImprovedOnlyMinAndSec));
```
MartinBriza: You're leaking memory here by assigning to DurationString without checking if it already contains a pointer.

MartinBriza:
And the same mistake is on this line https://github.com/toggl-open-source/toggldesktop/pull/3791/files#diff-52718a6cf2903beb74f6eae456fb7539L546

src/toggl_api_private.cc
NghiaTranUIT: Make sense. Thank for the hint 👍

MartinBriza:
And chunk_view->EndTimeString in gui.cc is maybe wrong too

NghiaTranUIT:
## Update 1
- all copy_string leaking is gone. Thank @MartinBriza for the help.
- However, there are some leaks about the TogglTimeEntryView in Timeline, but we already clear it after. Not sure what still happens

https://github.com/toggl-open-source/toggldesktop/blob/bb01f468c86fb19d75d0357626c6b845076be05d/src/gui.cc#L581

MartinBriza:
If you have an older macOS version, you could try running the app in `valgrind` (either with `--leak-check=full` or with `--tool=massif`) to have a visualization of where the memory gets stored.

NghiaTranUIT:
## Update 2: Benchmark - 30 min
### Prepare
- Checkout this PR and run TogglDesktop on Production mode via Xcode

### Benchmark Scenario 
1. Start an TimeEntry
2. Turn ON Timeline Recorder and grant the permission
3. Playaround 30 mins by:
4. Navigate between List and Timeline Tab
5. Open/Close Timeline Editor
6. See the Event Infor
7. Switch to other apps (Safari, Zeplin, ....) 
8. Jump back to TogglDeskop and see new Events are recorded.
9. Keep on 30 mins

### Result
#### Before
- It's 80 Mb RAM after the launch.

#### After
- Memory after stoping the running TimeEntry is at 97Mb
- Peak at 111.7 Mb
- Total time: 30 mins
Screen Shot 2020-02-07 at 17 41 37

### Conclusion
It seems that the memory is steadily growing, but it seems reasonable.

IndrekV:
Ok by me. If @MartinBriza gives it a green light then we can merge.

PR #3798, opened 2020-02-11
Change the Shortcut to switch to next/previous Timeline Day

### LINKED ISSUE #3755: Cmd-Left/Right doesn't work for entry description editing when in Timeline tab
### 💻 Environment
Platform: macOS
Toggl Version: 7.5.16

### 🐞 Actual behavior
Cmd-Left/Right doesn't work for entry description editing when in the Timeline tab.
It switches the dates in the calendar which is something I don't want to do when I'm editing the time entry.

### 💯 Expected behavior
Cmd-Left/Right should move my cursor to the beginning/end of the line when I'm editing the time entry description.

NghiaTranUIT:
### 📒 Description
This PR changes the switch day shortcut in order to prevent the conflict with the macOS shortcut in NSTextField (CMD + Left/Right to move the cursor to the begin or end of the word)

### 🕶️ Types of changes

### How does it fix?
Since all NSTextField has their own shortcut (CMD+Arrow or Option+Arrow), so it's better to change the Timeline Shortcut to different one.

### 🤯 List of changes
- [x] Change Day Timeline Shortcut to CMD+SHIFT+Arrow

### 👫 Relationships
Closes #3755

### 🔎 Review hints
1. Open Editor and focus on any TextField (Description, Project, ...)
2. Verify that CMD+Arrow to move the cursor to the start/end of the text
3. Verify that CMD+SHIFT+Arrow will switch the Timeline day.
4. Verify that both actions are not conflicted.

### DIFF
--- a/src/ui/osx/TogglDesktop/test2/Base.lproj/MainMenu.xib
+++ b/src/ui/osx/TogglDesktop/test2/Base.lproj/MainMenu.xib
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="15702" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES">
+<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="15705" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES">
     <dependencies>
         <deployment identifier="macosx"/>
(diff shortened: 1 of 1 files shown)


cyress:
Hi. I'm not sure if this is the right place for my feedback or if this should be a new issue.

The keyboard shortcut `CMD+SHIFT+Arrow` is already used under macOS. With it you can select text from the cursor to the beginning or end of a line.
I use the beta version of Toggl (7.5.49), which contains this adjustment. Every time you want to mark text from the middle of an entry description to the beginning of the line you jump to the timeline view.
Maybe you can use a shortcut which is not already used by macOS for text marking or editing.
That would be great! 😃

PR #3836, opened 2020-02-20
Create Empty TE without stopping the current running TE

### LINKED ISSUE #3827: Clicking on timeline will stop running time entry
### 💻 Environment
Platform: macOS

### 🐞 Actual behavior
If you have a running TE and then click on an empty space in timeline the time entry is stopped and new time entry is created. This is happening because the click creates a new entry and with the old list view creating new one always stopped the old one. In the timeline it should not happen as it is not expected.

### 💯 Expected behavior
Clicking on empty space in timeline should not stop running entry.
New time entry should be created but running entry should not be touched with this action.

### 🔨 Steps to reproduce

1. Start timer
2. Open timeline view
3. Click on an empty space
4. New time entry is created and running entry is stopped.

NghiaTranUIT:
### 📒 Description
This PR would introduce a method to create a TE without stopping the running TE. It's essential to avoid stopping when clicking on the Timeline

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Introduce a method in lib to create a TE
- [x] Add optional param on -start to prevent stopping the running TE if need

### 👫 Relationships
Closes  #3827

### 🔎 Review hints
1. Try to start or continue a TE in List
Verify that:
- The current running TE is stopped
- A new TE is starting (normal behavior)
2. Start an TE
3. Click on the Empty space of Timeline View
Verify that:
- The running TE is not stopped 
- A new Empty TE is created and able to edit

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -2751,7 +2751,8 @@ TimeEntry *Context::Start(
     const std::string tags,
     const bool prevent_on_app,
     const time_t started,
-    const time_t ended) {
+    const time_t ended,
+    const bool stop_current_running) {
 
     // Do not even allow to add new time entries,
     // else they will linger around in the app
@@ -2792,7 +2793,8 @@ TimeEntry *Context::Start(
                           project_guid,
                           tags,
                           started,
-                          ended);
+                          ended,
+                          stop_current_running);
     }
 
     error err = save(true);
@@ -4487,7 +4489,8 @@ void Context::displayPomodoro() {
                                              "",  // project_guid
                                              "pomodoro-break",
                                              0,
-                                             0);  // tags
+                                             0,
+                                             true);  // tags
 
         // Set workspace id to same as the previous entry
         pomodoro_break_entry_->SetWID(wid);
--- a/src/user.cc
+++ b/src/user.cc
@@ -163,9 +163,12 @@ TimeEntry *User::Start(
     const std::string project_guid,
     const std::string tags,
     const time_t started,
-    const time_t ended) {
+    const time_t ended,
+    const bool stop_current_running) {
 
-    Stop();
+    if (stop_current_running) {
+        Stop();
+    }
 
     time_t now = time(nullptr);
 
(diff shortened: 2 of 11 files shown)


skel35:
@NghiaTranUIT Please fix the failing Linux build

NghiaTranUIT:
Fixed @skel35 🙌

src/context.cc
```diff
@@ -4487,7 +4489,8 @@ void Context::displayPomodoro() {
                                              "",  // project_guid
                                              "pomodoro-break",
                                              0,
-                                             0);  // tags
+                                             0,
+                                             true);  // tags
```
skel35: The `//tags` comment is misleading. Please move it to the correct place or just remove it.

src/user.cc
```diff
@@ -163,9 +163,12 @@ TimeEntry *User::Start(
     const std::string project_guid,
     const std::string tags,
     const time_t started,
-    const time_t ended) {
+    const time_t ended,
+    const bool stop_current_running) {
```
skel35: Sorry, but I dislike this change as it decreases maintainability and increases cyclomatic complexity. 
I actually don't like how the method looked previously too: having `duration`, `started`, and `ended` parameter is a confusing API. 
I think this is an opportunity to refactor this code a little. My suggestion is to create a separate method, e.g. `Create` which won't use the current time, won't stop the currently running entry. We can use that method for creating entries in manual mode as well. `Start` method can call `Create` method inside, or they can both call a common private method.

NghiaTranUIT: Thank you for the proposal. 
I follow your approach by 
- Introducing the Create func: For creating a TE without stop the running TE
- Move to Start to private func (_start)
- Revert the change on Start func

Please help me to review a second round 🤚

skel35: Initially, I've had a little different refactoring in mind, so that Start and Create functions are completely separated everywhere, in Context as well. But now I see that it's quite difficult to achieve because there would be a lot of copied code in the Context class.
The current changes don't really do any good without further refactorings, so I suggest to return to your original solution. Sorry for the confusion.

NghiaTranUIT: I just reverts two last commits and it's the original implementation 🙌

PR #3844, opened 2020-02-21
Re-select the last workspace in Project Creation view (mac)

### LINKED ISSUE #3841: While creating new project, default WS isn't selected by default
### 💻 Environment

Platform:  MacOS
OS Version:  10.15.3
Toggl Version:  7.5.42

### 🐞 Actual behavior

When creating new project, default WS isn't always selected in the popup. 

### 💯 Expected behavior

Default WS should be chosen automatically. 

### 📦 Additional info

[Case reference](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/534741/conversations/25829407894)

NghiaTranUIT:
### 📒 Description
Same the title 🤚 
As there is no knowledge about the default workspace, so the solution could be that we re-select the last workspace selection.

### 🕶️ Types of changes
 **Improvement** (non-breaking change which adds functionality)

### 🤯 List of changes
- [x] Save the last Workspace ID and re-select if need

### 👫 Relationships
Closes #3841

### 🔎 Review hints
1. Open Project Creation view. Verify
- The default workspace is correct (Compare with the web) 
2. Select any workspace
3. Click add to create a new project
4. Open Editor again in different TE
5. Open Project Creation View. Verify:
- The Workspace is matched with the last workspace

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -6052,4 +6052,8 @@ void Context::TrackInAppMessage(const Poco::Int64 type) {
     }
 }
 
+Poco::UInt64 Context::GetDefaultWorkspaceID() {
+    return user_->DefaultWID();
+}
+
 }  // namespace toggl
--- a/src/context.h
+++ b/src/context.h
@@ -523,6 +523,8 @@ class TOGGL_INTERNAL_EXPORT Context : public TimelineDatasource {
 
     void TrackInAppMessage(const Poco::Int64 type);
 
+    Poco::UInt64 GetDefaultWorkspaceID();
+
  protected:
     void uiUpdaterActivity();
     void checkReminders();
--- a/src/toggl_api.cc
+++ b/src/toggl_api.cc
@@ -1499,3 +1499,8 @@ char_t *toggl_format_duration_time(void *context,
                           const uint64_t timestamp) {
     return copy_string(toggl::Formatter::FormatDurationForDateHeader(timestamp));
 }
+
+uint64_t toggl_get_default_workspace_id(
+    void *context) {
+    return app(context)->GetDefaultWorkspaceID();
+}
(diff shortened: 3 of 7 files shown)


skel35:
>As there is no knowledge about the default workspace, so the solution could be that we re-select the last workspace selection.
Actually, the library is well-aware of the default workspace. When the time entry is created without a project, then its workspace is the default workspace. We only need to pick it up from the time entry.

NghiaTranUIT:
Thank for the hint, let me dive into the library and get the default workspace. However, I would keep the logic to remember the last Workspace creation since it's useful for anyone who need to quickly create bunch of projects.

NghiaTranUIT:
I improve the ticket by fetching the default workspace 🙌

skel35:
@NghiaTranUIT 
> However, I would keep the logic to remember the last Workspace creation since it's useful for anyone who need to quickly create bunch of projects.

I've looked into the web app and mobile app to see how they do it. In the web app, there is an explicit workspace selection before project creation dialog appears so it doesn't apply to our situation. In the mobile app, the workspace for the project creation is always the user's default workspace. Same in the Windows app. Furthermore, we did not receive user requests specifically for this, and while this change can help some users, it can harm others. Therefore, changing this behavior should be a product decision. Feel free to offer this suggestion to the product manager, but for now, let's keep the standard default project behavior.

skel35:
Please check out my suggested solution for issue #3841:
https://github.com/toggl-open-source/toggldesktop/pull/3860

NghiaTranUIT:
I agree that we this logic is consistent with other platforms.

NghiaTranUIT:
Fixed #3860 by @skel35. Closed this PR.

PR #3845, opened 2020-02-21
[Ready] Implement memory protection and smart containers for all BaseModel classes

MartinBriza:
### 📒 Description
This PR adds a new set of classes to our library. For a more thorough documentation, you can refer to the source file itself, I put a doxygen/javadoc-style comment over each method and class, explaining what it does.

##### `ProtectedContainerBase` 
This is a base class providing a mutex to the two implementations below

##### `ProtectedContainer<T>`
A smart container, currently implemented by: 
 * `std::vector<T>` to contain all `BaseModel`-derived instances
 * `std::map<guid, T>` to allow fast (log(n)) searching by GUID (which will help us especially with the future work on just using GUIDs for everything, instead of IDs and local IDs)

##### `ProtectedModel<T>`
This is a simplified version of `ProtectedContainer` designed to contain only one instance of the `BaseModel`-derived class. Will be used only for `User` and `Settings`.

##### `locked<T>`
A thin pointer wrapper that includes a lock around a mutex provided by `ProtectedContainerBase`. 
Returned from all `Protected*` methods.
It has no copy constructor and copy assignment is not allowed either - it's only possible to `std::move` it. The whole work is designed so you NEVER need to store it anywhere and it's not easy to extract the internal pointer (intentionally). If you find yourself needing to save a `BaseModel` pointer (for example a `TimeEntry`), you should save and search by its GUID instead.

##### Changes to `BaseModel` (and all derived classes)
Every `BaseModel`-derived class constructor takes a pointer to a `ProtectedContainerBase` instance to know "who" is its parent. The constructors were also made private and all instances should be created by using the `create(...)` method in `ProtectedContainerBase` classes. The `parent` argument is passed to the automatically.

#### Usage
Most of current `ProtectedContainer` instances are now in `RelatedData`, for example like this:
```c++
ProtectedContainer<TimeEntry> TimeEntries { this };
```
You can then do the following:
```
locked<TimeEntry> te = TimeEntries.create();
te->setSomething();
locked<TimeEntry> anotherTe = TimeEntries.byGUID("abcdefg");
// .. etc
```

### Current state
This is not complete yet, I'm opening a PR so I hear your opinions on this.
The following things are or need to be done according to the tickmarks:
- [x] Port `memory.h` over to this project
- [x] Make it build
- [x] Reenable Timeline (had to disable it because working with it has a weird API)
  - [ ] Fix Timeline tests
- [x] Reenable tests
- [x] Port over the `memory.h` test I have in my repo
- [ ] Remove unneeded methods
- [x] Improve handling of missing GUID fields
- [x] Smarter way to sort elements by a model value
- [ ] More efficient sorting and/or insertion

Notice some of the points are not implemented yet:
 - I'm leaving cleanup for later phases of this project
 - For making the container more efficient, we'll need to evaluate if we need to keep the order of insertion of items like Tags because then we could just switch to `std::set` that has (in comparison to `std::vector`) waaaaay better complexity for insertion (while also keeping the items sorted for looping over them if we provide a sorting predicate)

### 🕶️ Types of changes

### 👫 Relationships
Should fix some crash bugs, especially those that are hard to track but it introduces new issues in exchange.

Some things are left to be fixed in the future, especially how `Settings` and `Timeline...` items are created and passed around the library.

Future design plans are also to create constructors for `BaseModel` classes that initialize the members straight away (which would let us insert them to the right position right away instead of doing that after the fact). That means the constructor would take arguments like: UID, Name, Duration, etc. Another way to use this is to construct them straight from the JSON (by passing the `Json::Value` as an argument to the constructor) or from the database (by passing a `Poco::RecordSet`).

### 🔎 Review hints
There are, say, 6 main parts of the code to review:
 - `src/util/memory.h` that includes THE THING itself
 - `src/related_data.{h,cc}` for changes and consolidation
 - `src/related_data.{h,cc}` - `CompareClients` and `CompareProjects` methods in particular because I'm still not sure if this'll keep the order right
 - `src/related_data.{h,cc}` and `src/user.{h,cc}` for switching the hierarchy (`RelatedData` now owns `User`)
 - `src/context.{h,cc}` - one thing stands out here - we're not storing the `TimeEntry` pointer for pomodoro break but its GUID instead
 - everything else - changed pure pointers to `locked<T>` (or `auto` when I was lazy, it doesn' matter) and C++98 iterators to either `auto` or `for (auto item : container)` loops

The overall code structure stayed more or less the same so it's mostly about changing one boilerplate for another all over the library.


MartinBriza:
Bump, force push. Made the whole thing build again but it seems to freeze right after starting a time entry, probably because of an underlying race condition.

MartinBriza:
After 380c4cd9376fc8294a8cb7e259b1f24f96d7e881, there will be no more force pushing. Now the app works mostly fine, except for long delays when doing stuff. This happens because some methods can access User data directly without locking the User instance and some others have to lock User first. I'll fix this by inserting the `User` instance into `RelatedData` to have a flatter dependency graph.

MartinBriza:
Tests are mostly working again.

MartinBriza:
This branch is now ready to be reviewed. I'll fix build on macOS and Windows tomorrow (or feel free to do it yourself and push your change yourself, basically just switching to C++17 should do the trick).

I changed the target branch to `kingfisher/master` because without other changes, this is not ready for production. Another step to make this more usable will be to flatten the thread layout and move as much stuff away from the `User` class as possible because every single thing (like idle notification, error display, etc.) is fighting over access to it, resulting in pretty noticeable lag (that's okay) or even freezing (that's not okay and should be fixed) sometimes.

The gist of this PR is still in the `src/util/memory.h` file and in `src/related_data.h`.

MartinBriza:
And now it even builds and starts on all platforms.

NghiaTranUIT:
@MartinBriza if you're happy, we can merge this PR and continue the work 👍

MartinBriza:
> * Overall the main features are remained and I was able to start/stop TEs, create Project and tracking time.
> * There are a little freeze when switching between List and Timeline (30-40 Time Entries): I'm not sure whether or not it's the overhead when we introducing new Lick wrapper. I performed a Profile in both the `master` and this PR, but the result are similar.

The freezes are mostly caused by almost every thread locking the `User` instance basically all the time. In a future PR, I'm going to address this by restructuring how the data is laid out, so not as many threads need to access `User`.

As for merging this, it's up to you, personally I'd like to have more eyes on this but I'm aware there are other PRs that need revies as well.

PR #3847, opened 2020-02-21
Project ToolTip UI

### LINKED ISSUE #3558: Project tooltip new UI
### 💻 Environment
Platform: Windows

### 📒 Description
Implement project tooltip according to new UI instead of plain text tooltip.
Should look like this: 
image

skel35:
### 📒 Description
Implementation of the custom Project ToolTip based on the new design.

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Implement new general ToolTip style
- [x] Implement new Project ToolTip design
- [x] Use new Project ToolTip design for projects and tasks in the autocomplete list
- [x] Use new Project ToolTip design for project label (Timer, MiniTimer, Time Entry List)

### 👫 Relationships
Closes #3558

### 🔎 Review hints
Verify that the new project tooltip is shown:
- for projects and tasks in the autocomplete list
- for project label (Timer, MiniTimer, Time Entry List)
Verify that the new project tooltip matches the design.
Verify that other tooltips adjust to the Light/Dark mode correctly.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/TogglDesktop.csproj
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/TogglDesktop.csproj
@@ -655,6 +655,7 @@
     <Page Include="ui\Resources\DesignUpdate\TabControl.xaml" />
     <Page Include="ui\Resources\DesignUpdate\TextBox.xaml" />
     <Page Include="ui\Resources\DesignUpdate\ToggleSwitch.xaml" />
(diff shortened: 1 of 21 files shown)


skel35:
@NghiaTranUIT You've actually found a legit bug, it happened because the description label was still visible, even though it was empty, so the user could see an empty tooltip when hovering over the narrow left part of the time entry.
It is a trivial fix, so I won't ask to re-review again.
Thanks for the review!

Issue #3459, opened 2019-10-23
Migrate Windows app from .net framework 4.x to .net core 3


Abdulbaqi-Alshareef:
### 💻 Environment

Platform: Windows

### 📒 Description
As the team is planning to re-design the Windows app, migration to .net core 3 should be considered.

### ⭐️ Why do you want this?
To get all benefits that .net core 3 offer over .net framework 4.x 
1- Better performance.
2- Can use C# 8 with all its features 
3- Access to latest APIs.
and more
https://docs.microsoft.com/en-us/dotnet/core/whats-new/dotnet-core-3-0

skel35:
.NET Core is definitely the way to go in the future for our Windows app.
I'll list the blockers that we need to solve before we can make the migration:
- [ ] Our dependency `Hardcodet.NotifyIcon.Wpf` hasn't released a .NET Core version yet. https://github.com/hardcodet/wpf-notifyicon/issues/4. 
- [ ] We're planning on using `MahApps.Metro` when implementing the Windows design update and they haven't released a stable version with .NET Core 3 support https://github.com/MahApps/MahApps.Metro/projects/3
- [ ] Some users are on old Windows versions unsupported by .NET Core 3 (Windows 7 pre-SP1 and earlier, Windows 8, Windows 10 pre-1607). We should make sure the percentage of these users is not critical (less than a few percents) and consider warning them somehow.
- [ ] Auto-update UI/UX. We're currently having a silent update that works ok as our app installs itself and restarts pretty fast. The update to .NET Core app will take more time as we have to include either cold start if we're going with self-contained app or the installation of the .NET Core runtime. Now we don't even have any progress bar when performing the auto-update.

PR #3850, opened 2020-02-24
Update to MahApps.Metro 2.0

### LINKED ISSUE #3770: Maximizing window is not working properly when transparency is on
### 💻 Environment
Platform: Windows
Toggl Version: Windows design update feature branch

### 🐞 Actual behavior
The window is not rendered correctly and is impossible to use until restored back.

### 💯 Expected behavior
The window should be rendering correctly when maximized.

### 🔨 Steps to reproduce
Method 1. Double-click on the titlebar.
Method 2. Right-click on the titlebar -> Maximize.

skel35:
### 📒 Description
Update the MahApps.Metro dependency to version 2.0.
Benefits:
- Bugfixes and updates compared to previous version 1.6.5.
- Will be easier to migrate to .NET Core 3 in the future
- #3770 does not occur if the Maximize button is not shown (and it should not be shown according to the design).

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Update to MahApps.Metro 2.0
  - [x] Update old resource names to new names
  - [x] Replace old `System.Windows.Interactivity` to newer open-source `Microsoft.Xaml.Behaviors.Wpf`
  - [x] Fix the style of titlebar buttons (hover/pressed states in particular)
  - [x] Copy the combobox virtualization fix from the MahApps.Metro (as our ComboBox template had the copy of MahApps.Metro's template with modifications)

### 👫 Relationships
Closes #3770
Point 2 of #3459

### 🔎 Review hints
- Verify #3770 does not occur when maximizing the window (double-click the titlebar)
- Click around the app to see if anything looks visually off or crashes in the runtime (I've tested it myself for a couple of hours)

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/TogglDesktop.csproj
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/TogglDesktop.csproj
@@ -871,7 +871,7 @@
       <Version>1.0.8</Version>
     </PackageReference>
     <PackageReference Include="MahApps.Metro">
-      <Version>1.6.5</Version>
+      <Version>2.0.0-alpha0660</Version>
     </PackageReference>
     <PackageReference Include="NHotkey.Wpf">
       <Version>1.2.1</Version>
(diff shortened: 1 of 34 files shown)


skel35:
@NghiaTranUIT We're currently on .NET Framework 4.6.1 (master still on 4.5) and the idea is to migrate to .NET Core 3 in the future. .NET Core 3 does not ship with Windows, so we will have to ship the .NET Core runtime with the app itself. I think we should improve our auto-update UX before we're doing it because the application size will be significantly bigger (~50-100 Mb estimated installer size) and so the installation will be slower. #3459 is the issue to track, but we're not making the switch yet.

PR #3854, opened 2020-02-25
FixTimeline Editor doesn't update TE data properly

### LINKED ISSUE #3829: Create new TE from Timeline doesn't work properly
### 💻 Environment

Platform: macOS 
OS Version: 10.14.3
Toggl Version: 7.5.16 and 7.5.42

### 🐞 Actual behavior

1. Create a new entry from the Timeline 
2. Add a new project 
3. Add a new tag (“Language”) to the project 
4. No response from the "create new tag“ button 
5. Try "No." for a couple times 
6. Esc 
7. Timeline entry was added 
8. Open entry and see multiple “Language” tags added

Tested on my desktop app and got a similar result. Only in my case the TE was created without a Project nor a tag when checking database. Plus app showed "Data you are sending..." error (probably cause I'm not an admin in WS I was testing with). 
Screenshot 2020-02-18 at 17 42 56
Screenshot 2020-02-18 at 17 44 44

### 🔨 Steps to reproduce

1. 
2. 
3. 

### 📦 Additional info
Logs with [this case](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/3724360/conversations/25817385430).

NghiaTranUIT:
### 📒 Description
There are some case that the Timeline Editor doesn't update the TE properly.
**GIF**

**How to reproduce:**
1. Open Toggl and switch to Timeline Tab
2. Click on any TE to open the Editor
3. Click again on this TE to close 
4. Click again on this TE to open the Editor
5. Try to select any TE, which has different Project
6. Observe: The Timeline Pill is updated correctly, but the Editor is not

### Problem
We don't have any method to call the Library to close the Editor in order to reset some values. Therefore, when the Timeline Editor is closed, the Library doesn't know it state. Then, at the second time, when we update the TE, the Library doesn't send the updated TE to the Editor -> Missing data

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Add close method in library and reset some values
- [x] Call Close when the Editor in Timeline is closed.
- [x] Remove some unusual log

### 👫 Relationships
Similar with user bug. Closes  #3829

### 🔎 Review hints
1. Open Timeline
2. Try to open/close the Editor
3. Select some TE and verify that:
- The Editor is updated correctly.

### DIFF
--- a/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/DesktopLibraryBridge.h
+++ b/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/DesktopLibraryBridge.h
@@ -88,6 +88,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void)startEditorAtGUID:(NSString *)GUID;
 
+- (void)closeEditor;
+
 - (void)setEditorWindowSize:(CGSize)size;
 
 - (CGSize)getEditorWindowSize;
--- a/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/DesktopLibraryBridge.h
+++ b/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/DesktopLibraryBridge.h
@@ -40,8 +40,6 @@ NS_ASSUME_NONNULL_BEGIN
 							 projectID:(uint64_t)projectID
 						   projectGUID:(NSString *)projectGUID;
 
-- (void)togglEditor;
-
 - (void)updateTimeEntryWithDescription:(NSString *)descriptionName guid:(NSString *)guid;
 
 - (void)updateTimeEntryWithTags:(NSArray<NSString *> *)tags guid:(NSString *)guid;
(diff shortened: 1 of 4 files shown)


NghiaTranUIT:
Removed the changed on lib and use `toggl_view_time_entry_list` instead 🙌

skel35:
@NghiaTranUIT You have still created a duplicate wrapper method, it already has a wrapper in `DesktopLibraryBridge` called `togglEditor()`

NghiaTranUIT:
You're truly a Hawkeye that we need in the Avenger team @skel35 🙌

src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/DesktopLibraryBridge.h
```diff
@@ -40,8 +40,6 @@ NS_ASSUME_NONNULL_BEGIN
 							 projectID:(uint64_t)projectID
 						   projectGUID:(NSString *)projectGUID;
 
-- (void)togglEditor;
```
NghiaTranUIT: I remove this method since it's never used in any place. Meanwhile, the name is mis-used.

skel35: makes sense 👍

PR #3869, opened 2020-03-02
Fix mini timer minimum size bug

### LINKED ISSUE #3867: Mini timer tags/billable shown instead of start/stop button when width is small
### 💻 Environment
Platform: Windows
Toggl Version: `feature/winui/master` branch

### 🐞 Actual behavior
Screenshot_2020-02-28_at_13 16 12

### 💯 Expected behavior
Duration and Start/Stop button should be shown fully instead of "$" or "tags" icons.

### 🔨 Steps to reproduce
1. Start an entry which is billable and/or contains some tags
2. Open mini timer
3. Resize the mini timer to the minimum width

skel35:
### 📒 Description
This PR fixes the mini timer minimum size bug.

This also makes ProjectLabelViewModel and TimeEntryLabelViewModel immutable which generally leads to simpler code. Making them immutable does not have any noticeable performance impact because the mutations occurred rarely and usually involved multiple properties at a time.

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Move tags and billable icons into TimeEntryLabel
- [x] Make ProjectLabelViewModel read-only
- [x] Make TimeEntryLabelViewModel read-only
- [x] Remove TimeEntryInfoPanel user controls as it is now just a duration TextBlock

### 👫 Relationships
Closes #3867

### 🔎 Review hints
- Verify that #3867 does not reproduce.
- Try updating the properties of the time entries and make sure they get properly updated in the timer and the time entry list.


skel35:
@NghiaTranUIT Good catch, thanks! Turns out the text wrapping was on and it caused this issue.
Please re-review 🙇

PR #3870, opened 2020-03-03
Add caching for GitHub Actions Windows builds

skel35:
### 📒 Description
Adds caching for GitHub Actions Windows builds

### 🕶️ Types of changes
- **Improvement** (non-breaking change which adds functionality)

### 🤯 List of changes
- [x] Add caching for POCO bin and lib folders
- [x] Build `TogglDesktop.NoPoco.sln` if there is a cache hit, build `TogglDesktop.sln` otherwise.
- [x] Install NSIS only when actually building the installer

### 🔎 Review hints
Verify the builds are successful and the running time decreased.


skel35:
When we merge this to `master` the installer gets built so it will be automatically checked 😄 
The change is too trivial to test is separately IMO.
If it doesn't work then I can revert the change and continue working on this.

PR #3885, opened 2020-03-09
Bug/GitHub action update release update json

NghiaTranUIT:
### 📒 Description
This PR would fix the issue that `update.json` and `release.json` don't commit and update properly (for DEV macOS release procedure)

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Fix the directory to generate the release and update.json

### 👫 Relationships
No Ticket

### 🔎 Review hints
Since the PR change is super tiny but It's hard to review this PR until we merge to the master (to trigger the release procedure)


NghiaTranUIT:
This PR is ready for reviewing 🙌

dist/parse_releases.sh
```diff
@@ -11,7 +11,7 @@ var data = require('./releases.json'),
   channel = process.env['CHANNEL'],
   version = process.env['VERSION'],
   timestamp = process.env['TIMESTAMP'],
-  dirpath = process.env['DIRPATH'] || "",
+  dirpath = process.env['DIRPATH'] || "tmp/",
```
skel35: After reading the code it looks like it should fail earlier, at line 8:
https://github.com/toggl-open-source/toggldesktop/blob/0c146ba3071870bbe827447b707525476dbda2ca/dist/parse_releases.sh#L8
because I don't see how `releases.json` file ends up in the root folder of the repo.

I would suggest calling `cd tmp` before executing the `parse_releases.sh` script so that we don't care about other path dependencies.

skel35: Ah, my mistake. Line 8 is actually reading the file from the `tmp` folder, that's why it works correctly

NghiaTranUIT: Yup, that's right 👍

PR #3886, opened 2020-03-09
Clean up pushChanges and make it generic for all models

MartinBriza:
### 📒 Description
This is another step in the direction of having everything as generic as possible for all models we're working with in the library.
Request construction is now delegated to `BaseModel` (and virtual methods are used to provide relevant data by the actual implementations) and instead of doing it in a ton of methods, I moved everything into lambdas - this is not necessarily the cleanest way to do this but (from my point of view) it helps to see the separation of business logic in the `Context` class so we can eventually rip this whole syncing thing out and have it as a completely separate class or something.

### 🕶️ Types of changes
Doesn't really fit neither of these categories :D

### 👫 Relationships
Part of the library overhaul slash kingfisher project.

### 🔎 Review hints
See if pushing changes still works as expected. Tell me if you like the code or not.

PR #3895, opened 2020-03-11
Implement automatic (and fast) sorting for RelatedData models

MartinBriza:
### 📒 Description
This change makes use of `std::set` as the main storage container for our data models instead of `std::vector`. The biggest benefit is that we now get element sorting "for free" if we want it. This is especially demonstrated by the next-to-last commit (c9565ed) where I only did the following:
 * added a `CompareTimeEntries` function to `RelatedData` and passed it to the `ProtectedContainer` constructor
 * used  `GetContainer()->shift(this)` in `TimeEntry::SetStart` because it changes the sorting parameter
 * and finally, disabled sorting when constructing the time entry view for the UI to demonstrate it actually works

While this makes some operations (accessing by index) slower, it makes others MUCH faster - now inserting, searching and deleting have O(log(n)) complexity instead of O(n). We also don't need to sort the elements each time we want to construct a list of them (as demonstrated above).

I didn't change all usages of `std::sort` throughout the code but it's planned for the final cleanup.

### 🕶️ Types of changes

### 👫 Relationships
Part of the kingfisher overhaul
Merge after #3886 

### 🔎 Review hints
Test if UI elements are still sorted in an expected way.


MartinBriza:
This just doesn't happen to me on Linux and I left my macbook at home. Can you please post the full backtrace? I'm especially interested at which lines are there on lines 4, 5 and 6.

NghiaTranUIT:
### The crash
Sorry but I couldn't find the best way to print out useful backtrace. They were sketchy so I take a screenshot instead.
Screenshot: https://www.dropbox.com/s/oysi41didfechja/Screenshot_TogglDesktop.zip?dl=0

**Warning from Xcode at the crash point**
Screen Shot 2020-03-12 at 16 19 23

### Incorrect behavior (Missing data)
Same account but login from `master` and `kingfisher/sorting` that have different Time Entry List.

Confirmed:
- Same TogglDesktop scheme (for staging)
- Same Endpoints (toggl.space)
- Check console log, it's debugging mode. I noticed by reading the path of the database. it's development database

**Master**
Screen Shot 2020-03-12 at 16 06 39

=> There is no Database Constraints violation

** kingfisher/sorting**
Screen Shot 2020-03-12 at 16 05 20

=> There is a Database Constraints violation issue after going to the main view
=> Sometime it's crashed (https://github.com/toggl-open-source/toggldesktop/pull/3895#pullrequestreview-373363297)

## How to reproduce
1. Checkout master
2. Clear and select TogglDesktop scheme (staging)
4. Start the app and Logout all account if neeed
5. Login to X Account (PM @NghiaTranUIT for the credential)
6. Remember the Time Entry List (order and data)
7. Logout
8 Stop the app
9. Checkout `kingfisher/sorting`
10. Start the app and login with the same account
11. Compare the Time Entry List

=> It's different and get the database constraints error.

MartinBriza:
Thanks, Nghia, I managed to fix this crash in the latest commit.

NghiaTranUIT:
Thank for addressing the crash. However, I usually encounter the hang when opening the Editor on a running Time Entry.

It's related to this report https://github.com/toggl-open-source/toggldesktop/pull/3886#pullrequestreview-373344996 from https://github.com/toggl-open-source/toggldesktop/pull/3886

Video: https://www.dropbox.com/s/uxahd0djjnc3dig/TogglDesktop_Freeze_2.mov?dl=0

NghiaTranUIT:
Btw, the issue `Incorrect behavior (Missing data)` are gone too. The Time Entry list is the same for  master and this PR 👍

NghiaTranUIT:
Ticket for the deadlock: https://github.com/toggl-open-source/toggldesktop/issues/3920

PR #3914, opened 2020-03-23
Fix possible crash when selecting the default workspace (mac)

### LINKED ISSUE #3909: Crash on macOS when opening time entry editor
### 💻 Environment
Platform: macOS
Toggl Version: 7.5.62

### 🐞 Actual behavior
Crash

### 💯 Expected behavior
No crash

### 🔨 Steps to reproduce
Unknown, but there are at least 59 [Bugsnag reports](https://app.bugsnag.com/toggldesktop/toggldesktop/timeline?filters[event.since][0][type]=eq&filters[event.since][0][value]=30d&filters[error.status][0][type]=eq&filters[error.status][0][value]=open&filters[search][0][type]=eq&filters[search][0][value]=TogglDesktop.ProjectCreationView&filters[search][1][type]=eq&filters[search][1][value]=selectDefaultWorkspace&pivot_tab=event)

NghiaTranUIT:
### 📒 Description
I couldn't reproduce this bug and found no race condition issue on macOS UI, but we have some in TogglLibrary

Screen Shot 2020-03-23 at 14 46 38

The [crash log](https://app.bugsnag.com/toggldesktop/toggldesktop/timeline?filters%5Bevent.since%5D%5B0%5D%5Btype%5D=eq&filters%5Bevent.since%5D%5B0%5D%5Bvalue%5D=30d&filters%5Berror.status%5D%5B0%5D%5Btype%5D=eq&filters%5Berror.status%5D%5B0%5D%5Bvalue%5D=open&filters%5Bsearch%5D%5B0%5D%5Btype%5D=eq&filters%5Bsearch%5D%5B0%5D%5Bvalue%5D=TogglDesktop.ProjectCreationView&filters%5Bsearch%5D%5B1%5D%5Btype%5D=eq&filters%5Bsearch%5D%5B1%5D%5Bvalue%5D=selectDefaultWorkspace&pivot_tab=event) and the [code base](https://github.com/toggl-open-source/toggldesktop/blob/master/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/Project/ProjectCreationView.swift#L308-L313) are totally fine (Use guard statement to safely cast objs).

The only thing it could cause the crash is in 
```
func selectRow(at index: Int) { }
```
since it might access an invalid range.

### 🕶️ Types of changes

### 🤯 List of changes
- Get item safely from the workspace arrays

### 👫 Relationships
Closes #3909

### 🔎 Review hints
- Play around (Start/Stop/Create Project/Workspace, ...) and make sure no crash

### DIFF
--- a/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/AutoComplete/DataSource/AutoCompleteViewDataSource.swift
+++ b/src/ui/osx/TogglDesktop/Feature/TimeEntryEditor/AutoComplete/DataSource/AutoCompleteViewDataSource.swift
@@ -86,8 +86,7 @@ class AutoCompleteViewDataSource: NSObject {
     }
 
     func selectRow(at index: Int) {
-        guard index >= 0 else { return }
-        let item = items[index]
+        guard let item = items[safe: index] else { return }
         delegate?.autoCompleteSelectionDidChange(sender: self, item: item)
     }
 


skel35:
@NghiaTranUIT If `selectRow` method actually caused the crash, is there an explanation as to why it's absent in the stack trace?  Could it happen because the method gets inlined in the Release build (I'm not sure how Swift compiler works)?
image

NghiaTranUIT:
Yup, it's the thing I concern about how Bugsnag records the stack call of the crash. I'm not sure why it doesn't point exactly the line number, which AppCenter does.

So, I guess the problem is from `selectRow` method since it has unsafe access from the array. On the other hand, the `selectDefaultWorkspace` method is totally fine, there are no race-condition issue, no force cast.

nidarasheed:
[Case reference](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/534741/conversations/26418427841)

PR #3947, opened 2020-03-31
Make Autotracker support multiple window titles

### LINKED ISSUE #3915: Make Autotracker support multiple window titles
Extend the Autotracker feature in the library to support multiple window titles (or parts of a window title, called "term" in the code) for a single autotracker rule.
- Saving to / reading from the database
- Matching multiple titles with timeline events
- Converting into view struct for passing to the platform clients
- Change the API method for adding an autotracker rule to support multiple titles
- Add the API method for editing an autotracker rule
- Make sure backward compatibility is maintained (add DB migration if necessary)

skel35:
### 📒 Description
Make Autotracker rules support multiple window titles

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Saving to / reading from the database
- [x] Matching multiple titles with timeline events
- [x] ~Converting into view struct for passing to the platform clients~ no changes needed 
- [x] ~Change the API method for adding an autotracker rule to support multiple titles~ no changes needed
- [x] Add the API method for editing an autotracker rule
- [x] Make sure backward compatibility is maintained (add DB migration if necessary)
- [x] Add tests to cover the new functionality

### 👫 Relationships
Closes #3915

### 🔎 Review hints
Code review is most important as this PR doesn't really add any user-visible functionality.

Although it is possible to test multiple titles matching:
- Create some Autotracker rule and save it
- Quit the app
- Open `toggldesktop.db` and modify your Autotracker rule's `term` field (`autotracker_settings` table) to include several terms separated by \t character, for example, `slack	firefox` (\t character between `slack` and `firefox`).
- Open the app and check if Autotracker is matching a window by finding just one of the terms.

### DIFF
--- a/src/autotracker.cc
+++ b/src/autotracker.cc
@@ -14,7 +14,7 @@ static const char kTermSeparator = '\t';
 bool AutotrackerRule::Matches(const TimelineEvent &event) const {
     for (const auto& term : terms_) {
         if (Poco::UTF8::toLower(event.Title()).find(term)
-            != std::string::npos) {
+                != std::string::npos) {
             return true;
         }
     }
--- a/src/autotracker.cc
+++ b/src/autotracker.cc
@@ -9,25 +9,37 @@
 
 namespace toggl {
 
+static const char kTermSeparator = '\t';
+
 bool AutotrackerRule::Matches(const TimelineEvent &event) const {
-    if (Poco::UTF8::toLower(event.Filename()).find(term_)
+    for (const auto& term : terms_) {
+        if (Poco::UTF8::toLower(event.Filename()).find(term)
             != std::string::npos) {
-        return true;
-    }
-    if (Poco::UTF8::toLower(event.Title()).find(term_)
+            return true;
+        }
+        if (Poco::UTF8::toLower(event.Title()).find(term)
             != std::string::npos) {
-        return true;
+            return true;
+        }
     }
     return false;
 }
 
-const std::string &AutotrackerRule::Term() const {
-    return term_;
+const std::vector<std::string>& AutotrackerRule::Terms() const {
+    return terms_;
 }
 
-void AutotrackerRule::SetTerm(const std::string &value) {
-    if (term_ != value) {
-        term_ = value;
+void AutotrackerRule::SetTerms(const std::string &value) {
+    if (TermsString() != value) {
+        terms_.clear();
+        if (!value.empty()) {
+            std::stringstream ss(value);
+            while (ss.good()) {
+                std::string tag;
+                getline(ss, tag, kTermSeparator);
+                terms_.push_back(tag);
+            }
+        }
         SetDirty();
     }
 }
@@ -57,7 +69,7 @@ void AutotrackerRule::SetTID(const Poco::UInt64 value) {
 std::string AutotrackerRule::String() const {
     std::stringstream ss;
     ss << " local_id=" << LocalID()
-       << " term=" << term_
+       << " terms=" << TermsString()
        << " uid=" << UID()
        << " pid=" << pid_
        << " tid=" << tid_;
@@ -70,6 +82,17 @@ std::string AutotrackerRule::ModelName() const {
 
 std::string AutotrackerRule::ModelURL() const {
     return "";
+}
+
+const std::string AutotrackerRule::TermsString() const {
+    std::stringstream ss;
+    for (auto it = terms_.begin(); it != terms_.end(); ++it) {
+        if (it != terms_.begin()) {
+            ss << kTermSeparator;
+        }
+        ss << *it;
+    }
+    return ss.str();
 }
 
 }  // namespace toggl
--- a/src/autotracker.cc
+++ b/src/autotracker.cc
@@ -13,10 +13,6 @@ static const char kTermSeparator = '\t';
 
 bool AutotrackerRule::Matches(const TimelineEvent &event) const {
     for (const auto& term : terms_) {
-        if (Poco::UTF8::toLower(event.Filename()).find(term)
-            != std::string::npos) {
-            return true;
-        }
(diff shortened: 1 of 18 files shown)


src/database.cc
```diff
@@ -2461,7 +2461,7 @@ error Database::saveModel(
                       "insert into autotracker_settings(uid, term, pid, tid) "
                       "values(:uid, :term, :pid, :tid)",
                       useRef(model->UID()),
-                      useRef(model->Term()),
+                      useRef(model->TermsString()),
```
MartinBriza, line 2464: Using `useRef` with a r-value (that is a value - not reference - returned from a function) is dangerous. I think in this case there should be just plain getter call without the `useRef` wrapper. Same applies above in the update.

MartinBriza: An alternative would be just storing the delimited string. I see the API is mostly consumed with it anyway.

src/autotracker.cc
```diff
@@ -9,25 +9,33 @@
 
 namespace toggl {
 
+static const char kTermSeparator = '\t';
+
 bool AutotrackerRule::Matches(const TimelineEvent &event) const {
-    if (Poco::UTF8::toLower(event.Filename()).find(term_)
```
MartinBriza, line 13: Where did filename go?

src/context.cc
```diff
@@ -3957,6 +3957,67 @@ error Context::AddAutotrackerRule(
...
+            logger.warning("cannot add autotracker rule, user logged out");
+            return noError;
+        }
+
+        Task* t = nullptr;
+        if (tid) {
+            t = user_->related.TaskByID(tid);
+        }
+        if (tid && !t) {
+            return displayError("task not found");
```
MartinBriza, line 3990: Overall, with these errors, I'd probably add a bit more information, especially since it seems they'll get displayed to the user. Getting just "task not found" or "task does not belong to project" when trying to modify an autotracker rule could be quite confusing to the user.

src/toggl_api.cc
```diff
@@ -1399,6 +1399,20 @@ int64_t toggl_autotracker_add_rule(
     return rule_id;
 }
 
+bool_t toggl_autotracker_update_rule(
+    void *context,
+    const int64_t rule_id,
+    const char_t *term,
```
MartinBriza, line 1405: terms?

src/autotracker.cc
skel35: Currently, we don't match by filename anyway - `TimelineEvent` isn't populated with the filename:
https://github.com/toggl-open-source/toggldesktop/blob/ce67d39051d9d55594384a2b25931aa39152163c/src/window_change_recorder.cc#L79-L83
So I removed it for now to avoid confusion. We can bring it back in a separate PR, double-checking if the UI and support article informs the user of this capability.

src/toggl_api.cc
skel35: 👍

src/autotracker.cc
MartinBriza: :+1:

src/context.cc
skel35: User isn't supposed to see these errors. Non-emptiness of the project/task is validated by clients, and the projects get assigned from the list of current user's projects.
I see your point that the error messages can be improved for exceptional situations, but I don't know which use case to optimize messages for, and we have these error messages all over the place, so that would be a separate task IMO.

MartinBriza: My rationale is that calling `displayError` should happen only to let the user know there's something wrong. These are not errors that just get output to the log but they will bubble through to the UI as well. If the user is actually not supposed to see them, I suggest using `Logger` or returning them to the calling function so the API consumer knows about the mistake they made.

You're right we can't fix them everywhere right now but I think that we should do things right, or at least better, when we're adding new methods to the API.

skel35: Changed those cases to log the errors instead of showing them to the user, using the existing pattern. You're right, no reason to show them to the user even in exceptional situations.

src/database.cc
skel35: I'd prefer to keep a vector of strings in memory, as we're optimizing for terms matching because it can be called much more frequently than the API methods, even though parsing probably won't be slow as users aren't expected to add lots of titles per single rule.
I'll check the Poco::Data docs first so I can better understand potential issues with `useRef` or other ways to do things.

src/context.cc
MartinBriza: Thanks!

src/database.cc
MartinBriza: I totally get that. A completely alternative solution, while introducing C++17 (that will get to TogglDesktop in near future anyway) would be to use `std::string_view` and create a "view" over the delimited single string. This avoids splitting and copying parts of this string when we want to do comparisons. 

However, that'd increase the complexity of this PR a lot and would be a considerable time sink on your part, I assume.

Anyway, as far as I remember, having just `model->TermsString()` in the mentioned part of code (without wrapping it in `useRef`) should be enough. But it's definitely a good idea to consult the docs first.

MartinBriza: One other way occured to me when taking a shower: You could keep the code as is, but have a `static thread_local std::string` variable inside `TermsString` and return a reference to it. That way you're safe.

skel35: Looks like `bind()` is recommended to use in this scenario: https://pocoproject.org/docs/00200-DataUserManual.html
>It is possible to use bind() instead of use(). The bind() call will always create a copy of the supplied argument.

I think I like it a bit more `thread_local std::string` just to keep things simpler in the model class.

`string_view` looks useful, it would be cool to use it later on.

MartinBriza: Alrighty, sounds good. I'll wait for the CI build and merge if it passes.

PR #3949, opened 2020-03-31
Resurrect old Windows<->Library interop tests

skel35:
### 📒 Description
Resurrect old Windows<->Library interop tests.
They are not supposed to run in the CI for now, as they might fail.
But mostly they work and are very convenient to use in the Windows development environment.

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Rewrote old Windows<->Library interop tests using XUnit


skel35:
Will merge this PR to master without review as it doesn't impact the build or any other dev env except Windows, where it simply replaces one non-buildable test project with another runnable but non-stable one.

PR #3964, opened 2020-04-02
Make Autotracker support time range and days of week

### LINKED ISSUE #3855: Do not use hardcoded database and logs file paths
### 💻 Environment
Platform: Windows

### 📒 Description
Our app currently uses a hardcoded database file path `%LOCALAPPDATA%/TogglDesktop/toggldesktop.db` regardless of the directory it is running from. Same for log files.
This means that all our installations and debug runs will share the same database and logs files. This works due to the fact that only one instance of the app can run at the same time, but it has downsides, especially during development. For example, the logs are intermixed betweens production builds and debug builds, running integration tests messes up the local database, etc.

### LINKED ISSUE #3916: Make Autotracker support selected time range and days of the week (library)
Extend the Autotracker feature in the library to support a time range and days of the week restrictions for a single autotracker rule.
- Saving to / reading from the database
- Matching a time range and days of the week with timeline events
- Converting into view struct for passing to the platform clients
- Change the API method for adding an autotracker rule
- Change the API method for editing an autotracker rule
- Make sure backward compatibility is maintained (add DB migration if necessary)

skel35:
### 📒 Description
Make Autotracker support time range and days of week

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Saving to / reading from the database
- [x] Matching a time range and days of the week with timeline events
- [x] Converting into view struct for passing to the platform clients
- [x] Change the API method for adding an autotracker rule
- [x] Change the API method for editing an autotracker rule
- [x] Make sure backward compatibility is maintained (add DB migration if necessary)
- [x] App tests
- [x] API tests

### 👫 Relationships
Closes #3916
Closes #3855 

### 🔎 Review hints

The code review should be the main focus, as this PR doesn't change any user-facing functionality.
The thing I'm most unsure of is the days of week API - I decided to go for `uint8` as a representation of days of week bitset for the reason of API simplicity (1 field instead of 7 fields or a separate struct) and memory efficiency. The downside is that it's not clear which bit corresponds to which days of the week. This is why I added a small API method `toggl_days_of_week_into_uint8()` that can be used to make the conversion or as a reference.

Manual test scenario:
1. Run the app to make sure the database has been migrated. Verify the app is running without issues.
2. Create an autotracker rule and save it.
3. Open the database file and set start time, end time, and/or days of the week to the added autotracker rule.
4. Reopen the app to make sure the edited rule is loaded.
5. See if the start time, end time, and days of the week settings are respected.

### DIFF
--- a/src/autotracker.cc
+++ b/src/autotracker.cc
@@ -13,7 +13,7 @@ static const char kTermSeparator = '\t';
 
 bool AutotrackerRule::Matches(const TimelineEvent &event) const {
     const Poco::LocalDateTime event_time(Poco::DateTime(event.EndTime()));
-    if (!days_of_week_[event_time.dayOfWeek()]) {
+    if (!days_of_week_.none() && !days_of_week_[event_time.dayOfWeek()]) {
         logger().debug("Autotracker rule is not enabled on this weekday");
         return false;
     }
--- a/src/autotracker.cc
+++ b/src/autotracker.cc
@@ -41,7 +41,7 @@ bool AutotrackerRule::Matches(const TimelineEvent &event) const {
         }
     }
 
-    for (const auto& term : terms_) {
+for (const auto& term : terms_) {
         if (Poco::UTF8::toLower(event.Title()).find(term)
                 != std::string::npos) {
             return true;
--- a/src/autotracker.cc
+++ b/src/autotracker.cc
@@ -41,7 +41,7 @@ bool AutotrackerRule::Matches(const TimelineEvent &event) const {
         }
     }
 
-for (const auto& term : terms_) {
+    for (const auto& term : terms_) {
         if (Poco::UTF8::toLower(event.Title()).find(term)
                 != std::string::npos) {
             return true;
--- a/src/autotracker.cc
+++ b/src/autotracker.cc
@@ -12,6 +12,11 @@ namespace toggl {
 static const char kTermSeparator = '\t';
 
 bool AutotrackerRule::Matches(const TimelineEvent &event) const {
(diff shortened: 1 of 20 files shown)


src/gui.h
```diff
@@ -295,11 +295,17 @@ class TOGGL_INTERNAL_EXPORT AutotrackerRule {
     AutotrackerRule()
         : ID(0)
     , Terms("")
-    , ProjectName("") {}
```
MartinBriza, line 298: While this is not a big deal, for a small improvement in readability and reduction in number of lines of code I suggest using inline member initializers instead of initializing in the constructor.
That means instead of listing members in the constructor like this:
`    , ProjectName("")`
You would set the initial value right in the member declaration (below) like this:
`    std::string ProjectName { "" };`

MartinBriza: The constructor then doesn't even have to be declared at all but for clarity sake it's probably better to have it.

src/autotracker.cc
```diff
@@ -12,7 +12,36 @@ namespace toggl {
...
+            Poco::LocalDateTime end(
+                event_time.year(), event_time.month(), event_time.day(), h, m, event_time.second());
+            if (event_time > end) {
+                logger().debug("It's too late for this autotracker rule", " [", event_time.hour(), ":", event_time.minute(), "]", " (allowed until ", h, ":", m, ")");
+                return false;
+            }
+        }
+    }
+
+for (const auto& term : terms_) {
```
MartinBriza, line 44: indent?

skel35:
Addressed the abovementioned issues.
Ready for re-review.

PR #3965, opened 2020-04-03
Menu isn't responsive after launching the app

### LINKED ISSUE #3873: Toggl Desktop app doesn't let you click on Menu button after being launched
### 💻 Environment

Platform: MacOS
OS Version: 10.15.3
Toggl Version: 7.5.42

### 🐞 Actual behavior

If you open Toggl desktop  (after it was shut down) and then try to click on menu buttons at the top, you cannot.  They do not respond. Clicking outside the Toggl Desktop window, then back on it, and then clicking menu items works.

### 💯 Expected behavior

The menu should work. 

### 🔨 Steps to reproduce

1. Quit Toggle Desktop  
2. Open it back up 
3. While the focus is still on the app, click Menu buttons. 
4.  Click out of the desktop app
5. Click on the desktop app
6. Now try the menu. 

### 📦 Additional info

[Slack](https://toggl.slack.com/archives/C029WCX0T/p1583304296076200)

NghiaTranUIT:
### 📒 Description
This PR would fix the issue which causes the Menu Bar is not responsive

## Problem
We have `LSUIElement` in TogglDesktop and TogglDesktop-AppStore which cause the bug.
[LSUIElement](https://developer.apple.com/documentation/bundleresources/information_property_list/lsuielement?language=objc) intends for the app, which doesn't have a dock icon. Usually an app from Menu Status Bar. However, TogglDesktop is a normal app (have Dock, Main Window and Status Menu), so it causes a bug.

In the past, we override by using `[NSApp setActivationPolicy:NSApplicationActivationPolicyRegular];`, but it doesn't work anymore

### 🕶️ Types of changes

### 🤯 List of changes
- Remove LSUIElement and [NSApp setActivationPolicy]

### 👫 Relationships
Closes #3873

### 🔎 Review hints
- Make sure we can click on the menu bar when the app is launched
- All menu bar functionality and shortcuts are working as usual

### DIFF
--- a/src/ui/osx/TogglDesktop/TogglDesktop-AppStore-Info.plist
+++ b/src/ui/osx/TogglDesktop/TogglDesktop-AppStore-Info.plist
@@ -59,8 +59,6 @@
 	<string>public.app-category.productivity</string>
 	<key>LSMinimumSystemVersion</key>
 	<string>10.11</string>
-	<key>LSUIElement</key>
-	<true/>
 	<key>NSHumanReadableCopyright</key>
 	<string>Copyright © 2017 Toggl Desktop developers. All rights reserved.</string>
 	<key>NSMainNibFile</key>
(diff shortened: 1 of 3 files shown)


IndrekV:
Good job! Simple solution.

PR #3969, opened 2020-04-06
Use Formatter::FormatDuration in date header

### LINKED ISSUE #3795: Total time for the day not being displayed in a decimal format (unlike web app and mobile app)
### 💻 Environment

Platform:  macOS
OS Version: OS X 10.15.3
Toggl Version: 7.5.16

### 📒 Description

Total time for the day will be displayed in a classic format (HH:MM) even if the user has decimal format selected in the settings, this causes users to think there are discrepancies between web app and desktop app.

**Web app**
Screenshot 2020-02-09 at 10 37 13

**Desktop app**
Screenshot 2020-02-09 at 10 39 30

jiri:
### 🕶️ Types of changes

### 🤯 List of changes
Fixes Formatter call in `context.cc`.

### 👫 Relationships

Fixes #3795.

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -744,9 +744,10 @@ void Context::updateUI(const UIElements &what) {
                                     group_entry_view.Duration = toggl::Formatter::FormatDuration(
                                         group_entry_view.DurationInSeconds,
                                         Formatter::DurationFormat);
-                                    group_entry_view.DateDuration =
-                                        Formatter::FormatDurationForDateHeader(
-                                            date_durations[group_entry_view.DateHeader]);
+                                    group_entry_view.DateDuration = toggl::Formatter::FormatDuration(
+                                        date_durations[group_entry_view.DateHeader],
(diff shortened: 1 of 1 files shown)


skel35:
I've thought a bit more about this PR.
I think it shouldn't be merged into `master` by itself due to several issues:
1. Our day header duration is updated once every 10 seconds. This PR shows date header duration with seconds precision, which means the user will be seeing these updates (e.g. "06:51:10", "06:51:20", "06:51:31") instead of smooth time flow.
2. The day header duration in the web and mobile apps, unlike the desktop app, does not count a currently running entry. If we're removing inconsistencies we should remove this one. This will automatically address p.1.
3. The design may rely on the old format which is longer than the new format. This is affecting the tray tooltip in the Windows app, it should be tweaked before we merge it.


I suggest closing this PR without merging and implement a more thorough fix.
I'll create a ticket mentioning this PR.

PR #3973, opened 2020-04-07
Bug/3956

### LINKED ISSUE #3956: Autocomplete dropdown font is too big and ugly
### 💻 Environment
Toggl Version: `feature/linux-qml` HEAD

### 🐞 Actual behavior
Pretty much described in the subject. The dropdown should fit the overall UI better. Inspiration can be found in the macOS app.

### 📦 Additional info
#3618

jiri:
### 📒 Description
Tweaked the text size, margins and weights to better resemble the macOS version.

### 🕶️ Types of changes

### 👫 Relationships

Closes #3956

### 🔎 Review hints

Screen Shot 2020-04-07 at 10 56 02


MartinBriza:
Can you please give the lines a tad more contrast? I don't think they're pronounced enough. 

Also, I think there should be some sort of a separator at the top of the list, be it a line or a shadow.

jiri:
I originally didn't add the top separator because it's not in the macOS version, but I added one now :) I also tweaked the contrast a bit.

MartinBriza:
Yes, it's not there but on macOS but the dropdown doesn't fill the entire width of the window and it has a slightly different color than the timer, so the difference is more noticeable. 
However, this doesn't look very good. I think you should use a shadow as mentioned in my original comment. Experiment if the timer casting a shadow on the dropdown looks better or the dropdown casting a shadow on the timer does.

MartinBriza:
I also think the separator line is now too pronounced when using a light theme.

MartinBriza:
There are no commits addressing my last comments so I'm closing this and will work on it myself later.

PR #3998, opened 2020-04-14
Reorganize the library codebase into folders

MartinBriza:
### 📒 Description
This change shifts the library files around so things that logically belong together are in the same folder. It also creates Groups in the XCode project. I didn't do anything like that for Windows because it's absolutely annoying but I'm willing to do it if Taras wants to have everything as neat as we do on Linux and macOS.

I also removed a bunch of lines in the Makefile to make it build on my machines but I'm pretty sure that's just unnecessary legacy stuff.

### 🕶️ Types of changes
- Uhh, nothing above, it's just a part of a cleanup

### 🔎 Review hints
Everything should build and everything should work as before.


third_party/poco/config.make
```diff
@@ -1,12 +1,12 @@
 # config.make generated by configure script
 POCO_CONFIG = Darwin64-clang-libc++
-POCO_BASE = /Users/nghiatran/Documents/Work/Toggl/toggldesktop/third_party/poco
-POCO_BUILD = /Users/nghiatran/Documents/Work/Toggl/toggldesktop/third_party/poco
+POCO_BASE = /Users/martinbriza/code/toggldesktop/third_party/poco
```
NghiaTranUIT, line 3: Is it safe to be hard code? I assume that config.make is automatically generated from your current machine.

MartinBriza: Your machine was there before and I probably commited this by mistake. Let me remove any changes from `third_party` before merging.

skel35:
@MartinBriza I'll push the folders grouping for Windows solution 😄

MartinBriza:
At this point I think it really is not worth it. If we find ourselves using more include paths, we could try to optimize it. 
However I still hope we're going to switch to CMake everywhere for the library at some point. That would render any of this tweaking unnecessary.

MartinBriza:
There are two approvals so I guess this is OK to merge.

PR #4050, opened 2020-05-05
Attempt at fixing the crash in EventQueue destructor (lib)

MartinBriza:
This could fix the crash.

The rationale behind this is checking if the thread is valid. If it is, we need to determine if the destructor was called from the main thread or not.

I definitely don't recommend calling it from the EventQueue thread though, using the main thread to destroy everything seems to be the right call.


skel35:
not relevant anymore as another solution was chosen

PR #4053, opened 2020-05-07
Add duration textbox to the mini timer

### LINKED ISSUE #4052: Add duration text field to the mini timer
### 💻 Environment
Platform: Windows

### 📒 Description
Add the duration text field to the mini timer, similar to the one in the old Windows app design.

skel35:
### 📒 Description
Add duration textbox to the mini timer

### 🕶️ Types of changes

### 👫 Relationships
Closes #4052

### 🔎 Review hints
Thoroughly test the mini timer feature.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/controls/MiniTimer.xaml
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/controls/MiniTimer.xaml
@@ -4,6 +4,7 @@
              xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" 
              xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
              xmlns:toggl="clr-namespace:TogglDesktop"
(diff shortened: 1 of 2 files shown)


NghiaTranUIT:
On the other hand, the behavior is different for the following inputs:
- Enter `10` -> It starts at 10 mins
- Enter `10.1` -> It starts at 10 hours and 6min => I thought that it should be 10 min and 6 seconds

I believe that it's how our library is currently parsing the text, but it seems unintuitive 🤔

skel35:
That's the exact parsing that we've previously had, so I do not plan to change it in the scope of this task. So yeah, could be something for the future improvements.
I'll also mention that limiting the input only for digits, colon, full stop, and commas is too restrictive - the user can enter "5 mins" or "1h" and our app would parse it properly, so we don't want to break this functionality.

NghiaTranUIT:
That makes sense. I will open a ticket to address how we parse the duration string 👍

PR #4066, opened 2020-05-13
Use /me/flags to determine which syncing strategy to use

MartinBriza:
### 📒 Description
For the upcoming sync changes, we need to be able to switch between both protocols if things go haywire. This change makes the desktop app consume the flag from the API and then select which syncing strategy to use. If the server doesn't respond 200 or the flag is missing, we fallback to the legacy sync protocol.

You can now also use the following C++ compiler defines to force the syncing strategy of your choice on compile time:
```
TOGGL_SYNC_FORCE_BATCHED
TOGGL_SYNC_FORCE_LEGACY
```

### 🕶️ Types of changes

### 🤯 List of changes
- /api/v9/me/flags is now consumed on syncer bootup

### 🔎 Review hints
Everything should work as usual, except the syncer will start syncing after 1+ seconds later (depending on how long does the server take to respond).

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -82,7 +82,7 @@ Context::Context(const std::string &app_name, const std::string &app_version)
 , quit_(false)
 , ui_updater_(this, &Context::uiUpdaterActivity)
 , reminder_(this, &Context::reminderActivity)
-, syncer_(this, &Context::syncerActivity)
+, syncer_(this, &Context::syncerActivityWrapper)
 , update_path_("")
 , overlay_visible_(false)
 , last_message_id_("") {
@@ -4806,7 +4806,20 @@ void Context::reminderActivity() {
     }
 }
 
-void Context::syncerActivity() {
+void Context::syncerActivityWrapper() {
+    enum {
+        STARTUP,
+        LEGACY,
+        BATCHED
+    } state
+#if defined(TOGGL_SYNC_FORCE_BATCHED)
+    { BATCHED };
+#elif defined(TOGGL_SYNC_FORCE_LEGACY)
+    { LEGACY };
+#else
+    { STARTUP };
+#endif
+
     while (true) {
         // Sleep in increments for faster shutdown.
         for (int i = 0; i < 4; i++) {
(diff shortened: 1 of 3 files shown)


src/context.cc
```diff
@@ -4816,63 +4829,117 @@ void Context::syncerActivity() {
...
+                        // there was a typo in the initial set of flags, use both variants to be sure
+                        if (root[kSyncStrategyLegacy1].isBool() && root[kSyncStrategyLegacy1].asBool())
+                            state = BATCHED;
+                        else if (root[kSyncStrategyLegacy2].isBool() && root[kSyncStrategyLegacy2].asBool())
+                            state = BATCHED;
+                        else
+                            state = LEGACY;
+                        logger.log("Syncer - Syncing protocol was selected: ", (state == BATCHED ? "BATCHED" : "LEGACY"));
+                    }
                 }
```
NghiaTranUIT, line 4872: I suggest having an `else` statement here to rollback to `LEGACY` mode if the endpoint gets errors.

It would be:
```C++
else {
    state = LEGACY;
}
```
Thus, in the worst scenario that `/api/v9/me/flags` is down, the app still does Legacy Sync properly

MartinBriza: If the endpoint is down, the whole API is very likely to be down as well, resulting in not getting any data from/to the API anyway.

MartinBriza:
> Smoke testing and the sync func remains as a Legacy Mode. However, let check my comment to make sure that the sync always work

Yes, you'd have to have that flag enabled in backend for your account.

src/context.cc
MartinBriza: That also accounts for cases like when you don't have internet connection at all, too.

NghiaTranUIT: That makes sense 👍

NghiaTranUIT:
How can I enable my flag in BE? 🤔 Can you enable for me at `zevisuna@pachilly.com` (Testing email) onthe  Staging?

MartinBriza:
I don't know to be honest. Maybe it's possible to do that through Backoffice.

NghiaTranUIT:
So, if it's not ready for going to the production, I would like to approve it but not merge to the master for the safety 👍

skel35:
@MartinBriza is this PR production-ready by itself?
It looks to me that the users with this PR but without the next Sync-related PRs can end up with a `BATCHED` sync strategy that is not implemented here, thus missing some functionality. If that's the case, then this PR should be merged into some other branch than `master` and wait for other PRs implementing the batched sync functionality.

MartinBriza:
Now with `BATCHED` this will use legacy sync anyway, since there's no alternative implementation. Ready to re-review (or just merge straight away, there's no real difference to the previous version except for the fallback).

PR #4071, opened 2020-05-14
Basic push against the sync server

MartinBriza:
### 📒 Description
This PR implements the simplest subset of pushing to the sync server, currently located on a temporary host.

It also requires C++17 support (which may need to be enabled on platforms other than Linux, let's wait if GitHub Actions builds fail).

SaveToJSON methods now take an argument to generate a v9 JSON (some properties have different names and some have to be missing completely instead of sending `null`).

### 🕶️ Types of changes

### 🤯 List of changes
- Sync initiative: basic push

### 👫 Relationships
Requires #4066 merged first.

### 🔎 Review hints
For testing, an up-to-date version of the sync server has to be installed on localhost. See https://github.com/toggl/sync-api . Also, it's a bit easier to test with using the `TOGGL_SYNC_FORCE_BATCHED` compile-time flag.
The URL in urls.cc is subject to change in the future.

Since per-property updates (see docs) were introduced while this PR was in progress, it works basically just for creating new entities. Updates will get fixed in another PR.

This branch of code will be disabled until the beta flag is flipped, so it should be safe to merge into master.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.2)
 project(TogglDesktop)
 
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 17)
 
 # Set up automatic resource generation to make Qt development easier
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
--- a/src/const.h
+++ b/src/const.h
@@ -126,4 +126,8 @@
 #define kGoogleAccessToken "google_access_token"
 #define kAppleAccessToken "apple_token"
 
+// there was a typo in the initial set of flags, use both variants
+#define kSyncStrategyLegacy1 "dekstop_sync_client"
+#define kSyncStrategyLegacy2 "desktop_sync_client"
+
 #endif  // SRC_CONST_H_
--- a/src/context.cc
+++ b/src/context.cc
@@ -82,7 +82,7 @@ Context::Context(const std::string &app_name, const std::string &app_version)
 , quit_(false)
 , ui_updater_(this, &Context::uiUpdaterActivity)
 , reminder_(this, &Context::reminderActivity)
-, syncer_(this, &Context::syncerActivity)
+, syncer_(this, &Context::syncerActivityWrapper)
 , update_path_("")
 , overlay_visible_(false)
 , last_message_id_("") {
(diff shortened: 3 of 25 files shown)


skel35:
>This branch of code will be disabled until the beta flag is flipped, so it should be safe to merge into master.
How do we flip the beta flags? Is this something rolled out for an x% of users?
As some users are not updating for months, the `master` branch at any point may end up in some user's hands when the beta flag is flipped (at least until we have a solid "teapotting" strategy).

MartinBriza:
It's not rolled out for anyone. You can enable it for yourself in the Backoffice and the plan is to eventually enable it for a portion of users, possibly for everyone when things go fine (but even then only the people with the app that actually consumes the flag will get the new syncing strategy).

In Desktop, we can always utilize the compile time definition flag to force our stable releases to always use the legacy sync before we deem the batched implementation stable enough. 

As for the macOS build failure, I'm looking into that now. It's caused by the fact we don't have C++17 enabled on macOS yet.

skel35:
@MartinBriza Got it, thanks!
To me, it looks like we _should_ turn the compile-time definition flags in this PR to use the legacy sync strategy as `master` branch at any point can become a stable release.

MartinBriza:
Maybe we could consider doing rebuilds for stable versions as well. If we just turn it off in master, we'll never get any testers.

skel35:
@MartinBriza I was probably thinking about the #4066 one where there was a TODO comment in place of batched sync implementation.
If you feel this implementation can already go to some users then we don't need to define that compilation flag.

>Maybe we could consider doing rebuilds for stable versions as well. If we just turn it off in master, we'll never get any testers.

True, but doing rebuilds for stable versions requires altering our release procedure, so if you'd like to do it please write down a more specific proposal on how we should do it.
Another option I see is to hide the new strategy under smth like `if (channel != "stable")` if we don't want the `stable` channel users to see it.

MartinBriza:
> @MartinBriza I was probably thinking about the #4066 one where there was a TODO comment in place of batched sync implementation.

I suspect this PR and 4066 will go in in master in close succession so I don't consider it an issue. 

> If you feel this implementation can already go to some users then we don't need to define that compilation flag.
> 
> > Maybe we could consider doing rebuilds for stable versions as well. If we just turn it off in master, we'll never get any testers.
> 
> True, but doing rebuilds for stable versions requires altering our release procedure, so if you'd like to do it please write down a more specific proposal on how we should do it.
> Another option I see is to hide the new strategy under smth like `if (channel != "stable")` if we don't want the `stable` channel users to see it.

The proposal: branch the code at the tag that was previously beta, commit a `#define` disabling batched, tag this commit with the highest version number at that point + 1, let GitHub build it, use it as stable. This obviously breaks clean dev -> beta -> stable succession but it would be a temporary measure for about a month.

MartinBriza:
I'm closing this PR for now. I may reopen it later but I'm going to change the order of the PRs.

PR #4078, opened 2020-05-17
Implement Undo functionality

### LINKED ISSUE #4077: Support Undo deletion of time entries
### 💻 Environment

Platform: Windows

### 📒 Description

It would be nice if the Windows App has the Undo functionality so the user can restore mistakenly deleted time entries, same as what macOS app doing.

Abdulbaqi-Alshareef:
### 🕶️ Types of changes

New feature** (non-breaking change which adds functionality) 

### 🤯 List of changes

-Support Undo deletion.
-Avoid interrupting the running time entry so you can undo the deletion while having time entry running.

Undo

### 👫 Relationships

Closes #4077

### 🔎 Review hints
-Delete one or more time entries.
-Right-click on TimeEntry List (same as you do when removing time entry) and choose "Undo delete time entry"
-This PR only supports Undo not Redo(I can add it if requested).
-Undo only supports the deletion of time entries. 
-You can use undo only using the mouse, ctrl + z reserved for another functionality.


Abdulbaqi-Alshareef:
-You are right, at first, I wrote TimeEntrySnapshot class at first steps implementing the function but now no need for it.
-I added Ctrl+Z shortcut to undo deletion, and it only works if the time entry list got focused and that's difficult for the user due to the timer's behavior to retain focus.

Issue #4021, opened 2020-04-24
Preferences window doesn't fit the screen vertically (1080px height, 150% scaling)


skel35:
### 💻 Environment
Platform: Windows
Toggl Version: 7.5.117

### 🐞 Actual behavior
Preferences window doesn't fit the screen vertically 

### 💯 Expected behavior
Preferences window should fit the screen vertically

### 🔨 Steps to reproduce
1080px screen height, 150% scaling.

### 📦 Additional info
[Case reference](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/unassigned/conversations/26739780774)

skel35:
Fixed by #4079

PR #4079, opened 2020-05-17
Fix the max size of Preferences window

Abdulbaqi-Alshareef:
### 📒 Description

The Height of Preferences window set to high constant number (760) so it may not fit on the low display resolution, and because the window is not resizable the user can do nothing to access the bottom of the window.

### 🕶️ Types of changes

### 🤯 List of changes

Set the Maximum height of the window to be the height of the working area of the screen.
### 🔎 Review hints

-Set the display resolution to low resolution.
-open Preferences window, the window should fit the screen.


skel35:
should fix #4021

PR #4086, opened 2020-05-20
Account locking error messages

### LINKED ISSUE #4074: Account locking (Password Security project)
### 💻 Environment
Platform: Lib

### 📒 Description
The backend will start sending the header `X-Remaining-Login-Attempts` together with the failed login response.
The desktop app should display an error message when the login has failed depending on the value of the aforementioned header.

If `X-Remaining-Login-Attempts` value is `1` the desktop app should display an error message: `Incorrect email or password. One more try before account gets locked for 5 minutes.`.
If `X-Remaining-Login-Attempts` value is `0` the desktop app should display an error message: `Incorrect email or password. Account is locked for 5 minutes. Account owner has been notified.`.
Otherwise, the error message should be `Incorrect email or password. Please try again`.

skel35:
### 📒 Description
Password Security project. Account locking part.

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Show proper error message when account is locked or about to get locked
- [x] Make sure Login view on Windows shows only one type of error at a time

### 👫 Relationships
Closes #4074 

### 🔎 Review hints

Method 1
- Ask someone from BE in #project-security-features to deploy account locking to Staging.

Method 2
- Manipulate the HTTP response via some debugging proxy tool so that the response contains the necessary header

Method 3
- Manipulate the code to force the response to contain the necessary header

### DIFF
--- a/src/const.h
+++ b/src/const.h
@@ -95,6 +95,9 @@
 #define kMissingWS "You no longer have access to your last workspace"
 #define kOutOfDatePleaseUpgrade "Your version of Toggl Desktop is out of date, please upgrade!"
 #define kThisEntryCantBeSavedPleaseAdd "This entry can't be saved - please add"
+#define kOneLoginAttemptLeft "Incorrect email or password. One more try before account gets locked for 5 minutes."
+#define kAccountIsLocked "Incorrect email or password. Account is locked for 5 minutes. Account owner has been notified."
+#define kIncorrectEmailOrPassword "Incorrect email or password. Please try again."
 
 #define kModelAutotrackerRule "autotracker_rule"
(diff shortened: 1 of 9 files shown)


skel35:
@NghiaTranUIT I went through the Notion docs of this feature again and the conclusion was that the account doesn't get locked if the user is still logged in ([Notion](https://www.notion.so/toggl/31-03-2020-f2e3f6359f794c0f832deeb223b85633#08621caef59a4918b2fdff2165ba3ec1)). So this error shouldn't occur in the case you described.
The error is a bit specific to the login flow, but I think it's fine to keep the simpler implementation for now.

PR #4128, opened 2020-06-02
Add properties to data models 2.0

MartinBriza:
### 📒 Description
This PR slightly (especially compared to the previous version of the same PR) changes how data members are handled in the data model classes.

Adds a `Property<T>` class that will track the changes to the separate properties (for syncing to the server but probably even the database in some future stage).

All property getters are now gone. There is very little change to the current code because `const T& Property<T>::operator()` is implemented.
Setters are still mostly the same but they call `Property<T>::Set(T)` internally. 

WARNING: Please note that calling for example `client->GUID.Set("whatever");` won't change the `Dirty` flag and this change won't be stored to the database. The right way is to call `client->SetGUID("whatever");`.
Again, this is something that should be changed in a future version of properties - doing the setters in this PR would be a very disruptive change and we can live without it now.

One extra thing I did here was to sanitize the setter arguments (it doesn't make sense to pass `const` integral types as arguments) and reorganized the structure of the class headers slightly so they are more readable.

### 🕶️ Types of changes

### 👫 Relationships
Depends on #4112.

### 🔎 Review hints
There should be no change to behavior of the app. 
This will be a bit annoying to review because it'll be mostly looking if I made a typo somewhere.

### DIFF
--- a/src/model/time_entry.cc
+++ b/src/model/time_entry.cc
@@ -29,23 +29,23 @@
 namespace toggl {
 
 bool TimeEntry::ResolveError(const error &err) {
-    if (durationTooLarge(err) && Stop() && Start()) {
+    if (durationTooLarge(err) && StopTime() && StartTime()) {
         Poco::Int64 seconds =
-            (std::min)(Stop() - Start(),
+            (std::min)(StopTime() - StartTime(),
                        Poco::Int64(kMaxTimeEntryDurationSeconds));
         SetDurationInSeconds(seconds);
         return true;
     }
-    if (startTimeWrongYear(err) && Stop() && Start()) {
+    if (startTimeWrongYear(err) && StopTime() && StartTime()) {
         Poco::Int64 seconds =
-            (std::min)(Stop() - Start(),
+            (std::min)(StopTime() - StartTime(),
                        Poco::Int64(kMaxTimeEntryDurationSeconds));
         SetDurationInSeconds(seconds);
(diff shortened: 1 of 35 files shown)


src/model/time_entry.cc
```diff
@@ -434,16 +434,22 @@ void TimeEntry::LoadFromJSON(Json::Value data) {
...
     if (data.isMember("pid")) {
         SetPID(data["pid"].asUInt64());
+    } else if (data.isMember("project_id")) {
+        SetPID(data["project_id"].asUInt64());
     } else {
         SetPID(0);
     }
     if (data.isMember("tid")) {
         SetTID(data["tid"].asUInt64());
+    } else if (data.isMember("project_id")) {
```
skel35, line 451: there should be `task_id` here.

lucky to have found this, didn't notice while reading the code, accidentally discovered during testing

skel35: I suppose we can also write tests for this in `app_test.cc`

skel35: oh, I just realized this particular line is already merged to master in the previous PR

skel35: pushed this fix to `master`

MartinBriza:
I rebased the branch against master but then I noticed you have already done the same so I changed the branch to follow your branch. I also checked once again if there are any typos but I didn't find any. 
Ready for re-review I guess.

PR #4164, opened 2020-06-10
Fixing tags dropdown behavior in edit view

### LINKED ISSUE #4004: When creating a new Tag, dropdown only clears by pressing TAB or clicking on the main timer screen
### 💻 Environment

Platform:  Windows
OS Version:  Windows 10
Toggl Version:  7.5.92

### 🐞 Actual behavior

When you add a tag to a time entry you can't un-focus it by clicking the element behind the drop down. You either need to click the main timer screen or press the "Tab" key to clear the drop down.

### 💯 Expected behavior

Dropdown should clear when clicking the element behind it as well.

### 📦 Additional info

Screenshots in the [case](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/2759100/conversations/26576827705) 
Relevant slack [thread](https://toggl.slack.com/archives/C029WCX0T/p1586934528273600?thread_ts=1586887838.269700&cid=C029WCX0T)

atsvirchkova:
### 📒 Description
Changes the behavior of tags dropdown so that it closes when user clicks outside it.

### 🕶️ Types of changes

### 👫 Relationships

 Closes #4004 

### 🔎 Review hints
Now the behavior for prject dropdown and for tags dropdown is the same.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/controls/TagList.xaml
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/controls/TagList.xaml
@@ -25,7 +25,7 @@
                                        ConfirmCompletion="autoComplete_OnConfirmCompletion"
                                        ConfirmWithoutCompletion="autoComplete_OnConfirmWithoutCompletion"
                                        KeepOpenWhenSelecting="True"
-                                       StaysOpen="True"
+                                       StaysOpen="False"
                                        IsOpenChanged="AutoComplete_OnIsOpenChanged"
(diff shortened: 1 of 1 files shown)


atsvirchkova:
The dropdown has different behavior depending on how you unwrap it. If you do this using the down button, then yes, it stays open until you again press the up button. But if it upwraps when you start typing, then it closes when you click on form (before this fix, it didn't). Looks like this kind of behavior is by design from popup. So, does this need to be changed too?

PR #4168, opened 2020-06-11
Fix project autocomplete not displaying the right text (linux)

### LINKED ISSUE #3714: Project selection does not work in 7.5.4 on Linux
I have cleared cache.

To reproduce...start a timer (or edit an existing entry) and try to select a project. The click does not register. It does not work with the keyboard either. Apologies if this was already fixed later on. I will try to update the AUR if so.

MartinBriza:
### 📒 Description
This fixes the selection of projects in the autocomplete in the time entry editor. There was a bug when selecting the project, the text was not displayed right (and sometimes not stored properly as well).

### 🕶️ Types of changes

### 👫 Relationships
Closes #3714

### 🔎 Review hints
Check if project selection works right now.


MartinBriza:
Yeah make that a different bug. This make it works at least when you press Enter while inside the field.

PR #4170, opened 2020-06-12
Refactoring Timer

### LINKED ISSUE #3730: Extract Timer viewmodel

atsvirchkova:
### 📒 Description
Timer.xaml.cs and MiniTimer.xaml.cs contain almost the same code. This removes most of the duplication and places the common code into the view model.

### 🕶️ Types of changes

### 🤯 List of changes
Timer
MiniTimer

### 👫 Relationships

Closes #3730 

### 🔎 Review hints
What is left:
1. Some duplication in the code behind which refers directly to other wpf controls. This can't be placed into view model. Probably we can do a common xaml control for timer and mini Timer and just style it differently.
2. Some strange logic with auto complete popup. I noticed this only a few hours ago. It should not open when text in the description box is changed programmacaly. This logic was lost because of binding.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimerViewModel.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimerViewModel.cs
@@ -15,7 +15,7 @@ namespace TogglDesktop.ui.ViewModels
         private readonly DispatcherTimer secondsTimer = new DispatcherTimer();
         private Toggl.TogglTimeEntryView runningTimeEntry;
         private bool acceptNextUpdate;
-        private bool isMiniTimer;
+        private readonly bool isMiniTimer;
         private Toggl.TogglAutocompleteView completedProject;
 
         public TimerViewModel(bool isMiniTimer)
@@ -24,7 +24,6 @@ namespace TogglDesktop.ui.ViewModels
             StartStopCommand = ReactiveCommand.Create(startStop);
             CancelProjectSelectionCommand = ReactiveCommand.Create(clearSelectedProject);
             ManualAddButtonCommand = ReactiveCommand.Create(AddNewTimeEntry);
-            TryOpenEditView = ReactiveCommand.Create<string>(s => tryOpenEditViewIfRunning(s));
 
             setupSecondsTimer();
 
@@ -61,8 +60,6 @@ namespace TogglDesktop.ui.ViewModels
 
         public ReactiveCommand<Unit, Unit> ManualAddButtonCommand { get; }
 
-        public ReactiveCommand<string, Unit> TryOpenEditView { get; }
-
         private void setupSecondsTimer()
         {
             secondsTimer.Interval = TimeSpan.FromSeconds(1);
@@ -79,7 +76,6 @@ namespace TogglDesktop.ui.ViewModels
         {
             using (Performance.Measure("timer responding to OnRunningTimerState"))
             {
-                
                 SetRunningTimeEntry(te);
                 secondsTimer.IsEnabled = true;
             }
@@ -147,6 +143,7 @@ namespace TogglDesktop.ui.ViewModels
 
         private void start()
         {
+            var durationString = DurationText;
             using (Performance.Measure("starting time entry from timer"))
             {
                 var guid = Toggl.Start(
@@ -158,9 +155,9 @@ namespace TogglDesktop.ui.ViewModels
                     completedProject.Tags,
                     isMiniTimer
                     );
-                if (isMiniTimer && !string.IsNullOrEmpty(guid) && !string.IsNullOrEmpty(DurationText))
+                if (isMiniTimer && !string.IsNullOrEmpty(guid) && !string.IsNullOrEmpty(durationString))
                 {
-                    Toggl.SetTimeEntryDuration(guid, DurationText);
+                    Toggl.SetTimeEntryDuration(guid, durationString);
                 }
                 if (completedProject.Billable)
                 {
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimerViewModel.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimerViewModel.cs
@@ -1,0 +1,209 @@
+﻿using ReactiveUI;
+using ReactiveUI.Fody.Helpers;
+using System;
+using System.Reactive;
+using System.Windows;
+using System.Windows.Threading;
+using TogglDesktop.AutoCompletion.Items;
+using TogglDesktop.Diagnostics;
+using TogglDesktop.ViewModels;
+
+namespace TogglDesktop.ui.ViewModels
+{
+    public class TimerViewModel : ReactiveObject
+    {
+        private readonly DispatcherTimer secondsTimer = new DispatcherTimer();
+        private Toggl.TogglTimeEntryView runningTimeEntry;
+        private bool acceptNextUpdate;
+        private bool isMiniTimer;
+        private Toggl.TogglAutocompleteView completedProject;
+
+        public TimerViewModel(bool isMiniTimer)
+        {
+            this.isMiniTimer = isMiniTimer;
+            StartStopCommand = ReactiveCommand.Create(() => startStop());
+            CancelProjectSelectionCommand = ReactiveCommand.Create(() => clearSelectedProject());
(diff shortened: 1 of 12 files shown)


src/ui/windows/TogglDesktop/TogglDesktop/ui/views/Timer.xaml.cs
```diff
@@ -27,64 +26,16 @@ public partial class Timer
         public Timer()
         {
             this.InitializeComponent();
-
-            this.setupSecondsTimer();
+            ViewModel = new TimerViewModel(false);
+            ViewModel.WhenValueChanged(x => x.IsRunning).Subscribe(_ => StartStopClick?.Invoke(this, EventArgs.Empty));
```
skel35, line 30: `IsRunning` is changed != start/stop was clicked. `IsRunning` can be changed from the outside, so this is a logical error.

src/ui/windows/TogglDesktop/TogglDesktop/ui/views/Timer.xaml.cs
```diff
@@ -27,64 +26,16 @@ public partial class Timer
         public Timer()
         {
             this.InitializeComponent();
-
-            this.setupSecondsTimer();
+            ViewModel = new TimerViewModel(false);
+            ViewModel.WhenValueChanged(x => x.IsRunning).Subscribe(_ => StartStopClick?.Invoke(this, EventArgs.Empty));
+            ViewModel.WhenValueChanged(x => x.DurationText).Subscribe(e => RunningTimeEntrySecondPulse?.Invoke(this, e));
```
skel35, line 31: It's easy to change the subscribers of `RunningTimeEntrySecondPulse` to subscribe to the WhenValueChanged observable instead. This way there will be no need for the `RunningTimeEntrySecondPulse` event. Same applies to the `StartStopClick` event.

src/ui/windows/TogglDesktop/TogglDesktop/ui/controls/MiniTimer.xaml.cs
```diff
@@ -184,114 +104,6 @@ protected override void OnGotKeyboardFocus(KeyboardFocusChangedEventArgs e)
...
-                    completedProject.TaskID,
-                    completedProject.ProjectID,
-                    "",
-                    completedProject.Tags,
-                    IsMiniTimer
-                );
-
-                if (!string.IsNullOrEmpty(guid) && !string.IsNullOrEmpty(durationString))
-                {
-                    Toggl.SetTimeEntryDuration(guid, durationString);
```
skel35, line 244: the functionality for setting starting duration from the mini timer has disappeared after this refactoring

src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimerViewModel.cs
```diff
@@ -0,0 +1,209 @@
...
+        private bool isMiniTimer;
+        private Toggl.TogglAutocompleteView completedProject;
+
+        public TimerViewModel(bool isMiniTimer)
+        {
+            this.isMiniTimer = isMiniTimer;
+            StartStopCommand = ReactiveCommand.Create(() => startStop());
+            CancelProjectSelectionCommand = ReactiveCommand.Create(() => clearSelectedProject());
+            ManualAddButtonCommand = ReactiveCommand.Create(() => AddNewTimeEntry());
+            DescriptionAutoCompleteConfirmCommand = ReactiveCommand.Create<RoutedEventArgs>(e => {
```
skel35, line 27: this command is not used anywhere

src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimerViewModel.cs
```diff
@@ -0,0 +1,209 @@
...
+
+        #endregion
+
+        public void clearSelectedProject()
+        {
+            ProjectLabelViewModel = null;
+            completedProject = default;
+        }
+
+        public void DescriptionAutoCompleteConfirm(object e)
```
skel35, line 189: the argument can be strongly typed here

atsvirchkova: Thank you for such a nice and detailed feedback! I agree with your comments, and I see, there is a couple of things I lost out of my view (about ```StartStopClick```, and I really didn't look much in the MiniTimer's code).

src/ui/windows/TogglDesktop/TogglDesktop/ui/controls/MiniTimer.xaml.cs
atsvirchkova: Is there any sense, that this is present only in MiniTimer, but not present in Timer?

skel35: It's temporarily missing from the main Timer's design. It will probably be added to the Timer later.

src/ui/windows/TogglDesktop/TogglDesktop/ui/views/Timer.xaml.cs
skel35: @atsvirchkova this one is still not addressed

atsvirchkova: Well, yes. The only difference between the `StartStopClicked` and `IsRunning` changing is that `IsRunning` can be set using keyboard. But the only thing that was done on `StartStopClicked` is calling `closeEditPopup()`. Is it intended to work differently in these two cases (I mean the keyboard case and the mouse case)?

skel35: Not only the keyboard. Any update coming from the library can cause `IsRunning` to change. It seems like `TimerViewModel` doesn't apply those updates, which is causing the bug https://github.com/toggl-open-source/toggldesktop/pull/4170#pullrequestreview-435640135

skel35: it seems like it's fine to close the edit popup every time when `IsRunning` is changed though

atsvirchkova:
Fixed the bug and returned back the `StartStopClick` event.

src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimerViewModel.cs
```diff
@@ -0,0 +1,216 @@
...
+        private bool isMiniTimer;
+        private Toggl.TogglAutocompleteView completedProject;
+
+        public TimerViewModel(bool isMiniTimer)
+        {
+            this.isMiniTimer = isMiniTimer;
+            StartStopCommand = ReactiveCommand.Create(startStop);
+            CancelProjectSelectionCommand = ReactiveCommand.Create(clearSelectedProject);
+            ManualAddButtonCommand = ReactiveCommand.Create(AddNewTimeEntry);
+            TryOpenEditView = ReactiveCommand.Create<string>(s => tryOpenEditViewIfRunning(s));
```
skel35, line 27: unused command

src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimerViewModel.cs
```diff
@@ -0,0 +1,216 @@
...
+using TogglDesktop.ViewModels;
+
+namespace TogglDesktop.ui.ViewModels
+{
+    public class TimerViewModel : ReactiveObject
+    {
+        private readonly DispatcherTimer secondsTimer = new DispatcherTimer();
+        private Toggl.TogglTimeEntryView runningTimeEntry;
+        private bool acceptNextUpdate;
+        private bool isMiniTimer;
```
skel35, line 18: why is this not read-only?

src/ui/windows/TogglDesktop/TogglDesktop/ui/controls/MiniTimer.xaml.cs
skel35: This functionality is still not working

atsvirchkova:
Fixed the issue with starting duration in Mini timer and also the small ones.

skel35:
closing in favor of #4204

PR #3979, opened 2020-04-08
[WIP] Bug/3407

jiri:
### 🕶️ Types of changes

- **New feature** (non-breaking change which adds functionality)

### 👫 Relationships

Closes #3407

PR #4189, opened 2020-06-18
Allow release builds to connect to Staging

### LINKED ISSUE #3407: Allow connecting to staging db using a command-line switch
### 💻 Environment
Platform: macOS/Windows/Linux

### 📒 Description
Allow connecting to staging DB using a command-line switch (e.g. `--staging=true`).

### ⭐️ Why do you want this?
- To make it possible to test the app in a staging environment.
- To make it possible for non-developers to run the app in a staging environment

skel35:
### 📒 Description
Allow release builds to connect to the Staging environment. This is #3979 + Windows support.

### 🕶️ Types of changes

### 🤯 List of changes
- [x] New API method `toggl_set_staging_override`
- [x] Support `--staging` argument on Windows and Linux, and `staging` argument on macOS (does this need to be fixed?)

### 👫 Relationships
Closes #3407

### 🔎 Review hints
Build the app in Release mode and run with `--staging` argument.
The app should connect to the Staging environment instead of Production.
Test on macOS, Windows, and Linux.

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -94,10 +94,6 @@ Context::Context(const std::string &app_name, const std::string &app_version)
         Poco::Net::HTTPSStreamFactory::registerFactory();
     }
 
-#ifndef TOGGL_PRODUCTION_BUILD
-    urls::SetUseStagingAsBackend(true);
-#endif
-
     Poco::ErrorHandler::set(&error_handler_);
     Poco::Net::initializeSSL();
 
--- a/src/context.cc
+++ b/src/context.cc
@@ -5649,7 +5649,7 @@ error Context::pushProjects(
 }
 
 error Context::updateProjectClients(const std::vector<Client *> &clients,
-                                     const std::vector<Project *> &projects) {
+                                    const std::vector<Project *> &projects) {
     for (auto it = projects.cbegin(); it != projects.cend(); ++it) {
         if (!(*it)->CID() && !(*it)->ClientGUID().empty()) {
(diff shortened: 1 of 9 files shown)


AndrewVebster:
Tested on macOS. Switches to staging if an argument contains "staging" word, so basically any of those will work: `--staging`, `staging`, `staging=false`, `staging=true`. 
Is this an expected behavior?

Also tested this use case:
1. Run on a prod. Signup with a new account.
2. Reopen the app with `staging` env.

Result: because on staging the account is missing app shows an error window. Also, maybe there is a bug with this window, but buttons are not working, so I could not login again.

Considerations: Maybe we should proactively log out the user if he changes the environment. What do you think?

skel35:
@AndrewVebster 
1. It makes sense to accept only "--staging", fixed it.

2. >Also, maybe there is a bug with this window, but buttons are not working, so I could not login again.

Is it the "You've lost access to this workspace" window? Please create a separate issue for the bug you've discovered.

3. >Considerations: Maybe we should proactively log out the user if he changes the environment. What do you think?

I think it's convenient for us, developers, and other people who might test on Staging, like the Support team, to stay logged in when switching between Staging and Production. Why do you think proactively logging out would be useful? other than the scenario you mentioned, where it should take one click to log out if the view wasn't buggy.

AndrewVebster:
2. Yes, I will create a bug for this.
3. Okay, I think, you're right.

PR #4201, opened 2020-06-23
Sync: Three-way merge Push

MartinBriza:
### 📒 Description
After rewriting this PR a few times, I settled upon a solution that will only work for time entries (Desktop doesn't support editing clients or projects anyway) and in as simple way as possible.
This is a base implementation that I will build upon in a (hopefully final) followup. For now, it supports only pushing updates and will probably barf (and by that I mean overwrite server changes) if you try to edit something in two clients at the same time.

Notable changes in the library:
 * Some `TimeEntry` setters now take the `userModified` boolean flag to denote if the change is coming from the user or from backend.
 * Template implementations in `User` were moved to the top of the file where they belong
 * The `Property<T>` class now provides the `value_type` typedef for usage with `std::decltype`

### 🕶️ Types of changes

### 🔎 Review hints
When used with the sync server, pushing, pulling and updating of entities should work as expected. You can watch the sync server JSON traffic in `stderr`.

### DIFF
--- a/src/model/client.cc
+++ b/src/model/client.cc
@@ -85,7 +85,7 @@ Json::Value Client::SyncMetadata() const {
 Json::Value Client::SyncPayload() const {
     Json::Value result;
     if (NeedsPOST()) {
-        result["id"] = Json::Int64(ID());
+        result["id"] = Json::Int64(-LocalID());
         result["wid"] = Json::Int64(WID());
     }
     if (NeedsPOST() || NeedsPUT()) {
--- a/src/context.cc
+++ b/src/context.cc
@@ -3239,6 +3239,10 @@ error Context::updateTimeEntryProject(
         p = user_->related.ProjectByGUID(project_guid);
     }
 
+    if (isUsingSyncServer() && p && p->WID() != te->WID()) {
+        return displayError("This version of Toggl Desktop does not support changing time entry workspaces.");
+    }
+
     if (p && !canChangeProjectTo(te, p)) {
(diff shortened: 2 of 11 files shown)


src/model/client.cc
```diff
@@ -85,7 +85,7 @@ Json::Value Client::SyncMetadata() const {
 Json::Value Client::SyncPayload() const {
     Json::Value result;
     if (NeedsPOST()) {
-        result["id"] = Json::Int64(ID());
+        result["id"] = Json::Int64(-LocalID());
```
skel35, line 88: is "client_assigned_id" field still needed if "id" field has the same value? can you show the Sync spec which refers to this? [this one](https://github.com/toggl/toggl-api-clients/tree/master/Sync%20Specification) seems to not go into details of how JSONs should look

MartinBriza: I'll answer this on Slack

MartinBriza:
The running time entry problem is a sync server bug that is being worked on.

As for the sending invalid data, I'll look into it.

MartinBriza:
The invalid data error is resolved in 9ee5e4e

MartinBriza:
Hmm, I suspect changing workspaces is not supported by the Sync server. We'll need to delete the time entry and create a new one for this.

MartinBriza:
I added an error message that this is not possible with sync server at this moment so we can move forward with the review process. 
I'll work on enabling this feature again in a followup PR.

PR #4204, opened 2020-06-25
Refactor timer

### LINKED ISSUE #3730: Extract Timer viewmodel

atsvirchkova:
📒 Description
Timer.xaml.cs and MiniTimer.xaml.cs contain almost the same code. This removes most of the duplication and places the common code into the view model.

🕶️ Types of changes

🤯 List of changes
Timer
MiniTimer

👫 Relationships
Closes #3730

🔎 Review hints
What is left:

Some duplication in the code behind which refers directly to other wpf controls. This can't be placed into view model. Probably we can do a common xaml control for timer and mini Timer and just style it differently.
Some strange logic with auto complete popup. I noticed this only a few hours ago. It should not open when text in the description box is changed programmacaly. This logic was lost because of binding.


skel35:
@atsvirchkova Please resolve the conflicts to allow merging to master. I suggest rebasing on `master`, merge commit is fine as well.

PR #4228, opened 2020-07-07
Fix BaseModel database performance by not making it dirty everytime

### LINKED ISSUE #4223: Performance slowdown in windows desktop app
### 💻 Environment

Platform: Windows
OS Version: Windows 10.1
Toggl Version: 7.5.208

### 🐞 Actual behavior
A slowdown appeared after commit 2084403d4be0f163702095bd78aa0744298db7e5. The lib function Toggl.Start() now takes about 700-900 ms, while before it used to take about 150-200 ms. After some investigation of c++ code I found that this call
`error err = db()->SaveUser(user_, true, &changes);` takes a lot longer than before.

### 💯 Expected behavior
No lags in application work.

### 🔨 Steps to reproduce

1. Just click the Start button.

### 📦 Additional info
Lags appear not only on Start button, but also on any action that makes saving to database, e. g. setting a project on running entry, a tag and so on.

MartinBriza:
### 📒 Description
This PR addresses the performance hit some Windows users were experiencing - performance went straight downhill after applying my performance patch.
As Anna correctly noted, the `Dirty` property was set on each call to `Set<Something>` even when there was no change. This happened mostly for the `ID` and `GUID` properties because I made this error only in `BaseModel` (I wonder why, because in other models it's fine).
This PR changes all models to check the `Property<T>::Set` return value to determine if there was any change so we're consistent and it also reduces the lines of code slightly.

### 🕶️ Types of changes

### 🤯 List of changes
- Windows app performance is fixed

### 👫 Relationships
Closes #4223

### 🔎 Review hints
No changes to the functionality should be present. You can check the duration of the database storage times in the logs.

### DIFF
--- a/src/util/property.h
+++ b/src/util/property.h
@@ -39,19 +39,32 @@ public:
         previous_ = current_;
     }
     /* Setters */
-    void Set(const T& value, bool makeDirty = true) {
+    bool Set(const T& value, bool makeDirty = true) {
+        if (value == current_) {
+            previous_ = value;
+            return false;
+        }
         if (makeDirty)
             previous_ = std::move(current_);
         else
             previous_ = value;
         current_ = value;
+        return true;
     }
-    void Set(const T&& value, bool makeDirty = true) {
-        if (makeDirty)
-            previous_ = std::move(current_);
-        else
+    bool Set(const T&& value, bool makeDirty = true) {
+        if (value == current_) {
             previous_ = std::move(value);
-        current_ = value;
+            return false;
+        }
+        if (makeDirty) {
+            previous_ = std::move(current_);
+            current_ = std::move(value);
+        }
+        else {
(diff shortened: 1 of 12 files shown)


src/util/property.h
```diff
@@ -39,19 +39,32 @@ class Property {
         previous_ = current_;
     }
     /* Setters */
-    void Set(const T& value, bool makeDirty = true) {
+    bool Set(const T& value, bool makeDirty = true) {
+        if (value == current_) {
+            previous_ = value;
```
atsvirchkova, line 44: What is the need to set `previous_` here, why not just return false right away?

MartinBriza: There's no particular _need_ to do that but in case the past value was different, this sets them to be the same.

PR #4231, opened 2020-07-09
Sync: enable changing TE workspaces

### LINKED ISSUE #4221: Sync: Changing TE Workspaces is not possible.
### 💻 Environment

Platform: 
OS Version: 
Toggl Version: 

### 🔨 Steps to reproduce

1. 
2. 
3.

MartinBriza:
### 📒 Description
This PR enables changing workspaces for time entries again.

This is achieved by:
 * Splitting the time entry into two, essentially creating a new one, server-side (that means clearing the ID in the one we're keeping locally)
 * Allowing cloning the `TimeEntry` and `BaseModel` class (without copying the `ID`, `GUID` and `LocalID` fields).
 * Clearing the `id` field in the database when updating time entries (now it was just left alone when the TE was updated) 
 * And finally, sending the new mock time entry (containing an old `ID) to the server for deletion, along with a "new" time entry in another workspace for creation.

### 🕶️ Types of changes

### 👫 Relationships
Closes #4221 

### 🔎 Review hints
Try changing TE workspaces, while online and while offline. Everything should work the same with both sync and legacy synchronization protocols.

### DIFF
--- a/src/database/database.cc
+++ b/src/database/database.cc
@@ -2179,15 +2179,11 @@ error Database::saveModel(
                           useRef(model->LocalID()),
                           now;
             } else {
-                // For cases where we need to remove the ID from the TE (like when changing the workspace)
                 *session_ <<
                           "update time_entries set "
-                          "id = null "
-                          "where local_id = :local_id",
-                          useRef(model->LocalID()),
-                          now;
-                *session_ <<
-                          "update time_entries set "
+                          /* For cases where we need to remove the ID from the TE (like when changing the workspace) */
+                          "id = null, "
+                          /* Regular column updates */
                           "uid = :uid, description = :description, wid = :wid, "
                           "guid = :guid, pid = :pid, tid = :tid, "
                           "billable = :billable, "
--- a/src/database/database.cc
+++ b/src/database/database.cc
@@ -2179,6 +2179,13 @@ error Database::saveModel(
                           useRef(model->LocalID()),
                           now;
             } else {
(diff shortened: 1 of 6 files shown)


src/database/database.cc
```diff
@@ -2179,6 +2179,13 @@ error Database::saveModel(
                           useRef(model->LocalID()),
                           now;
             } else {
+                // For cases where we need to remove the ID from the TE (like when changing the workspace)
+                *session_ <<
+                          "update time_entries set "
```
skel35, line 2184: why is this a separate query and not included into the subsequent query which is also an `update` query with the same `where` condition?

MartinBriza: I wanted it to stand out more in case something is not clear in the future. Considering how the update looks below, it felt like it would be easy to confuse things.
It could be inside too of course, I can move it if you'd like it better that way.

skel35: This is certainly more readable and clear, but the reason I mentioned this is performance, 2 requests are likely slower than one. If we know there is no noticeable difference when doing this on `time_entries` table with 1000-2000 rows then this is fine by me.

MartinBriza: This is definitely more of a special case. This is executed only when you move a TE from one workspace to another from the desktop client. Cases when this happens for more than one TE at a time are very rare, and I'd say even a chance for this happening for 2 TEs at the same time is next to none. 
Anyway, I don't really have a strong opinion where to have this and for any potential future features, it's better to do it in one query. I'll change it.

PR #4236, opened 2020-07-13
Fix the linux tarball build

### LINKED ISSUE #4235: linux-tarball workflow is broken (hanging) in CI
### 💻 Environment

Platform: Linux
Toggl Version: 7.5.217

### 🐞 Actual behavior
`linux-tarball` workflow is hanging in CI (timeout of 6 hours is reached). E.g. https://github.com/toggl-open-source/toggldesktop/runs/858673738?check_suite_focus=true.

### 💯 Expected behavior
`linux-tarball` should succeed.

### 🔨 Steps to reproduce
Run the release flow (e.g. by committing smth to `master`).

MartinBriza:
### 📒 Description
This PR fixes how the tarball is composed. Now it will rely on using https://github.com/jurplel/install-qt-action to build the app for the tarball which is much more stable than our current solution

### 🕶️ Types of changes

### 👫 Relationships
Fixes #4235

### 🔎 Review hints
See https://github.com/toggl-open-source/toggldesktop/runs/865751511?check_suite_focus=true (the particular GA branch is enabled there).


.github/workflows/main.yml
```diff
@@ -142,16 +142,32 @@ jobs:
...
+        sudo apt update
+        sudo apt install g++-7 -y
+        sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-7 60 --slave /usr/bin/g++ g++ /usr/bin/g++-7
+        sudo update-alternatives --config gcc
+    - name: Cache Qt
+      id: cache-qt
+      uses: actions/cache@v1
+      with:
+        path: ../Qt
+        key: QtCache
```
skel35, line 158: The key doesn't depend on the Qt version, so when the new version of Qt is released, the cache of the old version would still be used. Could this be an issue?

MartinBriza: We would have to manually bump the Qt version to actually upgrade it. The action installs 5.12 series Qt by default and the company stopped releasing LTS z-versions. If we manually change the Qt version, we should probably change the cache name as well. But yeah, good point, I'll change the name of the cache so it's clear we should change it when changing Qt version.

PR #4239, opened 2020-07-14
Upload macOS app installers as artifacts

### LINKED ISSUE #4219: Create and upload app installers as artifacts in GitHub Actions branch build for macOs builds
### 💻 Environment

Platform:  macOS

### 📒 Description
We need to produce the macos installer after each pull request and upload it as an artifact to GitHub.

This would be very useful for testing purposes both within our team, as the tester would not need to build an app locally, and outside our team if we want someone else to try out the branch build.

skel35:
### 📒 Description
Upload macOS app installers as artifacts

### 🕶️ Types of changes

### 🤯 List of changes
 - [x] Compose dmg without notarization and upload it as an artifact on every non-release build

### 👫 Relationships
Closes #4219 

### 🔎 Review hints
Look into the artifacts from this PR's build. Take the .dmg, install the app, verify that it can be installed and it works.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -92,9 +92,14 @@ jobs:
         export "APPLE_APPID_PASSWORD=${{ secrets.MACOS_APPID_PASSWORD }}"
         bash ./dist/osx/build.sh notarize
     - name: Compose dmg
-      if: github.event_name == 'release'
       run: |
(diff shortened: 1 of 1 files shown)


AndrewVebster:
nice! Downloaded dmg, installed, all works

PR #4245, opened 2020-07-15
Sync: Check if WS is premium for each individual item

MartinBriza:
### 📒 Description
Just determining if we can work with billable flags by looking into the user's workspaces wasn't enough. This will check if the workspace is premium for each individual TE/Project and fix erroneous behavior.

### 🕶️ Types of changes

### 👫 Relationships
This was reported on Slack, probably not on GH yet.

### 🔎 Review hints
~~Andrii, just run the app and it should get resolved automatically.~~ Try creating multiple new projects in two different workspaces and assign them to time entries consecutively. For example:
 * Create TE A, B
 * In workspaces X and Y, create projects:
   * X1 for A, 
   * Y1 for A
   * X2 for B
   * Y2 for A (intentional).
 * Assign X1 to B
 * Assign X2 to A
 * Assign Y1 to B
 * Assign Y2 to B

In the end, B should belong to workspace Y and project Y2. A should belong to workspace X and project X2. No errors should occur in the process.

### DIFF
--- a/src/model/project.h
+++ b/src/model/project.h
@@ -15,7 +15,7 @@ namespace toggl {
 
 class TOGGL_INTERNAL_EXPORT Project : public BaseModel {
  public:
-    Project() : BaseModel() {}
+    Project() : BaseModel() { EnsureGUID(); }
 
     Property<std::string> Name { "" };
     Property<std::string> Color { "" };
--- a/src/context.cc
+++ b/src/context.cc
@@ -6537,8 +6537,12 @@ error Context::syncHandleResponse(Json::Value &array, const std::vector<T*> &sou
                     }
 
                     model->LoadFromJSON(i["payload"]["result"], isUsingSyncServer());
-                    model->ClearDirty();
                     model->ClearUnsynced();
+                    error err = save(false);
+                    if (err != noError) {
+                        displayError(err);
+                        return err;
+                    }
                 }
             }
             else if (i["payload"]["result"].isMember("error_message") && i["payload"]["result"]["error_message"].isMember("default_message")) {
--- a/src/context.cc
+++ b/src/context.cc
@@ -5572,11 +5572,17 @@ error Context::pushBatchedChanges(
         std::cerr << "RESPONSE: " << responseJson.toStyledString() << std::endl;
         logger.debug("Sync response to request ", lastRequestUUID_, ": ", request.toStyledString());
 
-        syncHandleResponse(responseJson["clients"], clients);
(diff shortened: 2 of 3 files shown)


src/model/project.h
```diff
@@ -15,7 +15,7 @@ namespace toggl {
 
 class TOGGL_INTERNAL_EXPORT Project : public BaseModel {
  public:
-    Project() : BaseModel() {}
+    Project() : BaseModel() { EnsureGUID(); }
```
skel35, line 18: why is EnsureGUID() needed here specifically for Project?

MartinBriza: In my database, some projects ended up not having GUIDs, to be able to sync them properly, I need the GUID to be there at all times. In case this GUID is loaded from the database, it gets overwritten before it can be used elsewhere in the database.

skel35:
just discovered a bug with the default project getting changed, not sure if it'll reproduce on all platforms or Windows only
new-prj-becomes-default

skel35:
just checked, it reproduces on `master` as well, so it's not caused by this PR specifically

skel35:
can't reproduce it on another account for some reason, only on @toggl.com account

MartinBriza:
If it's already in master and is not related to this particular PR, please not get stuck on this and let's fix that separately.

src/model/project.h
MartinBriza: to be honest, no, but having it there doesn't hurt. guid should be there by default for all entities but unless we're able to reproduce it for other entities, i don't think it makes sense to add it elsewhere now

skel35:
Finally found a bug with very concrete steps to reproduce:
1. Clear cache and log in to the account with multiple workspaces (both can be free workspaces)
2. Start a TE
3. Set a project to W1 P1
4. Set a project to W2 P1
5. Press Sync - everything looks normal
6. Open webapp - TE is duplicated with different projects, neither of the duplicates is marked as deleted (can be checked in BO)
7. Open desktop app again, clear cache and log in - duplicated TE are shown now

MartinBriza:
Thanks for catching this, Taras, this was actually a symptom of a much bigger issue that basically prevented deletes from happening completely. It's fixed now with the latest commit.

src/model/project.h
MartinBriza: I can remove this line. However, the fact if the TE is synced or not is determined by if it has an ID or not. Having an extra GUID there right away doesn't change current behavior.

PR #4248, opened 2020-07-16
Sync: Only load last 10 days when starting (lib)

### LINKED ISSUE #4242: New syncing logic loads old time entries
### 💻 Environment
Platform: macOS/Windows/Linux (repro'd on Windows)
Toggl Version: 7.5.220

### 🐞 Actual behavior
Desktop app loaded TEs all the way from Apr 16 (2 months old)

### 💯 Expected behavior
Desktop app should load TEs until 10 days old initially, like the old syncing logic.

### 🔨 Steps to reproduce
1. Turn the desktop sync client flag on
2. Start the app

MartinBriza:
### 📒 Description
$subj

### 🕶️ Types of changes

### 👫 Relationships
Fixes #4242

### 🔎 Review hints
Only the last 10 days should be loaded initially. Loading more should work right again.

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -5404,6 +5404,11 @@ error Context::pullBatchedUserData() {
         if (user_->HasValidSinceDate()) {
             since = user_->Since();
         }
+        else {
+            // just pull the last 10 days on full sync
+            since = time(nullptr) - 10 * 24 * 60 * 60;
+            user_->HasLoadedMore.Set(false);
+        }
     }
 
     if (api_token.empty()) {


skel35:
I'm still getting some older TEs pulled as well, not all of them though.
image

skel35:
I've figured it out - I'm getting the older TEs only from one particular workspace which is currently my default workspace. Older TEs from the other workspaces are not being pulled.

skel35:
From the first glance, it looks like a sync server bug to me. I suppose all clients would have to do this filtering and it's quite counter-productive.

MartinBriza:
Then I guess we can assume this at least a partial fix. Sorry, I don't really have enough time to work on this further today (and tomorrow won't be a good day either, I assume).

skel35:
Sure. I'll merge this for now.

I've written into the #sync-testing channel so that someone else could verify this is indeed a server bug.

PR #4249, opened 2020-07-17
Fix bug with SSO linking using Google account

AndrewVebster:
### 📒 Description
After login with SSO for the first time, the user needs to link it with the existing Toggl account.
Previously this linking worked only for simple email&pass login. This commit also allows to use Google/Apple login to perform this linking.

### 🕶️ Types of changes
May break existing SSO functionality.

### 🤯 List of changes
- Returned back `sso_confirmation_code` and `need_enable_SSO` flag in the `Context` class.
- `LoginViewController` resets those properties when needed

### 🔎 Review hints
SSO needs to be retested again.
Additionally, make sure to try linking SSO with Google/Apple sign in.

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -87,7 +87,9 @@ Context::Context(const std::string &app_name, const std::string &app_version)
 , update_path_("")
 , overlay_visible_(false)
 , is_using_sync_server_(false)
-, last_message_id_("") {
+, last_message_id_("")
+, need_enable_SSO(false)
+, sso_confirmation_code("") {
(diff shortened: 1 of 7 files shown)


skel35:
Testing SSO linking with Google login on Staging, I'm getting "Invalid or expired confirmation code"
Screenshot 2020-07-20 at 10 44 33

AndrewVebster:
@skel35 when debugging I see that confirmation code is sent and it's correct. I suppose the issue is with the staging configuration.
Could you please test on Prod?

PR #4251, opened 2020-07-18
Update macOS project structure

AndrewVebster:
### 📒 Description
This PR aims to improve the structure of the macOS project. 
Previously, looking at folders, it was hard to understand what the app was doing; searching for a specific feature implementation was time-consuming. This was especially true for the old Objective C code.

Now structure looks like this:
Screenshot 2020-07-18 at 14 46 34
Screenshot 2020-07-18 at 14 47 56
Screenshot 2020-07-18 at 14 47 21
Screenshot 2020-07-18 at 14 47 40

### 🕶️ Types of changes

### 🤯 List of changes
- A lot of files were moved to a different location
- Removed test2 folder 😄
- Removed unused classes (custom views)
- Moved poco and openssl dylibs from `osx/TogglDesktop/Library` to `osx/TogglDesktop/TogglDesktopLibrary/Dependencies`
- Clean up project file - removed empty records and updated Build Settings (fixed 6 build warning)

### 🔎 Review hints
Checkout the branch and try to build the app. See if the new project structure makes sense for you.


AndrewVebster:
Work is done but leaving it as DRAFT because don't want to merge it before we release Sync & SSO to Beta.

skel35:
The choice of classes in `TogglDesktopLibrary/Models` seems arbitrary - why is `CountryViewItem` there but `AutocompleteItem` isn't?
I would go with a stronger Models/Views/Services separation rather than feature separation. But I'm curious about what kind of structure you're more used to.

I love the removal of unused code and getting rid of warnings.

Obviously the structure looks way better than it was before, so good job :)

AndrewVebster:
Thanks :)

About this structure. I know two main ways to group files: by type or by modules. Here I chose Modules/Features separation for such reasons:
1. I think it expresses the app's core business purpose more clearly. E.g. if newcomer opens the project he easily can see that app has a timer, some list, timeline, onboarding, etc.
2. This helps to build an app in a more modular way. Folders do not strictly enforce modularity but in the future, they can be easily replaced with libraries. This is basically related to the next item.
3. iOS team is building its app with a modular architecture - every feature is a separate module/library. So I thought that it'll be nice and beneficial to have a similar structure.

Regarding `TogglDesktopLibrary/Models`. There I wanted to store models that are mapped from Library models into Objective C / Swift. In some way, they can be seen as a part of the Library and its Entity layer. So I'd move `AutocompleteItem` there too.

So what do you think?

skel35:
I see your idea behind features separation and it makes perfect sense 👍 
My biggest issue is that models are scattered around, but if you're going to move them to the Models folder than it looks logical and is fine by me.

AndrewVebster:
Moved library-related `AutocompleteItem`, `AutotrackerRuleItem`, `Settings` and `DesktopLibraryBridge` to `TogglDesktopLibrary` folder.

Marked this PR as "ready for review" - I suppose it's safe to merge it now.

PR #4256, opened 2020-07-22
Sync: Fix loading of user preferences

### LINKED ISSUE #4254: "Record activity" setting is always being displayed as OFF (doesn't get persisted)
### 💻 Environment
Platform: Windows
Toggl Version: 7.5.227

### 🐞 Actual behavior
"Record activity" setting is shown as OFF

### 💯 Expected behavior
"Record activity" setting is shown as ON and the activity recording is working

### 🔨 Steps to reproduce
1. Turn "Record activity" setting ON. (The fact that it's ON can also be verified in BO)
2. Quit the app.
3. Start the app.

### 📦 Additional info
Can be reproduced only if the user has `dekstop_sync_client` flag on.

MartinBriza:
### 📒 Description
In Sync, the user preference JSON is split from the user and put in the "preferences" part of the server response. This PR fixes handling this by moving preference parsing to `User::LoadUserPreferencesFromJSON`. 
The same method is used in `Context::pullUserPreferences` which is then in turn used when logging in and when syncing.
It's probably not necessary to have it there in the future but for the sake of keeping the same behavior before we get rid of legacy sync and that a UI reload depends on it, I'm keeping it in place for now.

Also, as a part of fixing this, I noticed the `User::StoreStartAndStopTime` property is not used anywhere, so I got rid of it. It's still present in the `users` database table but SQLite doesn't support column drops so I just left it there, it doesn't feel like a big burden. I left a small note there for anybody who's modifying the table in the future that they can drop it.

### 🕶️ Types of changes

### 👫 Relationships
Fixes #4254 

### 🔎 Review hints
When using both sync and legacy protocols, the following properties should be handled right:
 * `RecordTimeline`
 * `TimeOfDayFormat`
 * `DurationFormat`
 * `CollapseTimeEntries`

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -6396,7 +6396,7 @@ error Context::pullUserPreferences() {
             return error("Failed to load user preferences");
         }
 
-        if (user_->LoadUserPreferencesFromJSON(root)) {
+        if (user_->LoadUserPreferencesFromJSON(root, false)) {
             // Reload list if user preferences
             // have changed (collapse time entries)
             UIElements render;
--- a/src/database/database.cc
+++ b/src/database/database.cc
@@ -1191,7 +1191,6 @@ error Database::LoadUserByID(
         std::string fullname("");
         std::string email("");
         bool record_timeline(false);
-        bool store_start_and_stop_time(false);
         std::string timeofday_format("");
         std::string duration_format("");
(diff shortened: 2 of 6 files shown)


skel35:
merged into release branch releases/v7.5.228

Issue #1824, opened 2016-02-05
Active app loses focus when timer is started with global shortcut


IndrekV:
User reports:

> Would be very handy if current active app remains active after hiding Toggl desktop app via hotkey. Currently you should press extra Cmd-Tab to activate it again.

sonayr:
Would love to see this fixed too, this little thing makes the desktop app feel clunky.

IndrekV:
@sonayr Can you write down a scenario about this issue. It's pretty old and I'm not sure what the exact case is here.

sonayr:
@IndrekV Sure thing, the issue here is when you use the global hotkey to show/hide the Toggl Desktop app on Mac it takes focus which is great but when you press the hotkey again to hide the app it doesn't give focus back to the application that previously had it. For example, if I have Chrome open and use the hotkey to show Toggl then re-use the same hotkey to hide Toggl and press the hotkey for new-window on chrome (Cmd-N) it instead starts a new timer for Toggl since it never resumed focus back to chrome (you have to click or Cmd-Tab to regain). Apps that do this really well on Mac are Todoist with its global hotkey to add new tasks and Iterm2 with its hide/show global hotkey.

PR #4271, opened 2020-07-24
Fix "Show dock icon" settings behaviour (mac)

### LINKED ISSUE #4117: Toggl icon still appears in the Dock even when the setting is turned off
### 💻 Environment

Platform:  macOS
OS Version:  macOS 10.15.4
Toggl Version:  7.5.146

### 🐞 Actual behavior

In TogglDesktop preferences, the Show dock option is turned off but the Toggl still appears in the Dock.

### 💯 Expected behavior

If the setting for Show dock icon is turned off, Toggl shouldn't appear in the dock

### 🔨 Steps to reproduce

I was able to replicate it using these steps:

1. Make sure that `Show recent applications in Dock` is checked in Dock preferences
2. Go to TogglDesktop - Preferences - uncheck the option for `Show dock icon`
3. Toggl icon will still remain in the Dock

### 📦 Additional info

If `Show recent applications in Dock` isn't checked in Dock preferences, Toggl icon will disappear as it should.

[Case reference](https://app.intercom.com/a/apps/ayixs927/inbox/2759100/conversations/27282211045) it includes the video of user replicating the issue

AndrewVebster:
### 📒 Description
Fixed bug when in preferences, the "Show dock icon" option is turned off but the Toggl still appears in the Dock.
Also if a user has a global shortcut to show/hide the app and quickly press it many times, the app icon could be duplicated.
See Issue #4117 for more details about the bug.

To fix the bug, removed two lines that were called on the show/hide global shortcut. The documentation doesn't explain why they are there, but they're definitely causing the issues.

### 🕶️ Types of changes

### 👫 Relationships
Closes #4117 

### 🔎 Review hints
- Test the "Show dock icon" option
- Check if no side effects are present when using show/hide global shortcut (or menu item)

Please also:
[] Test on old macOS version (I could not do it because haven't set it up yet)

### DIFF
--- a/src/ui/osx/TogglDesktop/App/AppDelegate.m
+++ b/src/ui/osx/TogglDesktop/App/AppDelegate.m
@@ -244,8 +244,6 @@ void *ctx;
 		 if ([[NSApplication sharedApplication] isActive] && [strongSelf.mainWindowController.window isVisible])
 		 {
 			 [self.mainWindowController.window close];
-             [NSApp setActivationPolicy:NSApplicationActivationPolicyProhibited];
-             [NSApp setActivationPolicy:NSApplicationActivationPolicyRegular];
 		 }
 		 else
 		 {


skel35:
haven't tested yet, but I've noticed this https://github.com/toggl-open-source/toggldesktop/commit/d5e9faf699016d6d332e9276d68a21d9220e668a.
@AndrewVebster can you check that old bug which is linked to this commit?

skel35:
that bug was there from 2016 without much activity, so I'd still go for your fix 😄

AndrewVebster:
Yeah, I've looked at #1824. I wouldn't say that this was a bug fix. It's more like an override for the default macOS behavior. When you close the app and it continues to run, the menu stays visible. It works like this for all apps that I know.
Changing this introduced strange behavior that we have now.

skel35:
OK, then I'll merge this one and reopen #1824

PR #4275, opened 2020-07-27
Try to fix crash

### LINKED ISSUE #4247: Windows app starts permanently crashing on Windows 7 when running it while updates are applied
### 💻 Environment
Platform: Windows
OS Version: Windows 7
Toggl Version: 7.5.193, 7.5.213

### 🐞 Actual behavior
Either
Windows app starts permanently crashing on startup.
Or
Windows app process hangs together with the updater process.

### 💯 Expected behavior
Windows app should run successfully.

### 🔨 Steps to reproduce
1. Run the older version of the app to make sure the update is being downloaded.
2. Wait for the update to download.
3. Quit the app.

Either

4. Try to quickly launch the app multiple times.

Or

4. Restart the PC.

### 📦 Additional info
Doesn't seem to reproduce on Windows 10.

[Case 1](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/2097915/conversations/106481300030303)
[Case 2](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/2097915/conversations/106481300030252)
[Case 3](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/2097915/conversations/106481300030260)

atsvirchkova:
### 📒 Description
There is a bug reproducible on Windows7: after downloading update and restarting the app, it crashes on startup. I think it may happen because some error appears before `ctx` initialization and Bugsnag crashes as it needs `ctx` to report the error.

### 🕶️ Types of changes

### 👫 Relationships

 Closes #4247 

### 🔎 Review hints
I'm not sure that this actually fixes the issue, as I didn't manage to reproduce the bug again: it reproduced after installing Toggl on clean Windows and then it crashed every time, when I tried to start it. But then after installing the earlier version and updating it to the newest one again, it worked properly.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/Program.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/Program.cs
@@ -32,7 +32,6 @@ static class Program
         Toggl.OnLogin += delegate (bool open, ulong user_id) {
             UserId = user_id;
         };
-        BugsnagService.Init();
         singleInstanceManager.BeforeStartup -= OnBeforeStartup;
     }
 
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/windows/MainWindow.xaml.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/windows/MainWindow.xaml.cs
@@ -255,6 +255,8 @@ namespace TogglDesktop
                 return;
             }
 
+            BugsnagService.Init();
+
             this.loadPositions();
 
             this.GetWindow<AboutWindow>().ViewModel.InitUpdateChannel(Toggl.UpdateChannel());


atsvirchkova:
I think you're right. I could leave the Bagsnag initialization where it is and instead implement the `UpdateChannel()` like this:
```
    public static string TryGetUpdateChannel()
    {
        return ctx!=IntPtr.Zero ? toggl_get_update_channel(ctx) : "";
    }
```

skel35:
This solution might also be dangerous as there's a moment when `ctx` is no longer `IntPtr.Zero`, but update channel is not initialized. So I would go for a safer solution.

skel35:
Closing in favor of #4331, #4333

PR #4277, opened 2020-07-28
Do not set focus to project text box after new client is created

### LINKED ISSUE #4184: After creating a client focus jumps to project name field
### 💻 Environment
Platform: Windows
Toggl Version: 7.5.193

### 🐞 Actual behavior
After creating a client focus jumps to project name field

### 💯 Expected behavior
After creating a client focus should not jump

### 📦 Additional info

skel35:
### 📒 Description
Do not set focus to project text box after a new client is created.
This logic has been there since the old design where it could've been relevant.

### 🕶️ Types of changes

### 🤯 List of changes
- Do not set focus to project text box after a new client is created.

### 👫 Relationships
Closes #4184 

### 🔎 Review hints
see #4184

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/views/EditView.xaml.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/views/EditView.xaml.cs
@@ -675,6 +675,11 @@ namespace TogglDesktop
             var item = asClientItem.Model;
 
             this.selectClient(item);
+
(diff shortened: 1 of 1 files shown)


skel35:
1. first commit only fixed the jump during autocomplete, added a fix for client creation as well.
2. I suppose jumping to the project textbox can still be useful if the project textbox is empty, as it needs to be filled anyway before the project can be created.

ready for review

PR #4284, opened 2020-07-31
Sync: Don't try to push broken data to backend

MartinBriza:
### 📒 Description
This PR adds a very simple check to the generation of the Sync JSON. If a foreign entity (client, project) GUID cannot be found, set a validation error to the entity and mark it as unsynced. For TEs, it will get displayed, for Projects and Clients, this will result in marking the TEs that use them as unsynced.

In an ideal case, this should never be triggered, it's a safeguard for potential issues in Sync implementation on our or server side.

### 🕶️ Types of changes

### 🔎 Review hints
The easiest way to reproduce this is adding a broken item to your database directly. Be sure to disable foreign keys by running `PRAGMA foreign_keys = 0` if you're using SQLiteBrowser3 because it enables them by default and TogglDesktop has them disabled. (Also, the foreign key wouldn't really help us in this case because it's only created for the `cid` column that actually CAN point nowhere.)

### DIFF
--- a/src/const.h
+++ b/src/const.h
@@ -53,6 +53,7 @@
 #define kInvalidDateError "Date year must be between 2006 and 2030"
 #define kStartNotBeforeStopError "Stop time must be after start time"
 #define kMaximumDescriptionLengthError "Maximum length for description (3000 chars) exceeded"
+#define kForeignEntityLost "Assigned foreign entity could not be found"
 
 #define kCheckYourSignupError "Signup failed - please check your details. The e-mail might be already taken."  // NOLINT
 #define kEndpointGoneError "The API endpoint used by this app is gone. Please contact Toggl support!"  // NOLINT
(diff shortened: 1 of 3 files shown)


src/const.h
```diff
@@ -53,6 +53,7 @@
 #define kInvalidDateError "Date year must be between 2006 and 2030"
 #define kStartNotBeforeStopError "Stop time must be after start time"
 #define kMaximumDescriptionLengthError "Maximum length for description (3000 chars) exceeded"
+#define kForeignEntityLost "Assigned foreign entity could not be found"
```
AndrewVebster, line 56:
```suggestion
#define kForeignEntityLostError "Assigned foreign entity could not be found"
```

AndrewVebster:
@MartinBriza approved, but propose to rename `kForeignEntityLost` to `kForeignEntityLostError` to keep the same naming pattern.
The rest looks good to me. You can merge now or after you change the name.

skel35:
@MartinBriza Please check this out so that we can include this PR into the next release.

PR #4285, opened 2020-07-31
Sync: Handle when only ID/GUID changed when removing a project 

MartinBriza:
### 📒 Description
About what subject says. In some cases, when removing a project from a time entry, nothing happens. This fixes it.

### 🕶️ Types of changes

### 👫 Relationships
I think there's no bug reported for this

### 🔎 Review hints
Try adding and removing projects in various situations.


skel35:
the logic seems good, but testing is blocked right now due to sync server errors 🚫

Issue #4269, opened 2020-07-24
Timeline UI on Windows


skel35:
### 💻 Environment
Platform: Windows

### 📒 Description
This is a primary issue for the "Timeline UI on Windows" feature.

Designs: https://zpl.io/amkgjjK.

For initial release:
- [x] List/Timeline tabs, use library methods (M also include flag implementation in back office) #4289
- [x] Move Record activity ToggleSwitch from Preferences to Timeline tab (S) #4290 
- [x] Info icon near Record activity with tooltip on hover (tooltip content same as in macOS app) (S) #4290
- [x] add DatePicker and connect days switching to the library methods (S) #4320
- [x] implement the DatePicker style with left/right buttons (M) #4320
- [x] split the main view into 3 blocks (time, time entries, activity) (XS) #4354
- [x] empty states for time entries and activity blocks (S) #4354
- [x] add time marks for time blocks + make every other hour darker (S) #4354
- [x] add Activity visualization (with info on hover) (L) #4370
- [x] display current time (S) #4270
- [x] add basic Time Entry pills visualization with an algorithm for laying them out when TEs overlap (note: designs are logically INCORRECT, refer to macOS app for the correct logic) (L) #4370
- [x] add buttons for creating TEs from gaps (M) #4370
- [x] show full TE details and not just pills if there is no overlap (S) #4370
- [x] add zooming (with +- buttons, Ctrl + and Ctrl - shortcuts) (M) #4408
- [x] add tooltips when hovering over TEs (S) #4409
- [x] open Edit view by click (XS) #4410
- [x] "Record activity" switch in the Timeline tab does not get synchronized with the value in the Preferences #4417
- [x] Timeline tab is shown with `desktop_timeline_ui` flag off #4418

Can be done after the initial release:
- [x] resizing TEs to change the start time or duration (S) #4434
- [ ] dragging TEs to change the start time (M) #4435
- [ ] right-click context menu (items for resolving overlaps (change first entry stop time, change last entry start time), Continue TE, Start from the end of this TE, Stop TE) (not present in designs, refer to macOS app) (S)
- [ ] click-drag to create TE (implement click to create first if click-drag is too difficult to do)
- [ ] onboarding popup (not present in designs)
- [ ] analytics tracking of context menu actions (S)

Optional
- [ ] sticky header when scrolling
- [ ] pinch to zoom

Notes:
- Designs have a horizontal scroll, but there's no need to add it as no users have this many overlapped TEs.
[...]

PR #4288, opened 2020-08-03
List/Timeline tabs and using timeline_ui flag from lib

### LINKED ISSUE #4289: Implement List/Timeline tabs and create timeline_ui flag in lib
### 💻 Environment

Platform: Windows

### 📒 Description
Implement List/Timeline tabs and create timeline_ui flag in lib.
This issue is part of bigger #4269

atsvirchkova:
### 📒 Description
Implementation of tabs switch for #4269

### 🕶️ Types of changes

### 🤯 List of changes
TimerEntryListView.xaml

### 👫 Relationships

 Closes part of #4269 Closes #4289 

### 🔎 Review hints
A variable `timeline_ui_flag` created in context.cc assuming that later it will be bound to a flag from back office.

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -7028,6 +7028,10 @@ bool Context::checkIfSkipPomodoro(TimeEntry *te) {
     return false;
 }
 
+bool Context::IsTimelineUiEnabled() {
+    return user_->timeline_ui_flag;
+}
+
 bool Context::isUsingSyncServer() const {
     return is_using_sync_server_;
 }
--- a/src/model/user.h
+++ b/src/model/user.h
@@ -210,6 +210,8 @@ class TOGGL_INTERNAL_EXPORT User : public BaseModel {
         const std::string &password,
         std::string *result);
 
+    bool timeline_ui_flag = true;
+
  private:
     void loadUserTagFromJSON(
         Json::Value data,
--- a/src/toggl_api.h
+++ b/src/toggl_api.h
@@ -1301,7 +1301,7 @@ extern "C" {
         const char_t *hexColor,
         TogglAdaptiveColor type);
 
-    TOGGL_EXPORT bool is_timeline_ui_enabled(
+    TOGGL_EXPORT bool toggl_is_timeline_ui_enabled(
         void* context);
 
 #undef TOGGL_EXPORT
--- a/src/toggl_api.h
+++ b/src/toggl_api.h
@@ -1301,7 +1301,7 @@ extern "C" {
         const char_t *hexColor,
         TogglAdaptiveColor type);
 
-    TOGGL_EXPORT bool toggl_is_timeline_ui_enabled(
+    TOGGL_EXPORT bool_t toggl_is_timeline_ui_enabled(
         void* context);
 
 #undef TOGGL_EXPORT
--- a/src/toggl_api.h
+++ b/src/toggl_api.h
@@ -332,7 +332,8 @@ extern "C" {
     typedef void (*TogglContinueSignIn)(
     );
 
-    typedef void (*TogglDisplayLoginSSO)(const char_t *sso_url);
+    typedef void (*TogglDisplayLoginSSO)(
+        const char_t *sso_url);
 
     typedef void (*TogglDisplayIdleNotification)(
         const char_t *guid,
@@ -1299,6 +1300,9 @@ extern "C" {
     TOGGL_EXPORT TogglRgbColor toggl_get_adaptive_rgb_color_from_hex(
         const char_t *hexColor,
         TogglAdaptiveColor type);
+
+    TOGGL_EXPORT bool is_timeline_ui_enabled(
+        void* context);
 
 #undef TOGGL_EXPORT
 
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/views/TimerEntryListView.xaml
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/views/TimerEntryListView.xaml
@@ -12,12 +12,83 @@
             <RowDefinition />
         </Grid.RowDefinitions>
 
-        <toggl:TimeEntryList Grid.Row="1"
-                             x:Name="Entries"
-                             FocusTimer="onFocusTimer"/>
+        <TabControl Grid.Row="1" Name="TabControl">
+            <TabControl.Resources>
+                <Style TargetType="TabItem">
+                    <Setter Property="Template">
+                        <Setter.Value>
+                            <ControlTemplate TargetType="TabItem">
+                                <StackPanel Background="{DynamicResource Toggl.CardBackground}">
+                                    <TextBlock x:Name="ContentSite"
+                                                   VerticalAlignment="Center"
+                                                   HorizontalAlignment="Center"
+                                                   Text="{TemplateBinding Header}"
+                                                   Margin="12,2,12,2"
+                                                   Style="{StaticResource Toggl.BodyText}"/>
+                                    <Border Name="Border"
+                                                BorderBrush="{DynamicResource Toggl.AccentBrush}" 
+                                                BorderThickness="1,1,1,1">
+                                    </Border>
+                                </StackPanel>
+                                <ControlTemplate.Triggers>
+                                    <Trigger Property="IsSelected" Value="True">
+                                        <Setter TargetName="Border" Property="BorderThickness" Value="1" />
+                                    </Trigger>
+                                    <Trigger Property="IsSelected" Value="False">
+                                        <Setter TargetName="Border" Property="BorderThickness" Value="0" />
+                                    </Trigger>
+                                </ControlTemplate.Triggers>
+                            </ControlTemplate>
+                        </Setter.Value>
+                    </Setter>
+                </Style>
+                <Style TargetType="TabControl">
+                        <Setter Property="Template">
+                            <Setter.Value>
+                                <ControlTemplate TargetType="{x:Type TabControl}">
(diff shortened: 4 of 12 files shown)


src/toggl_api.h
```diff
@@ -1300,6 +1301,9 @@ extern "C" {
         const char_t *hexColor,
         TogglAdaptiveColor type);
 
+    TOGGL_EXPORT bool toggl_is_timeline_ui_enabled(
```
skel35, line 1304: as this is only used on Windows it works, but for cross-platform we should use `bool_t`

src/ui/windows/TogglDesktop/TogglDesktop/ui/views/TimerEntryListView.xaml
```diff
@@ -4,20 +4,93 @@
...
     <Grid>
         <Grid.RowDefinitions>
             <RowDefinition Height="Auto"/>
             <RowDefinition />
         </Grid.RowDefinitions>
 
-        <toggl:TimeEntryList Grid.Row="1"
-                             x:Name="Entries"
-                             FocusTimer="onFocusTimer"/>
+        <TabControl Grid.Row="1" Name="TabControl" SelectedIndex="{Binding SelectedTab}">
```
skel35, line 17: Personally I think the actual TimeEntryListView layout will be more readable if the styles for TabItem and TabControl are moved into the resources, even if they are used only once. But I don't have a strong opinion on this. What's the benefit of putting the styles together with the view?

src/ui/windows/TogglDesktop/TogglDesktop/ui/views/TimerEntryListView.xaml
```diff
@@ -4,20 +4,93 @@
...
+                                        <Border x:Name="Border"
+                                                Grid.Row="1" Grid.ColumnSpan="2"
+                                                BorderThickness="{TemplateBinding BorderThickness}"
+                                                Background="{TemplateBinding Background}"
+                                                BorderBrush="{TemplateBinding BorderBrush}">
+                                            <ContentPresenter x:Name="PART_SelectedContentHost" ContentSource="SelectedContent" />
+                                        </Border>
+                                </Grid>
+                            </ControlTemplate>
+                            </Setter.Value>
```
skel35, line 76: The formatting (tabs) is off which makes it hard to read.

Do you think we should start using https://github.com/Xavalon/XamlStyler ?

skel35:
one more comment regarding UI - the height of the tab headers doesn't match the design in Zeplin

src/ui/windows/TogglDesktop/TogglDesktop/ui/views/TimerEntryListView.xaml
atsvirchkova: We already have a TabControl style in Resources, and it is automatically applied to all tabcontrols in the app. It is used in Preferences, but I was not able to use this style for this case as I needed slightly different behavior. I can create one more TimeEntryTabControl in Resources and with the `x:Key` property, but then I'll have to specify the name of the style for each tab (I know, there are only 2 tabs, but still it's more convenient to have a style applied to all tabs automatically).

atsvirchkova:
Ready for review. I tried to fix the pixel blurring by adding `UseLayoutRounding` property to TabControl, but as I cannot reproduce it, I don't know if it works.

src/model/user.h
```diff
@@ -210,6 +210,8 @@ class TOGGL_INTERNAL_EXPORT User : public BaseModel {
         const std::string &password,
         std::string *result);
 
+    bool timeline_ui_flag = true;
```
skel35, line 213: changing this to `false` doesn't hide the tab switcher

src/ui/windows/TogglDesktop/TogglDesktop/ui/views/TimerEntryListView.xaml
```diff
@@ -4,20 +4,101 @@
              xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" 
              xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
              xmlns:toggl="clr-namespace:TogglDesktop"
+             xmlns:vm="clr-namespace:TogglDesktop.ViewModels"
+             xmlns:mah="http://metro.mahapps.com/winfx/xaml/controls"
              mc:Ignorable="d" 
-             d:DesignHeight="300" d:DesignWidth="400">
+             d:DesignHeight="300" d:DesignWidth="400"
+             d:DataContext="{d:DesignInstance vm:TimerEntryListViewViewModel, IsDesignTimeCreatable=False}">
```
skel35, line 11: I'm seeing newly introduced binding errors, might explain why switching the timeline flag to false doesn't do anything:
>System.Windows.Data Error: 40 : BindingExpression path error: 'IsTimelineViewEnabled' property not found on 'object' ''MainWindow' (Name='')'. BindingExpression:Path=IsTimelineViewEnabled; DataItem='MainWindow' (Name=''); target element is 'Border' (Name=''); target property is 'Visibility' (type 'Visibility')
System.Windows.Data Error: 40 : BindingExpression path error: 'SelectedTab' property not found on 'object' ''MainWindow' (Name='')'. BindingExpression:Path=SelectedTab; DataItem='MainWindow' (Name=''); target element is 'TabControl' (Name='TabControl'); target property is 'SelectedIndex' (type 'Int32')
System.Windows.Data Error: 40 : BindingExpression path error: 'IsTimelineViewEnabled' property not found on 'object' ''MainWindow' (Name='')'. BindingExpression:Path=IsTimelineViewEnabled; DataItem='MainWindow' (Name=''); target element is 'TabItem' (Name=''); target property is 'Visibility' (type 'Visibility')
System.Windows.Data Error: 40 : BindingExpression path error: 'IsTimelineViewEnabled' property not found on 'object' ''MainWindow' (Name='')'. BindingExpression:Path=IsTimelineViewEnabled; DataItem='MainWindow' (Name=''); target element is 'Border' (Name=''); target property is 'Visibility' (type 'Visibility')

src/ui/windows/TogglDesktop/TogglDesktop/ui/views/TimerEntryListView.xaml
```diff
@@ -4,20 +4,101 @@
...
         <Grid.RowDefinitions>
             <RowDefinition Height="Auto"/>
             <RowDefinition />
         </Grid.RowDefinitions>
 
-        <toggl:TimeEntryList Grid.Row="1"
-                             x:Name="Entries"
-                             FocusTimer="onFocusTimer"/>
+        <TabControl Grid.Row="1" Name="TabControl" SelectedIndex="{Binding SelectedTab}"
+                    UseLayoutRounding="True">
```
skel35, line 87: I'm not seeing blurriness anymore 👍

src/model/user.h
atsvirchkova: It does, but you have to make full rebuild for this. I don't know why, but changes in user.h are noticed by VS compiler.

src/ui/windows/TogglDesktop/TogglDesktop/ui/views/TimerEntryListView.xaml
atsvirchkova: Lucky I am :)

src/model/user.h
skel35: +

src/ui/windows/TogglDesktop/TogglDesktop/ui/views/TimerEntryListView.xaml
skel35: Ah, the binding errors are probably due to viewmodel not being assigned when the view is constructed.
I think in principle it's better to not delay the viewmodel construction. But I see that you've probably done this so that we don't call the library methods before the library is initialized. Not sure what's the best way to go about it, perhaps calling library methods should occur only in response to some events or observables.

atsvirchkova: Yes, you're right. If I assign ViewModel in constructor, then a call to lib method may happen before context is initialized. I think this would be an endless source of bugs if we allow theoretically to call a lib method before full initialization, though I also don't like this solution.

skel35: https://github.com/toggl-open-source/toggldesktop/commit/8bcab7c6ea26a36beaf8417ba3467e0b9c12f6ff
This is my proposal, initialize in the MainWindow's constructor, but after StartUI(). This way there are no binding errors.

atsvirchkova:
Ready for review. Moved ViewModel creation later in ManWindow.xaml.cs. Thank you for noticing this simple solution, this is really much better.

src/context.cc
```diff
@@ -7028,6 +7028,10 @@ bool Context::checkIfSkipPomodoro(TimeEntry *te) {
     return false;
 }
 
+bool Context::IsTimelineUiEnabled() {
+    return user_->timeline_ui_flag;
```
skel35, line 7032: `user_` is not checked for `nullptr` here, which can cause a crash. This is currently reproducible when starting the app while logged out.

This also uncovers the issue that `TimerEntryListViewViewModel`'s lifetime is bigger than the `user_`'s lifetime. We should probably initialize the viewmodel's properties or the entire viewmodel on some library event, e.g. `Toggl.OnLogin` which should fire only when user gets logged in.

atsvirchkova:
Fixed. Now initialize the ViewModel on LoginEvent.

PR #4292, opened 2020-08-05
Record activity switch

### LINKED ISSUE #4290: Record activity on Timeline tab
### 💻 Environment

Platform: Windows

### 📒 Description
Implement Record activity switch and tooltip for it.

This issue is part of #4269

atsvirchkova:
### 📒 Description
Record activity switch and Info icon near Record activity with tooltip on hover (tooltip content same as in macOS app).

### 🕶️ Types of changes

### 🤯 List of changes
Added Timeline.xaml and TimelineViewModel

### 👫 Relationships

Closes #4290 
Relates to #4269 

### 🔎 Review hints
I implemented the Info tooltip as popup, to have the behavior like in MacOS (it closes only on click, not when user drags the mouse to another place). Though I'm not sure it's the best way. Another issue with tooltip was that it's not possible to click the "Learn more" link.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
@@ -1,0 +1,28 @@
+﻿using System;
+using System.Reactive;
+using System.Threading.Tasks;
+using ReactiveUI;
+using ReactiveUI.Fody.Helpers;
+
+namespace TogglDesktop.ViewModels
+{
+    public class TimelineViewModel : ReactiveObject
+    {
+        public TimelineViewModel()
+        {
+            this.WhenAnyValue(x => x.RecordActivity).Subscribe(x => SaveRecordActivity());
+            OpenTogglHelpUri = ReactiveCommand.Create(() =>
+                Toggl.OpenInBrowser("https://support.toggl.com/en/articles/3836325-toggl-desktop-for-windows"));
+        }
+
+        [Reactive] 
+        public bool RecordActivity { get; set; } = Toggl.IsTimelineRecordingEnabled();
+
+        private void SaveRecordActivity()
+        {
+            Task.Run(() => Toggl.SetTimelineRecordingEnabled(RecordActivity));
+        }
+
+        public ReactiveCommand<Unit, Unit> OpenTogglHelpUri { get; set; }
+    }
+}
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Reactive;
+using System.Reactive.Linq;
 using System.Threading.Tasks;
 using ReactiveUI;
 using ReactiveUI.Fody.Helpers;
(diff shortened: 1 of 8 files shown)


src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        [Reactive] 
+        public bool RecordActivity { get; set; } = Toggl.IsTimelineRecordingEnabled();
+
+        private void SaveRecordActivity()
+        {
+            Task.Run(() => Toggl.SetTimelineRecordingEnabled(RecordActivity));
+        }
+
+        public ReactiveCommand<Unit, Unit> OpenTogglHelpUri { get; set; }
```
skel35, line 26: why is setter public?

src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
```diff
@@ -0,0 +1,28 @@
...
+using ReactiveUI;
+using ReactiveUI.Fody.Helpers;
+
+namespace TogglDesktop.ViewModels
+{
+    public class TimelineViewModel : ReactiveObject
+    {
+        public TimelineViewModel()
+        {
+            this.WhenAnyValue(x => x.RecordActivity).Subscribe(x => SaveRecordActivity());
```
skel35, line 13: There is a nicer way to offload a subscription to the TaskPool:
```suggestion
            this.WhenAnyValue(x => x.RecordActivity).ObserveOn(RxApp.TaskpoolScheduler).Subscribe(Toggl.SetTimelineRecordingEnabled);

```
and no need for `SaveRecordActivity()`

src/ui/windows/TogglDesktop/TogglDesktop/ui/views/Timeline.xaml
```diff
@@ -0,0 +1,76 @@
...
+            Record activity
+        </mah:ToggleSwitch>
+        <Viewbox Width="15" Height="15" Grid.Row="0" Grid.Column="1"
+                 HorizontalAlignment="Left"
+                 VerticalAlignment="Bottom"
+                 Margin="12 0 0 0"
+                 Name="RecordActivityInfoBox"
+                 MouseEnter="RecordActivityInfoBoxOnMouseEnter">
+            <Grid>
+                <Ellipse Fill="{DynamicResource Toggl.DisabledTextBrush}"
```
skel35, line 36: We already have a nice info icon that is used in the Project creation view. See `Toggl.InfoIconGeometry`.
image
P.S. the icon is also present in Zeplin designs of this page, you had to use it if you didn't know about the existing one.

atsvirchkova:
Ready for review. Here the VM is initialized on Load event. Actually I could do something like `timerEntryListView.Timeline.ViewModel = new TimelineViewModel()` from MainWindow.xaml.cs like in previous case. But I'm not sure this is good practice to initialize all the ViewModels in MainWindow just to be sure the context is already created. This breaks the encapsulation principle.

atsvirchkova:
Rebased and changed so that viewModel is initialized on Login event also.

atsvirchkova:
Ready for review.

src/ui/windows/TogglDesktop/TogglDesktop/ui/views/TimerEntryListView.xaml.cs
```diff
@@ -55,6 +55,7 @@ private void onLogin(bool open, ulong userID)
             if (userID != 0)
             {
                 ViewModel = new TimerEntryListViewViewModel();
+                Timeline.ViewModel = ViewModel.TimelineViewModel;
```
skel35, line 58: minor: I'd just set the Timeline's `DataContext` property in XAML

PR #4300, opened 2020-08-07
Timer Main UI - Part 1

AndrewVebster:
### 📒 Description
This PR is updating timer in the macOS app.

### 🕶️ Types of changes
- **Breaking change** (fix or feature that would cause existing functionality to change) - can break Edit popover or other related to timer features

### 🤯 List of changes
- Added a few methods to the library. I hope names say why do I need them.
- A lot of UI code

### 👫 Relationships
The design is here: https://www.notion.so/toggl/Design-overview-links-to-Miro-7b172f44c2654a10b54c069117721495#9394c5103656443eb7bf6a680eecb368

### 🔎 Review hints
Sorry guys for this huge PR 🙈. Next time will do it better 😅

Please test the app, play with the timer, and tell me about any bugs or UX issues.
But also, make sure it's not already in the list below.

### Things I plan to do before the release
- [x] Timer is not yet updated when running TE is edited in another place (Edit popup or another app)
- [x] Add a search text field to project and tags dropdowns
- [x] Add create button to project and tags dropdown
- [x] Make Duration text field editable
- [x] Fix project and tags dropdown height
- [x] Fix project and tags dropdown close action. Now it's closed when clicked outside but when you click on the button that opens the dropdown it hides and opens dropdown instead of just closing it
- [ ] Adaptive layout - move buttons in one line with the description text field when the width of the window is large enough
- [x] Change position of project and tags dropdown according to design (if needed)
- [x] "Billable rates is not on your plan" tooltip
- [x] Implement switching between controls with Tab


AndrewVebster:
Thanks @skel35 
The bug was addressed in the latest commit. Also added tooltip item to the list of things to do.

Created main feature branch `feature/timer-ui-part1` and set it as a base branch for this PR so when code review is done we can close this PR (it is ready for review now).

AndrewVebster:
> Switching keyboard focus with Tab isn't working

Just to clarify, switching focus between Description and Duration text fields or something else?

skel35:
I suppose in the final solution project, tags, and billable buttons and/or search fields should be focusable by the keyboard as well. Basically, it should be convenient to use the app with the keyboard only. When we get rid of the Edit view, the convenience of using the app with the keyboard shouldn't decrease.

For now, at least switching between description and duration text fields should work. It provides a nice quick flow to start the timer from a certain value rather than from zero.

skel35:
Bug 1: The upper part of the project selection button is highlighted when description autocomplete is shown, the highlight should be trimmed: 
image
Bug 2: The tags filtering text field in the Edit view looks messed up (not sure if it will reproduce on your side): 
image

AndrewVebster:
Fixed mentioned bugs. Also made all controls focusable by a keyboard.

src/ui/osx/TogglDesktop/Features/Timer/TimerViewController.swift
```diff
@@ -0,0 +1,533 @@
...
+extension TimerViewController: ProjectCreationViewDelegate {
+    func projectCreationDidCancel() {
+        closeProjectAutoComplete()
+    }
+
+    func projectCreationDidAdd(with name: String, color: String, projectGUID: String) {
+        closeProjectAutoComplete()
+
+        // TODO: set project on UI
+        // problem is we don't have library method to get project by GUID and pass it to view model
```
skel35, line 527: perhaps open an issue for this?

skel35:
more issues
- when checking/unchecking a tag many times in a row I see weird bugs - sometimes a completely different tag gets checked or the checking/unchecking isn't executed, or is executed twice
- when moving between tags with arrow keys right after the project or a tag was selected, the keyboard selection sometimes jumps to the top. Is tags' autocomplete reacting to too many notifications?
- should the billable button be focusable when it's disabled?

AndrewVebster:
@skel35 re the first comment:
Bug 1: 
Sorry, I cannot reproduce this or maybe don't understand. When I do STR the description stays the same.

Bug 2:
Will do that. I was thinking about this too but wasn't sure. 

Bug 3:
On macOS you trigger focused buttons with Space key.

skel35:
Bug 1: I reproduced it only on the third attempt (the gif is slow for some reason):
Screen-Recording-2020-08-17-at-10 53 05
The actual bug might be in the Edit view. The description shouldn't be overwritten if it hasn't been edited in the description field.

Bug 3: Space works as intended. I'm still a bit bothered that nothing happens when pressing Enter, but if users don't really do it then it's fine.

src/ui/osx/TogglDesktop/Features/Timer/TimerViewController.swift
AndrewVebster: opened an issue for this https://github.com/toggl-open-source/toggldesktop/issues/4334

AndrewVebster:
@skel35 all mentioned bugs are addressed.
I'm not sure only about this one

> when checking/unchecking a tag many times in a row I see weird bugs - sometimes a completely different tag gets checked or the checking/unchecking isn't executed, or is executed twice

I've extended the clickable area for the cells as it was possible before to click between two tags and this was weird. Hope this fixes the behavior you've mentioned.

AndrewVebster:
@skel35 good catch 👍 thank you!
fixed

PR #4323, opened 2020-08-12
Fetch desktop_timeline_ui flag from back office and set it for user

atsvirchkova:
### 📒 Description
This is needed to be able to merge feature/windows_timeline_ui branch into master. This PR makes use of beta feature flag from back office, which would allow to do staged rollout.

### 🕶️ Types of changes

### 🤯 List of changes
context.cc
user.cc

### 👫 Relationships

Relates to #4269 

### 🔎 Review hints
Not sure at all where I should call `SetIsTimelineUiEnabled()` in context. The current version seems working fine.

### DIFF
--- a/.github/workflows/pr_lint.yml
+++ b/.github/workflows/pr_lint.yml
@@ -1,0 +1,25 @@
+name: PR Lint
+on:
+  pull_request:
+    paths:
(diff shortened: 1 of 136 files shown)


skel35:
outdated in favor of #4343

PR #4343, opened 2020-08-20
Perform only one v9/me/preferences call

### LINKED ISSUE #4324: Perform only one v9/me/preferences call
### 💻 Environment
Platform: Lib

### 📒 Description
v9/me/preferences and v9/me/preferences/desktop responses are exactly the same, so there should be no reason to perform both calls. We should perform only one of these calls.

### Acceptance Criteria
1. The app performs only one v9/me/preferences (or v9/me/preferences/desktop) per user login.
2. The app might perform additional v9/me/preferences calls during "full sync" = sync being manually triggered.
3. There should be a clear way to read the feature flags that affect the UI that is displayed.

### Dependencies
As #4318 reduces the number of v9/me/preferences calls, this should be implemented on top of #4318.

atsvirchkova:
### 📒 Description
v9/me/preferences and v9/me/preferences/desktop responses are exactly the same, so there should be no reason to perform both calls. We should perform only one of these calls.

### 🕶️ Types of changes

### 🤯 List of changes

- [ ] context.cc

### 👫 Relationships

 Closes #4324 

### 🔎 Review hints
Test if user settings are updated on user login, on app startup and on full sync.

### DIFF
--- a/src/model/user.h
+++ b/src/model/user.h
@@ -211,7 +211,7 @@ class TOGGL_INTERNAL_EXPORT User : public BaseModel {
         std::string *result);
 
     AlphaFeatures* AlphaFeatureSettings;
-    void User::LoadAlphaFeaturesFromJSON(const Json::Value& data);
+    void LoadAlphaFeaturesFromJSON(const Json::Value& data);
 
  private:
     void loadUserTagFromJSON(
--- a/src/model/user.h
+++ b/src/model/user.h
@@ -14,6 +14,7 @@
 #include "related_data.h"
 #include "types.h"
 #include "model/workspace.h"
+#include "model/alpha_features.h"
 
 #include <Poco/LocalDateTime.h>
 #include <Poco/Types.h>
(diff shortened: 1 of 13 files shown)


MartinBriza (changes requested):
Looks good! I found some smaller things that should be fixed before merging but nothing too serious.

src/model/user.h
```diff
@@ -208,6 +210,9 @@ class TOGGL_INTERNAL_EXPORT User : public BaseModel {
         const std::string &password,
         std::string *result);
 
+    AlphaFeatures* AlphaFeatureSettings;
+    void User::LoadAlphaFeaturesFromJSON(const Json::Value& data);
```
MartinBriza, line 214: Remove User:: and it will build

atsvirchkova: Thank you for help :)

atsvirchkova:
Moved AlphaFeatures to user class. Now ready for review again.

MartinBriza (approved):
Thank you. From my testing everything works right and I like the quality of the code, good job!
I'm intentionally just approving but not merging the changes due to the current feature freeze that's in place.

PR #4344, opened 2020-08-20
Start fetching in-app messages on user login instead of app startup

### LINKED ISSUE #4342: In-app message is not shown if the user isn't logged in to the app on startup
### 💻 Environment
Platform: macOS/Windows/Linux
Toggl Version: 7.5.247

### 🐞 Actual behavior
In-app message is not shown.

### 💯 Expected behavior
In-app message is shown.

### 🔨 Steps to reproduce
Test the in-app notification, steps can be taken from https://github.com/toggl-open-source/toggldesktop/pull/4316
Make sure the user is logged out (e.g. delete the database file) before starting the app.

### 📦 Additional info
`fetchMessage` method is called only in `StartEvents()` method and only if the user is logged in to the app before starting it.

skel35:
### 📒 Description
Start fetching in-app messages on user login instead of app startup.

Notes:
`fetchMessage(0)` starts the periodic fetching task. It is fine to call it on user login because the periodic tasks are stopped on every user logout.

### 🕶️ Types of changes

### 👫 Relationships
Closes #4342 

### 🔎 Review hints

Testing in Staging:

Comment out these lines: https://github.com/toggl-open-source/toggldesktop/blob/master/src/context.cc#L1567-L1570
EITHER change the date on your computer to 1st September 2020 because the message should be shown since 31st August, or change OR comment out the code that checks the 'from' date.
Rebuild the solution.
Run the app, observe the notification.

Test the scenario where the user is logged out before starting the app.

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -308,7 +308,6 @@ error Context::StartEvents() {
             analytics_.TrackOs(db_->AnalyticsClientID(), os_info.str());
             analytics_.TrackOSDetails(db_->AnalyticsClientID());
         }
-        fetchMessage(0);
     } catch(const Poco::Exception& exc) {
         return displayError(exc.displayText());
     } catch(const std::exception& ex) {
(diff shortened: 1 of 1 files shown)


MartinBriza:
Works good, looks good, merging!

PR #4352, opened 2020-08-24
Use CMake to build on all platforms (in addition to native methods)

MartinBriza:
### 📒 Description
This PR adds the possibility to build the library and "Linux app" using CMake and native Qt on all platforms.
This is a small step in the direction of a future possibility to use CMake to build the library for our current native apps. It also makes my life slightly easier when I am in Windows or macOS and am working on a Linux UI feature. 

### 🕶️ Types of changes

### 🔎 Review hints

**Linux**

 * There should be no change to what's been going on until now

**Windows**
 * BUILD
   * You need to install Qt (anything above I think 5.11 should be fine, you need QtNetworkAuth) ideally from the official binary installer from https://qt.io (or using `aqtinstall`: https://github.com/miurahr/aqtinstall) and OpenSSL
   * When building in the terminal, you'll need (env) `CMAKE_PREFIX_PATH` to point to your Qt `/lib/cmake` directory (usually something like `C:/Qt/5.15.0/mingw81_64/lib/cmake`), QtCreator sets this up for you automatically. I tested only MinGW builds but in theory, MSVC should work as well.
   * This applies to QtCreator and terminal both: (env) `OPENSSL_ROOT_DIR` needs to point to your OpenSSL root dir.
   * This applies to QtCreator and terminal both: (CMake variable) `OPENSSL_ROOT_DIR` needs to point to your OpenSSL root dir. (Yes, I know, this should be simplified)
   * You need to have `mc.exe` in PATH! This is really annoying but POCO requires this and it's somewhere in `C:/Program Files (x86)/Windows Kits/10/bin/<SOMETHING>/x64/`
  * RUN
    * Again, pretty annoying, but you need to copy all built `.dll` files from your `build/src`, `build/bin` and `build/third_party` to `build/src/ui/linux/TogglDesktop`. 
    * And also the `plugins` folder from the Qt root needs to be copied over to `build/src/ui/linux/TogglDesktop`
    * If you're in the terminal, it's best to copy everything over from the Qt `/bin/*.dll` folder because I haven't found out how to set this up automatically, yet.
    * Qt Creator is not able to set these things up automatically as well :/ (yet)

**macOS**
 * BUILD
    * Again, you need to install Qt either from the official binary installer from https://qt.io or from brew (`brew install qt`). If you don't have OpenSSL development headers, get those as well (`brew install openssl`)
    * (env) `CMAKE_PREFIX_PATH` needs to point at Qt CMake folder, with brew it is here: `/usr/local/opt/qt5/lib/cmake`
    * (env) `OPENSSL_ROOT_DIR` needs to point at the openssl root dir: with brew: `/usr/local/opt/openssl@1.1/`
    * (CMake variable) `OPENSSL_ROOT_DIR` needs to point at the openssl root dir
 * RUN
    * `LD_LIBRARY_PATH` should point to the Qt lib dir and then the app should just run. If working with Qt Creator, it should just work automatically.

CMake variables are set either in Qt Creator or by using the following CMake commandline:
```
cmake -DOPENSSL_ROOT_DIR=/usr/local/opt/openssl@1.1/ path/to/source/dir
```

For more information on how the build actually works, consult the included GitHub Actions workflow, it's fairly simple and builds the "Linux app" on all three platforms using CMake.


MartinBriza:
image
Linux app running on Windows

atsvirchkova (approved):
I built and ran this CL on linux and it works fine. Unfortunately, failed to build and run it on Windows.

PR #4353, opened 2020-08-24
Make running Linux app against Prod and Staging possible at the same time 

MartinBriza:
### 📒 Description
In the vein of making my own life easier, this PR makes it possible to run the Staging app at the same time when the Production app is running. That especially means the logs and database will now get stored as `toggldesktop-staging` instead of just `toggldesktop` for testing versions.

### 🕶️ Types of changes

### 🔎 Review hints
/I'm aware it doesn't merge, I'll fix it soon/
Try downloading a binary package, for example our current stable tarball version (not flatpak, that runs isolated), run it. Then run the staging app from your own build. Both should run and you should have two databases in your `~/.local/share/Toggl/Toggl Desktop`.


atsvirchkova (approved):
Works good, merging it.

PR #4355, opened 2020-08-24
Main view markup (Timeline UI)

### LINKED ISSUE #4354: Main view markup (Timeline UI)
### 💻 Environment

Platform: Windows

### 📒 Description
split the main view into 3 blocks (time, time entries, activity)
add time marks for time blocks + make every other hour darker

atsvirchkova:
### 📒 Description
split the main view into 3 blocks (time, time entries, activity)
add time marks for time blocks + make every other hour darker

### 🕶️ Types of changes

### 👫 Relationships

 Closes #4354

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/App.xaml
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/App.xaml
@@ -31,6 +31,7 @@
             <converters:AdaptProjectTextColorConverter x:Key="AdaptProjectTextColorConverter" />
             <converters:BooleanInvertToVisibilityConverter x:Key="BooleanInvertToVisibilityConverter"/>
             <converters:HasAnyFlagToVisibilityConverter x:Key="HasAnyFlagToVisibilityConverter"/>
+            <converters:EmptyListToCollapsedConverter x:Key="EmptyListToVisibleConverter" Inverse="True"/>
         </ResourceDictionary>
     </Application.Resources>
 </Application>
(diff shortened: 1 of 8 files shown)


skel35 (changes requested):
Looks great, I have just 2 small comments:

1. The header should be visible at all times when scrolled downwards, like in the macOS app. Otherwise, this is a bad UX. Even the scrolling behavior from the designs (where header gets hidden but reappears when scrolling upwards) is a bit questionable for a desktop app, looks more suited to the mobile app IMHO.
2. Let's just use `Toggl.SecondaryTextBrush` and do not introduce the new `Toggl.TimelineTimeColor`, to simplify things. The designer could've made a mistake there, or just took the colors from elsewhere, but SecondaryTextBrush looks good here as well.

atsvirchkova:
I have a question about headers. I now looked into designs and it says, that the headers really must be always visible, I missed it. But it also says, that the area with date picker must scroll. In macOS, the date picker always stays on top, so as headers. So may I use the macOS solution for this?
Making so that date picker would scroll and the headers would not is non-trivial.

skel35:
Yes, please use the macOS solution here.

atsvirchkova:
Fixed the headers and removed unnecessary colors. Ready for review.

skel35 (changes requested):
The previously mentioned issues are fixed, but I've noticed one more - scrolling over the empty state textblocks isn't working:
5YYDAGWNsp

atsvirchkova:
Good finding. Thank you. Fixed now.

skel35 (approved):
Looks good 👍

PR #4363, opened 2020-08-28
Linux UI Rebranding

MartinBriza:
### 📒 Description
This PR is the UI part of the rebranding for the Linux app.

It also reverts the applicationName change from the common text PR because that moved the settings folder which could result in loss of user data.

### 🕶️ Types of changes

### 🔎 Review hints
UI should look partially rebranded - it's not a huge change for the Linux app because the redesign is coming soon-ish.


atsvirchkova (approved):
Looks good. Even though pink Start button still looks weird for me...

PR #4364, opened 2020-08-28
update_releases.sh: Change git username only in GA (actions)

MartinBriza:
### 📒 Description
This PR makes the `update_releases` script only change the GH username if the script is being ran from GitHub Actions.
It also seems to add a new line to the end of the file because some editors just don't like it missing.

### 🕶️ Types of changes

### 🔎 Review hints
Check if the syntax is right, run just the `prepare` function. Observe if your git username has changed for this repo. Put something in the `GITHUB_WORKFLOW` env variable and again observe if your git username has changed.
It should only change when `GITHUB_WORKFLOW` is not empty.


skel35 (approved):
Looks good, simple fix, makes perfect sense

PR #4372, opened 2020-09-01
Improve Editor stability by removing possible nil TE (mac)

AndrewVebster:
### 📒 Description
Add code guard in some places to reduce the possibility of crashes.
Time entry on Edit screen shouldn't be `nil` - it's a possible cause of the issue.

### 🕶️ Types of changes

### 🔎 Review hints
Please test the Edit screen. 
Especially description field - there may be a possibility of regressed issues, like time when it wasn't updating when switching between different TEs or running TE.


MartinBriza (approved):
Solution looks promising and from my testing, no regressions were added. Let`s see how well does this behave in the dev channel

PR #4373, opened 2020-09-01
Fix application menu entry in some distros

### LINKED ISSUE #4332: Latest deb (v7.5.247) isn't installing correctly
### 💻 Environment
Platform: Linux

### 📒 Description
I haven't dug deeper but when installing toggl via latest deb (v7.5.247) the desktop entry and binaries in path are gone. 
No idea when it stopped working but v7.5.226 seems to be working correctly.

MartinBriza:
### 📒 Description
Some distros/DEs are fussy about the desktop file not conforming to specs. That's why this PR removes the `TryExec` (unuseful for us now) and `Encoding` (deprecated) entries from it, making it work right everywhere.

### 🕶️ Types of changes

### 👫 Relationships
Closes #4332

### 🔎 Review hints
Just merging this is fine, it's a tiny fix and we don't ship the `TogglDesktop` binary in the main `bin` directory of any package anymore. `TogglDesktop.sh` is there instead and it sets all paths up.

### DIFF
--- a/src/ui/linux/TogglDesktop/dist/com.toggl.TogglDesktop.desktop
+++ b/src/ui/linux/TogglDesktop/dist/com.toggl.TogglDesktop.desktop
@@ -1,12 +1,10 @@
 [Desktop Entry]
-Encoding=UTF-8
 Version=1.0
 Type=Application
 Terminal=false
 Name=Toggl Desktop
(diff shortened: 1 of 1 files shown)


atsvirchkova (approved):
Look good!

PR #4376, opened 2020-09-03
Activity blocks and TEs placing

### LINKED ISSUE #4370: Add Activity and TEs visualization [Timeline UI]
### 💻 Environment

Platform: Windows

### 📒 Description
This is a sub-task of #4269 .

- Implement displaying activity blocks and popup with info on each activity block hover.
- Add basic Time Entry pills visualization with an algorithm for laying them out when TEs overlap
- Add create TE from gaps functionality.

atsvirchkova:
### 📒 Description
Implements displaying activity blocks and popup with info on each activity block hover.
Add basic Time Entry pills visualization with an algorithm for laying them out when TEs overlap.
Add creating TEs from gaps.

### 🕶️ Types of changes

### 👫 Relationships

 Closes #4370 

### 🔎 Review hints
What should be tested:
 - Activity blocks placing. The color is just static for now. The height depends on duration, but the min possible is 10 and max is 50.
- TEs placing including overlapping case. The description panel is implemented, but does not include all the info yet. It does not contain duration text, and also I assume, there must be also some tags, etc. I'll check it out later with MacOS and include in the next PR.
- Placing gaps between TEs must work correctly. Also it must be possible to create a TE from gap, but opening the edit TE panel is not part of this PR.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/Toggl.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/Toggl.cs
@@ -15,6 +15,54 @@ namespace TogglDesktop
 {
 public static partial class Toggl
 {
+    #region Data types
+
+    public class TimelineEventView
+    {
+        public string Title;
+        public string Filename;
+        public Int64 Duration;
+        public string DurationString;
+        public bool Header;
+        // references subevents
+        public List<TimelineEventView> SubEvents;
+
+        public TimelineEventView(TogglTimelineEventView eventView)
+        {
+            Title = eventView.Title;
+            Filename = eventView.Filename;
+            Duration = eventView.Duration;
+            DurationString = eventView.DurationString;
+            Header = eventView.Header;
+            SubEvents = marshalList<TimelineEventView, TogglTimelineEventView>(eventView.Event, n => n.Event, t => new TimelineEventView(t));
+        }
+    }
+
+    public class TimelineChunkView
+    {
+        public UInt64 Started;
+        public UInt64 Ended;
+        public string StartTimeString;
+        public string EndTimeString;
+        public List<TimelineEventView> Events;
+
+        public TimelineChunkView(TogglTimelineChunkView chunk)
+        {
+            Started = chunk.Started;
+            Ended = chunk.Ended;
+            StartTimeString = chunk.StartTimeString;
+            EndTimeString = chunk.EndTimeString;
+            Events = marshalList<TimelineEventView, TogglTimelineEventView>(chunk.FirstEvent, n => n.Next, t => new TimelineEventView(t));
+        }
+
+        public override string ToString()
+        {
+            return EndTimeString;
+        }
+    }
+
+    #endregion
+
     #region constants and static fields
 
     public const string Project = "project";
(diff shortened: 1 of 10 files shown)


atsvirchkova:
I've just discovered a thing: when you choose a date in the calendar for timeline UI and the TEs for this date are **not loaded**, then the lib callback sends empty list of TEs. As far as I can see there is no "Load more" button for mac OS and it dispays TEs for a very long period. I don't know how to fix it at the moment, I'll investigate it later.

skel35:
@atsvirchkova this logic is handled by the macOS UI part. Not sure why it was done this way, but it is fairly trivial to implement - just keep the oldest date for which the time entries are loaded and call `loadMore` when the selected date is less than that.

atsvirchkova:
I thought, that such a solution is not effective - user can request a date which is half a year ago for example. Maybe we can change the lib logic so that it would send the actual TEs on onTimline event, not an empty list?

skel35:
> I thought, that such a solution is not effective - user can request a date which is half a year ago for example.
practically that won't really be an issue, our app won't load time entries that are older than 60 days anyway.

>Maybe we can change the lib logic so that it would send the actual TEs on onTimline event, not an empty list?
That would be a more correct solution and you can implement it, but it seems much faster to implement the same solution that macOS app already has.

skel35 (changes requested):
I'm seeing various issues with the activity blocks:
1. App name header is not shown. Instead, the window title is just duplicated.
2. For some reason I'm not seeing any grouping by app name, every entry has just one item underneath. This means something is wrong.
3. when I hover over one activity block then move the cursor to another activity blocks, the tooltip stays for the previous one
4. If I click on the activity block itself the tooltip disappears and no mouse hovering can enable it again.

atsvirchkova:
Thank you, working on it. As for activities grouping: I'm just showing the events which come from the library. Maybe I don't understand quite well the sub-events concept - I thought that some kind of grouping is done on C++ side and a parent event can contain a list of sub-events. I'm now checking how it works on the lib side.

src/ui/windows/TogglDesktop/TogglDesktop/Toggl.cs
```diff
@@ -15,6 +15,54 @@ namespace TogglDesktop
...
+        public List<TimelineEventView> SubEvents;
+
+        public TimelineEventView(TogglTimelineEventView eventView)
+        {
+            Title = eventView.Title;
+            Filename = eventView.Filename;
+            Duration = eventView.Duration;
+            DurationString = eventView.DurationString;
+            Header = eventView.Header;
+            SubEvents = marshalList<TimelineEventView, TogglTimelineEventView>(eventView.Event, n => n.Event, t => new TimelineEventView(t));
```
skel35, line 37: instead of `n => n.Event` there should be `n => n.Next`. this fixes the issue with every event showing only one subevent

skel35:
```suggestion
            SubEvents = marshalList<TimelineEventView, TogglTimelineEventView>(eventView.Event, n => n.Next, t => new TimelineEventView(t));

```

atsvirchkova: Thank you, this saved me a lot of time 🙇‍♀️

atsvirchkova:
Fixed issues mentioned above. Ready for review again now.

skel35 (approved):
Looks good, the main functionality is working well. Related improvements will be filed as separate tickets.

Great work so far 👍

PR #4399, opened 2020-09-09
Adapt autocomplete dropdown for app size updates (mac)

### LINKED ISSUE #4395: Time entry suggestions field and dropdown width is too small
### 💻 Environment

Platform: macOS
OS Version: macOS 10.15.6
Toggl Version: 7.4.260

### 🐞 Actual behavior
A user has brought up that while the MacOS app is smaller, the time entry suggestions drop down is too small and unreadable, see this image for reference.

### 💯 Expected behavior
The dropdown should show limited yet readable information

### 🔨 Steps to reproduce

1.  Minimize the width of your app window
2. Begin typing a time entry to show suggestions

### 📦 Additional info

image

AndrewVebster:
### 📒 Description
Now autocomplete dropdown width does not depend on the description field width.
It has max width of 500pt. It stays at this width if app is fullscreen or just has large size.
If app window width is smaller than max width, dropdown size will take app window width as a max value (excluding some padding).
Dropdown height is also calculated differently and updates on layout change.

### 🕶️ Types of changes

### 👫 Relationships
Closes #4395 

### 🔎 Review hints
Test autocomplete dropdown in various app screen size conditions


skel35 (approved):
LGTM

PR #4400, opened 2020-09-09
Fix Bugsnag configuration to run in Release only (mac)

AndrewVebster:
### 📒 Description
Bugsnag configuration in Xcode was causing high load on my Macbook CPU - 100% after building project a few times.
Bugsnag script to upload dSYM file was duplicated and called twice - removed one.
Added `if` to check when it's a build where dSYM is generated (Release mode) so we don't have useless calls in Debug mode.

Checked that Bugsnag installation is done according to the documentation:
- https://docs.bugsnag.com/platforms/macos/
- https://docs.bugsnag.com/platforms/macos/symbolication-guide/

### 🕶️ Types of changes


skel35 (approved):
looks good

I initially thought I should test the Busnag error sending, but I've now realized that this change doesn't impact actual Bugsnag library or logic, only symbolic information sending, so not much that can be tested

PR #4403, opened 2020-09-10
Fix Touch Bar Start button state update for running TE (mac)

### LINKED ISSUE #4381: MacBook Touch Bar Start button is not updated when user start new time entry
### 💻 Environment
Platform: macOS
OS Version: macOS 10.15
Device: Macbook with Touch Bar
Toggl Version: 7.5.258

### 🐞 Actual behavior
On MacBook Touch Bar user always sees Start button, even when there is a running time entry.

### 💯 Expected behavior
Touch Bar should represent the current timer state - when running, the Start button should change to Stop. Plus a currently running time entry description should be visible near the Stop button.

### 🔨 Steps to reproduce
1. Get MacBook with Touch Bar
2. Open Toggl (no TE is running)
3. On MacBook Touch Bar, press on Start button
4. Check that TE is running now
5. See that on Touch Bar there is still Start button, instead of Stop

AndrewVebster:
### 📒 Description
Fixes bug when on MacBook Touch Bar user always saw Start button, even when there was a running time entry.

Root cause: after the design update the `startButton` was not using `state` property to update between start/stop states.

### 🕶️ Types of changes

### 👫 Relationships
Closes #4381

### DIFF
--- a/src/ui/osx/TogglDesktop/Features/Timer/TimerViewController.swift
+++ b/src/ui/osx/TogglDesktop/Features/Timer/TimerViewController.swift
@@ -114,7 +114,7 @@ class TimerViewController: NSViewController {
         viewModel.onIsRunning = { [unowned self] isRunning in
             self.updateStartButton(forRunningState: isRunning)
             NotificationCenter.default.post(name: NSNotification.Name(rawValue: kStartButtonStateChange),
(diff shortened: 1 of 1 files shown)


skel35 (approved):
tested on Xcode's touch bar, works fine

Issue #3461, opened 2019-10-24
Tag changes/deletions and continuing entries in Desktop


nicktoggl:
### 💻 Environment

Platform: macOS
OS Version: macOS 10.15
Toggl Version: 7.4.1023

### 🐞 Actual behavior
The user is deleting tags on our website, however when they continue an entry via the desktop app, the original tag is still present and recreates the tag, even after a closing of the app, a resync, or a log out and back in.

### 💯 Expected behavior
When a tag is deleted, it should be removed from entries and continuing an entry should prevent the tag from been created, particularly after a log out and back in or resync.

### 🔨 Steps to reproduce
1. Add a tag to any entry in the desktop app
2. Go to workspace settings and delete the tag
3. The tag should still be present in the desktop app for that entry, try resync/log out and back in, in my testing it is still present for the entry in the desktop app

### 📦 Additional info
[User case](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/2097915/conversations/24233415841)

skel35:
Related to #2513

dianetoggl:
[User report](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/3540765/conversations/24693567043)

dianetoggl:
Hey @skel35 any updated on this one? :)

IndrekV:
We have not made any progress with this. We will look into it in January.

dianetoggl:
Hey @IndrekV just bumping this one to check for any updates :)

dianetoggl:
Hey @IndrekV any updates on this one? 😊

IndrekV:
I'm sorry but there is no update. This issue is connected to the syncing rewrite that the sync team is working on and I don't think there will be a fix before the sync team finishes.

nidarasheed:
[Case reference](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/534741/conversations/26915731740)

dianetoggl:
[Case reference](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/3540765/conversations/106481300026053) User adding tags via API, tags don't show in desktop app, but show in webapp. Tags created in webapp show in desktop

dianetoggl:
[Case reference](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/3540765/conversations/106481303545721)

nicktoggl:
[...]

PR #4412, opened 2020-09-14
Fix errors when updating tags in the local database

### LINKED ISSUE #4366: User unable to save tags on macOS desktop app
### 💻 Environment
Platform: macOS/Windows/Linux
OS Version: Any
Toggl Version: 7.5.245

### 🐞 Actual behavior

Choosing a tag and clicking anywhere on the popup, or pressing enter doesn't allow the tag to save

### 💯 Expected behavior

Tags should be saved by pressing enter or clicking anywhere on the popup

### 🔨 Steps to reproduce

I haven't been able to reproduce it but I believe Diane has.

### 📦 Additional info

User shared a couple of videos showing this behavior, [case reference](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/2759100/conversations/106481301370114)

skel35:
### 📒 Description
Fix errors when updating tags in the local database

### 🕶️ Types of changes

### 🤯 List of changes
- Copy the tags string instead of using the error-prone `useRef` on a `static thread_local` string.

### 👫 Relationships
Closes #4366
Related to #3461 (not the original issue, but most of the support cases in the comments of that issue)

### 🔎 Review hints
Test that #4366 is fixed.
Test that there are no performance issues when there are many tags in the time entry.

### DIFF
--- a/src/database/database.cc
+++ b/src/database/database.cc
@@ -40,6 +40,7 @@ using Poco::Data::Keywords::useRef;
 using Poco::Data::Keywords::limit;
 using Poco::Data::Keywords::into;
 using Poco::Data::Keywords::now;
+using Poco::Data::Keywords::bind;
 
 Database::Database(const std::string &db_path)
     : session_(nullptr)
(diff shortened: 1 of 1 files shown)


MartinBriza (approved):
Looks good!

PR #4416, opened 2020-09-14
Fix key detection for media keys on Linux

### LINKED ISSUE #4415: Setting media keys as shortcuts takes over keyboard (letter) keys
### 💻 Environment

Platform: Linux
OS Version: Arch Linux
Toggl Version: 7.5.260

### 🐞 Actual behavior

I've tried setting the "start/stop tracking" shortcuts in Toggl Desktop to special keys, ie the media keys on my keyboard.
I've specifically tried setting the `XF86Launch6` and `XF86Launch7` events. 
When I set up these keys, toggl binds to the keypress of the keys <kbd>c</kbd> and <kbd>3</kbd>, respectively (keyboard layout US, variant Colemak, in case this is layout-dependent). This obviously makes it hard to type.
The settings window does display the correct keys.

Meanwhile, it does not listen to the media keys.

### 💯 Expected behavior

I expect Toggl to respect the shortcuts I set up.

### 🔨 Steps to reproduce

1.  Install Toggl Desktop
2.  Set up shortcuts to media keys
3.  Try to use the media keys to start time tracking: doesn't work.
3.  Try to type and notice you run into problems with some keys.

thomwiggers:
### 📒 Description

Add a key table to vendored library qxtglobalshortcut5 to look up media keys

### 🕶️ Types of changes
- **Bug fix**

### 🤯 List of changes

* Fix detection of media keys on Linux

### 👫 Relationships

Closes #4415

### 🔎 Review hints

See #4415

### DIFF
--- a/third_party/qxtglobalshortcut5/gui/qxtglobalshortcut_x11.cpp
+++ b/third_party/qxtglobalshortcut5/gui/qxtglobalshortcut_x11.cpp
@@ -39,6 +39,8 @@
 #endif
 #include <X11/Xlib.h>
 
+#include "xcbkeyboard.h"
+
 namespace {
 
 const QVector<quint32> maskModifiers = QVector<quint32>()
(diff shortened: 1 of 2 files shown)


MartinBriza:
Thank you for the report AND the fix, Thom! I just verified that it fixes a problem on my keyboard (Czech, standard QWERTZ layout) when nothing would happen if the shortcut is set to the Calculator button. I see the code is coming from another qxt version so it should be fairly complete, therefore I think nothing is preventing me from merging this. Thank you again!

PR #4448, opened 2020-09-23
Check if dbus-launch exists before running it (linux)

### LINKED ISSUE #3357: dbus-launch: command not found when running toggldesktop installed with snap
I installed toggldesktop via snap. One time with Gnome Software Manager, the other time by command `sudo snap install --edge toggldesktop`.

I cannot start the tool because of an error message.

### 💻 Environment
Fresh installation of Ubuntu 19.04 Budgie-Edition.

Platform: Linux
OS Version: Ubuntu Budgie 19.04
Toggl Version: 7.4.991

### 🐞 Actual behavior
```
$ toggldesktop 
/snap/toggldesktop/6/bin/TogglDesktop.sh: line 24: dbus-launch: command not found
```

### 💯 Expected behavior
The app should start.

### 🔨 Steps to reproduce
1. Install Ubuntu Budgie 19.04
2. Install Toggl Desktop by using the Gnome Software installer
3. Try to start the app

MartinBriza:
### 📒 Description
On some distros, there's no `dbus-launch` for some reason (and I don't really understand why we're using it anyway but it doesn't harm anything so whatever). This PR will not run it in case it's not present (which seems to be the case in Linux Mint)

### 🕶️ Types of changes

### 🤯 List of changes
- Fixed a crash when running without dbus-launch

### 👫 Relationships
Fixes #3357 (not just related to snaps)

### 🔎 Review hints
You can try renaming `/usr/bin/dbus-launch` to something else for a bit and see if it works right. You'll also maybe have to `export XDG_CURRENT_DESKTOP=XFCE` or something else (from the list in the same condition I modified in this PR) to observe the difference.

### DIFF
--- a/src/ui/linux/TogglDesktop.sh
+++ b/src/ui/linux/TogglDesktop.sh
@@ -16,7 +16,8 @@ fi
 # Xubuntu, i3 and Cinnamon tray icon fix
 XDG=$XDG_CURRENT_DESKTOP
 
-if [[ "$XDG" = "X-Cinnamon" || "$XDG" = "XFCE" || "$XDG" = "Pantheon" || "$XDG" = "i3" || "$XDG" = "LXDE" || "$XDG" = "MATE" || "$XDG" = "Budgie:GNOME" ]]; then
+if command -v dbus-launch &> /dev/null && \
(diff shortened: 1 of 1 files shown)


atsvirchkova (approved):
Works fine 👍

PR #4460, opened 2020-09-24
Fix bug with in-app message not showing on app start (mac)

### LINKED ISSUE #4458: In-app message view is not shown on macOS after app launch
### 💻 Environment
Platform: macOS
OS Version: Any
Toggl Version: 7.5.255

### 🐞 Actual behavior
In-app message is not shown because it is covered with the list or timeline screen.

### 💯 Expected behavior
In-app message always should be shown to the user.

### 🔨 Steps to reproduce
1. Setup in-app message to be shown
2. Launch the app
3. Message is not visible

### 📦 Additional info
Screenshot 2020-09-24 at 11 41 15

AndrewVebster:
### 📒 Description
In-app message is always added to the view hierarchy before Main VC. This results in a situation when the Main VC is covering in-app message view and users do not see it.
This fix makes Main VC always a root view - adds it to the view hierarchy always below other views.

### 🕶️ Types of changes

### 👫 Relationships
Closes #4458

### 🔎 Review hints
1. Setup the in-app message
1. Launch the app and see if the message is shown

### DIFF
--- a/src/ui/osx/TogglDesktop/Common/Extensions/NSView+Ext.swift
+++ b/src/ui/osx/TogglDesktop/Common/Extensions/NSView+Ext.swift
@@ -25,4 +25,9 @@ extension NSView {
             }
         }, context: &theView)
     }
+
+    @objc
+    func addSubviewToBack(_ view: NSView) {
+        addSubview(view, positioned: .below, relativeTo: nil)
+    }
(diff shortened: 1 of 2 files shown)


skel35 (approved):
works well 👍

PR #4462, opened 2020-09-24
Open edit view on TE click

### LINKED ISSUE #4410: Open Edit view by clicking the time entry in the Windows Timeline UI
### 💻 Environment
Platform: Windows

### ⛓️ Dependencies
Part of #4269
Depends on #4370 

### 📒 Acceptance Criteria
- Clicking time entry opens the Edit view for this time entry
- Selected time entry looks highlighted
- Designs: https://app.zeplin.io/project/5d8d0415858722028d30888d/screen/5e17108f5f511c81e08f887e

atsvirchkova:
### 📒 Description
Opens Edit view on time entry click in Timeline UI.
Opens Edit view on a gap between time entries click.

### 🕶️ Types of changes

### 👫 Relationships

 Closes #4410 

### 🔎 Review hints
There is a default behavior in ScrollViewer that it brings into view the item, that has received focus. As a result, if user clicks on a time entry which is only partly visible, the scrollviewer will jump to fit it into view. I'm not sure if it's OK.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
@@ -248,7 +248,6 @@ namespace TogglDesktop.ViewModels
                         Started = prevEnd.Value + 1,
                         Ended = entry.Started - 1
                     };
-                    gaps.Add(block);
                     if (block.Height > 10) // Don't display to small gaps not to obstruct the view
                         gaps.Add(block);
                 }
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
@@ -141,7 +141,7 @@ namespace TogglDesktop.ViewModels
             {
                 var startTime = Toggl.DateTimeFromUnix(entry.Started);
                 var height = ConvertTimeIntervalToHeight(startTime, Toggl.DateTimeFromUnix(entry.Ended));
-                var block = new TimeEntryBlock()
+                var block = new TimeEntryBlock(entry.GUID)
                 {
                     Height = height < 2 ? 2 : height,
                     VerticalOffset = ConvertTimeIntervalToHeight(new DateTime(startTime.Year, startTime.Month, startTime.Day), startTime),
@@ -319,15 +319,23 @@ namespace TogglDesktop.ViewModels
         public ulong Started { get; set; }
         public ulong Ended { get; set; }
         public ReactiveCommand<Unit, Unit> CreateTimeEntryFromBlock { get; }
+        public ReactiveCommand<Unit,Unit> OpenEditView { get; }
+        private string _timeEntryId;
 
-        public TimeEntryBlock()
+        public TimeEntryBlock(string timeEntryId)
         {
+            _timeEntryId = timeEntryId;
+            CreateTimeEntryFromBlock = ReactiveCommand.Create(() => AddNewTimeEntry());
+            OpenEditView = ReactiveCommand.Create(() => Toggl.Edit(_timeEntryId, false, Toggl.Description));
             CreateTimeEntryFromBlock = ReactiveCommand.Create(AddNewTimeEntry);
         }
 
+        public TimeEntryBlock() : this(null) { }
+
         private void AddNewTimeEntry()
         {
-            Toggl.CreateEmptyTimeEntry(Started, Ended);
+            _timeEntryId = Toggl.CreateEmptyTimeEntry(Started, Ended);
+            Toggl.Edit(_timeEntryId, false, Toggl.Description);
         }
     }
 }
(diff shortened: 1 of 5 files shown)


skel35:
>There is a default behavior in ScrollViewer that it brings into view the item, that has received focus. As a result, if user clicks on a time entry which is only partly visible, the scrollviewer will jump to fit it into view. I'm not sure if it's OK.

I think this makes total sense 👍

src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
```diff
@@ -271,15 +282,28 @@ public class TimeEntryBlock
...
+            });
         }
 
+        public TimeEntryBlock() : this(null) { }
+
         private void AddNewTimeEntry()
         {
-            Toggl.CreateEmptyTimeEntry(Started, Ended);
+            TimeEntryId = Toggl.CreateEmptyTimeEntry(Started, Ended);
+            Toggl.Edit(TimeEntryId, false, Toggl.Description);
```
skel35, line 306: I'd use `OpenEditView` command here directly, it's more readable and it will help in case the logic of `OpenEditView` changes.

src/ui/windows/TogglDesktop/TogglDesktop/ui/windows/MainWindow.xaml.cs
```diff
@@ -548,6 +548,8 @@ private void editPopupVisibleChanged(object sender, DependencyPropertyChangedEve
             if (editPopup.IsVisible == false && ReferenceEquals(this.activeView, this.timerEntryListView))
             {
                 this.timerEntryListView.Entries.DeselectCells();
+                //TODO: Find a better way to undo the TE highlighting on Edit view closing
+                timerEntryListView.Timeline.ViewModel.SelectedTEId = null;
```
skel35, line 552: how about extracting an extension method?
```suggestion
                timerEntryListView.DeselectTimeEntries();
```

skel35 (changes requested):
- it seems like the selection border doesn't fully match the left-most part of the TE block (the "pill")
image
- the TE is moving 1-2 px down when selected, not a big deal but can be annoying for someone
- closing the Edit view feels really slow for me (500ms-1sec), both in the Timeline and the List view. On older versions it generally feels faster in the List view (<300ms). Maybe we can run the profiler and see what's the bottleneck and if there are any low-hanging fruits here. E.g. this line
`this.WhenAnyValue(x => x.SelectedTEId, x => x.TimeEntryBlocks).Subscribe(_ =>
                TimeEntryBlocks?.ForEach(te => te.IsEditViewOpened = SelectedTEId == te.TimeEntryId));`
seems suspicious, but not sure if it's the bottleneck.

atsvirchkova:
- The issue with the rounded border near the TE is handled in the next PR here https://github.com/toggl-open-source/toggldesktop/pull/4491
- About the performance of edit view close: yes, it has definitely started to work slower as on each close event TEs are considered changed and we recalculate all the TEs and activity blocks layout, which is sad. The current logic is: we listen to time entry change event and invoke date re-selection to get the list of updated TEs and activity blocks from the lib. At first I tried to save time relaying on the data which comes with time entry changed event, but it may have contained grouped time entries, so this was not reliable. In the last commit I tried to run all the heavy logic in different thread, seems to work fine for me.

src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
```diff
@@ -239,6 +248,7 @@ private void GenerateGapTimeEntryBlocks(List<Toggl.TogglTimeEntryView> timeEntri
                         Started = prevEnd.Value + 1,
                         Ended = entry.Started - 1
                     };
+                    gaps.Add(block);
                     if (block.Height > 10) // Don't display to small gaps not to obstruct the view
                         gaps.Add(block);
```
skel35, line 253: are you adding the same gap twice?

skel35 (changes requested):
Can you please do some measurements regarding how much time the opening and closing of the Edit view take before and after this PR? E.g. with 100 TEs in the list.
So that we can confirm there's no significant performance downgrade.

src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
atsvirchkova: Looks like faulty merge conflicts resolving, will fix this.

atsvirchkova:
Ok, will do some time measurements for this. Thank you.

atsvirchkova:
Some measurements on my PC on Edit View closing:
before the whole Timeline UI (7.5.260): 125-160 ms
on this PR, when everything is done in main thread: 210-250 ms
on this PR using `Task.Run()` for TEs and activity blocks layout: 160-190 ms

skel35 (approved):
Not ideal, but the bottleneck is still related to the List view, so it can be tackled separately.

Approved 👍

PR #4470, opened 2020-09-25
Change default TE duration to 1 hour

### LINKED ISSUE #4445: Timeline UI - click should create 1 hour duration TEs by default
### 💻 Environment
Platform: macOS

### 📒 Description
When the user clicks to create TEs (not click-drags) the created TE should have a 1-hour duration by default, instead of 0 duration like now.

skel35:
### 📒 Description
Detect clicks and click-drags for tiny durations and create a TE with a default duration of 1 hour instead. This is consistent with the Calendar view implementation in the web app.

### 🕶️ Types of changes

### 👫 Relationships
Closes #4445 

### 🔎 Review hints
Check the code
Test out the UX

Not sure if it's a good way to detect clicks, please comment if you think you know a better way

### DIFF
--- a/src/ui/osx/TogglDesktop/Features/Timeline/TimelineDashboardViewController.swift
+++ b/src/ui/osx/TogglDesktop/Features/Timeline/TimelineDashboardViewController.swift
@@ -585,7 +585,7 @@ extension TimelineDashboardViewController: TimelineCollectionViewDelegate {
         }
 
         // Create
-        let defaultDuration: TimeInterval = 1
+        let defaultDuration: TimeInterval = 60 * 60 // 1 hour
         startNewTimeEntry(at: startTime, ended: startTime + defaultDuration)
 
         // Onboarding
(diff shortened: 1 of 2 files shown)


AndrewVebster:
Though it works, I'd prefer to handle clicks differently. 
E.g. in `TimelineCollectionView` use `func mouseUp(with event: NSEvent)`. To detect if this was a single click we can use `event.clickCount == 1`. But to make this work as expected we also should postpone calling `timelineDidStartDragging` on delegate before we know that this was not just a click but a drag event. Probably `func mouseDown()` should be refactored.

Side note or question about the UX. Some users are complaining that TE on the timeline is created with one click instead of a double-click (including me 🙄). E.g. you hover over TE -> see popup -> want to disable it and click outside accidentally -> unneeded TE is created. So I like a more native way, like the Calendar app on macOS.

skel35:
@AndrewVebster ready for review

AndrewVebster:
Found a bug.
STR: do one click on Timeline, without moving the mouse, so `mouseDragged()` is not called.

AR:
Screenshot 2020-10-13 at 10 59 56

Probably happens because `dragEntryStartTime` property is a computed property and equals `min(mouseDragStartTime, mouseDragEndTime)`. `mouseDragEndTime` at this point is 0.
Using `mouseDragStartTime` at line 85 should fix the issue.

AndrewVebster (approved):
Great 👍 No annoying etries with 0 duration 🙌 😄

PR #4471, opened 2020-09-25
Fix Linux autocomplete focus issues

### LINKED ISSUE #4446: Autocomplete keyboard navigation is inconsistent
### 💻 Environment
Platform: Linux
OS Version: All
Toggl Version: 7.5.279

### 🐞 Actual behavior
This is really hard to describe in full depth in writing but some of the issues we're facing with autocomplete on Linux are:
 * Tab behavior (for keyboard navigation between input fields) is inconsistent when the popup is/isn't open
 * The input text sometimes is cleared and sometimes isn't (for example: when you select a project in the timer, it stays and it shouldn't)
(and probably something else)

### 💯 Expected behavior
Consistent and intuitive behavior when entering both time entries and projects

### 📦 Additional info
I'm now working on rewriting the signals in the autocomplete widget so we can control everything about it.

MartinBriza:
### 📒 Description
This issue fixes the overall weird autocomplete behavior described partially in #4446 . It was really hard to pull this off because the way we're hacking around the combobox dropdown is probably not the right way (sorry!) and fixing one thing sometimes breaks something else (however I think it's really fine now).

### 🕶️ Types of changes

### 🤯 List of changes
- Autocomplete behavior is now more consistent

### 👫 Relationships
Closes #4446 

### 🔎 Review hints
Autocomplete dropdown is used in three places:
 * in the timer
 * in time entry editor for the description
 * in time entry editor for the project

### DIFF
--- a/src/ui/linux/TogglDesktop/autocompletecombobox.cpp
+++ b/src/ui/linux/TogglDesktop/autocompletecombobox.cpp
@@ -129,17 +129,15 @@ void AutocompleteComboBox::onDropdownSelected(AutocompleteView *item) {
             emit projectSelected(item->ProjectLabel, item->ProjectID, item->ProjectColor, item->TaskLabel, item->TaskID);
             emit billableChanged(item->Billable);
             emit tagsChanged(item->Tags);
-            emit timeEntrySelected(item->Text);
-            setCurrentText(item->Description);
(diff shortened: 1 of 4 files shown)


atsvirchkova (approved):
Autocomplete is working more consistent now 👍

PR #4477, opened 2020-09-29
Add a seconds tooltip to start and end in TE Editor (linux)

### LINKED ISSUE #3366: Tooltip which displays time up to seconds precision (Linux)
### 💻 Environment
Platform: Linux

### 📒 Description
When hovering over Start time/End time fields it would be nice to see values with more precision, up to seconds. Hovering over these fields currently does nothing anyway.

### ⭐️ Why do you want this?
As a user, I'd like to view more detailed information about my time entries.

MartinBriza:
### 📒 Description
This adds a seconds-precise tooltip to the started and ended fields in the time entry editor.

### 🕶️ Types of changes

### 🤯 List of changes
- Hovering over the time entry start and end in the editor will show more precise start and end times

### 👫 Relationships
Closes #3366

### 🔎 Review hints
Just hover over the fields mentioned. Before there was no tooltip.


atsvirchkova (approved):
Looks and works good 👍

PR #4483, opened 2020-09-30
Add running time entry information to system tray (linux)

### LINKED ISSUE #2827: Add running time entry info to tray icon tooltip
### 💻 Environment

Platform: Linux

### 📒 Description
Add running time entry info to tray icon tooltip

Format:
`Description - Task Project Client (duration) | Total today: {Today's Total Duration}`

### ⭐️ Why do you want this?
Allows seeing the current state of the timer by hovering tray icon. Avoids the need to always open app window to check tracked time.

A similar thing was added to Windows with #2739 and #2504

MartinBriza:
### 📒 Description
This PR adds running time entry information to the system tray on Linux. I opted to not implement total daily tracked time because the library doesn't expose it directly and implementing it in however way would either be a hack or it would possibly break current functionality on Windows (and probably on macOS as well).

### 🕶️ Types of changes

### 🤯 List of changes
Closes #2827

### 🔎 Review hints
In GNOME, you'll probably need the TopIcons Plus extension enabled to actually see the icon because GNOME developers hate the system tray (for real).

### DIFF
--- a/src/ui/linux/TogglDesktop/systemtray.cpp
+++ b/src/ui/linux/TogglDesktop/systemtray.cpp
@@ -7,13 +7,18 @@
 #include "./systemtray.h"
 #include "./mainwindowcontroller.h"
 
-SystemTray::SystemTray(MainWindowController *parent, QIcon defaultIcon) :
-    QSystemTrayIcon(parent),
-    notificationsPresent(true)
+SystemTray::SystemTray(MainWindowController *parent, QIcon defaultIcon)
+    : QSystemTrayIcon(parent)
+    , notificationsPresent(true)
 {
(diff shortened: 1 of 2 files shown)


atsvirchkova (changes requested):
I see a problem from the functional point of view: the tooltip doesn't display the TE duration correctly. It freezes the time on 3-4 sec and then never changes.
Screenshot_20201006_130413

MartinBriza:
Ah, you're right, how could've I not noticed that? I'll add a timer to update it periodically.

MartinBriza:
Fixed

atsvirchkova (approved):
Now it works fine. Good job 👍

PR #4489, opened 2020-10-02
Add a dialog with a link for browserless environments (linux)

### LINKED ISSUE #4465: Linux: Handle if user has no web browser set up
### 💻 Environment
Platform: Can't really think of any now, but a Linux DE/WM without proper XDG environment setup

### 📒 Description
Sometimes, users on Linux use homebrew environments like custom WMs without setting it up according to XDG (there are even projects that straight up ignore XDG for some reason, I'm looking at you MPV). 
That breaks clicking links in the app and Google login for those users.
I think showing a small window with the copiable URL and a button to put it into the clipboard will be enough. 

### ⭐️ Why do you want this?
To make links work for everybody.

MartinBriza:
### 📒 Description
Well, this PR adds a dialog that will theoretically open when the user has no browser set up. However, the thing is I really don't know how to properly test this because even Elementary (from the original user report) manages to open links just fine with current versions of TogglDesktop. Anyway, adding this won't (hopefully) hurt anything.

### 🕶️ Types of changes

### 🤯 List of changes
- Added a backup dialog with URL display for environments without a browser

### 👫 Relationships
Closes #4465 

### 🔎 Review hints
Don't know how to test this. I'll update the PR if I find a way.

### DIFF
--- a/src/ui/linux/TogglDesktop/CMakeLists.txt
+++ b/src/ui/linux/TogglDesktop/CMakeLists.txt
@@ -20,8 +20,10 @@ set(BINARY_SOURCE_FILES
     autocompletelistmodel.cpp
     autocompleteview.cpp
     clickablelabel.cpp
+    clipboarddialog.cpp
     colorpicker.cpp
     countryview.cpp
+    desktopservices.cpp
     errorviewcontroller.cpp
     feedbackdialog.cpp
     genericview.cpp
@@ -48,6 +50,7 @@ set(BINARY_SOURCE_FILES
 
     # it's better to list UI files
     aboutdialog.ui
+    clipboarddialog.ui
     colorpicker.ui
     errorviewcontroller.ui
     feedbackdialog.ui
--- a/src/ui/linux/TogglDesktop/aboutdialog.cpp
+++ b/src/ui/linux/TogglDesktop/aboutdialog.cpp
@@ -47,7 +47,7 @@ void AboutDialog::displayUpdate(const QString update_url) {
 
 void AboutDialog::on_updateButton_clicked() {
     qDebug() << "on_updateButton_clicked url=" << url;
-    QDesktopServices::openUrl(QUrl(url));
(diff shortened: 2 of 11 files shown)


MartinBriza:
Alright, I managed to reproduce and test it: 
image
but unless you want to set up a nested Ubuntu container with X forwarding, you'll probably have to trust me.

Aaanyway, if you really really want to try it out, here's how I did it:
### Linux
There's a way to test this out in Windows, you'll need to have SELinux in permissive mode (or off completely) because the context here is wrong and I don't know how to fix it.
First, run a container with bare bones Ubuntu 16.04 (`podman` is a 1:1 replacement for `docker` so that could work as well):
```
podman run -e DISPLAY=$DISPLAY -v /tmp/.X11-unix:/tmp/.X11-unix:rw --security-opt label=type:container_runtime_t -ti docker.io/library/ubuntu:16.04 bash
```
Then somehow get TogglDesktop inside the container, so from inside, run this for instance:
```
apt update
apt install ssh
scp <username>@<host_ip>:<path_to_toggldesktop.deb> .
# for example: scp mbriza@192.168.0.101:toggldesktop_7.0.0_amd64.deb .
``` 
Which will copy the `.deb` installer over from your host machine. Then install and run it:
```
apt install ./toggldesktop_7.0.0_amd64.deb
TogglDesktop.sh
```
Inside the UI, it's easiest to test by clicking the "Log in with Google" button which should prompt the dialog from the screenshot (because there's no internet browser inside the container).

### Windows
Surprisingly, it's also possible to test this out from Windows. The steps are very similar to ones from Linux, except instead of running a `podman` container, you'll have to:
 * Get Ubuntu from WSL
 * Run a X11 Server, like Xming
Now, if you REALLY REALLY REALLY want to know how to do that, please contact me directly. Thanks :D

atsvirchkova (approved):
Tested if it works well with browser, and all is good 👍  Didn't setup a browserless environment on my PC, let's assume that it also works fine 😃

PR #4491, opened 2020-10-02
Tooltips for time entries in the Windows Timeline UI

### LINKED ISSUE #4409: Tooltips for time entries in the Windows Timeline UI
### 💻 Environment
Platform: Windows

### ⛓️ Dependencies
Part of #4269
Depends on #4370 

### 📒 Acceptance Criteria
- There should be tooltips shown when hovering over the time entries
- Tooltips design: https://app.zeplin.io/project/5d8d0415858722028d30888d/screen/5e171080712d4083ac1a0bb0 https://app.zeplin.io/project/5d8d0415858722028d30888d/screen/5e171099b53d11940a76d9ab
- Empty description should be displayed as "No Description"

atsvirchkova:
### 📒 Description
Display tooltips if the mouse hovers over time entry.

### 🕶️ Types of changes

### 👫 Relationships

Closes #4409 

### 🔎 Review hints
Test if tooltip appears on mouse hover and that it's designed according to task requirements.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/Resources/DesignUpdate/Icons.xaml
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/Resources/DesignUpdate/Icons.xaml
@@ -92,4 +92,12 @@
     <Viewbox Stretch="Uniform" x:Key="Toggl.DatePickerRight">
         <Path Width="25" Height="25" Fill="{DynamicResource Toggl.DarkGrayBrush}" Data="M14.098 12L8.736 6.639l.528-.528L15.152 12l-5.888 5.889-.528-.528z"/>
     </Viewbox>
+    <Style x:Key="Toggl.TagIcon" TargetType="{x:Type Path}">
+        <Setter Property="Fill" Value="{DynamicResource Toggl.SecondaryTextBrush}"/>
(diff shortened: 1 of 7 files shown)


skel35 (changes requested):
1. Vertical position of the tooltip is weird - it's always in the bottom half; I'd expect it to be either centered or around the cursor position. Even though it works similarly for the activity tooltip - I think it's less of an issue over there because the activity block is always small.
Especially it looks weird when only the top of the TE block is visible - the tooltip is outside of the window. This doesn't happen with the activity block tooltip.
image
2. The filled circle with a project color is not visible on the tooltip
image

skel35:
I've tried out the tooltip with small TEs, and the vertical position makes sense in these scenarios.
The solution could be to vertically center the tooltip only if height > X px.

atsvirchkova:
Here is two options for the offset:
image
I think the first one is the best. What do you think @skel35 ?

skel35:
@atsvirchkova The second one seems better for me too, but there's still an issue with the popup being displayed outside the Timeline view:
image

atsvirchkova:
yes, I didn't think about the top 😔 I'm not sure how difficult it will be to identify the visible area. There is an option to use mouse placement like this:
image

skel35:
This jumping tooltip is really annoying, the previous solution looked better
ObOSR12XH0

atsvirchkova:
Can't understand how it's jumping for you, because it shouldn't 🤔 . For me it looks like this:
tooltip

atsvirchkova:
OK, I think I found the solution. The placement strategy is the initial one (to the right of TE), but now it makes sure always to place inside the main view area (at least10 px lower than the top, and 10 px upper than the bottom).
image

skel35 (approved):
The last solution looks pretty good.
Good job!

PR #4496, opened 2020-10-05
Set strict versions for macOS app dependencies

### LINKED ISSUE #4495: Set strict versions for macOS app dependencies
### 💻 Environment
Platform: macOS

### 📒 Description
Best practices for using external dependencies recommend to always set a specific version for every dependency. 
This is needed to have more control, know what we use, and to improve stability (dependencies won't be updated to a potentially dangerous version without our consent).

AndrewVebster:
### 📒 Description
This PR sets a specific version for every macOS app dependency.

CocoaPods dependency manager is updated to the latest stable.

SwiftLint is also updated to the latest stable to be the same as we use on our CI. `NotificationCenter.default.removeObserver(self)` removed from few files as a result of this update because now it's a linter violation.

### 🕶️ Types of changes
- Refactoring

### 👫 Relationships
Closes #4495


skel35 (approved):
the change makes sense

works, but I had to run `pod repo update` in addition to `make init_cocoapod`

PR #4514, opened 2020-10-07
Fix wrong delete usage in bugsnag-qt 

### LINKED ISSUE #4253: Linux app crashes on network disconnection
### 💻 Environment

Platform: 
OS Version: 
Toggl Version: 

### 🐞 Actual behavior

Linux app crashes when user is running in a task (timer started) and the network disconnects.
image

### 💯 Expected behavior
App should not crash

### 📦 Additional info

[Case reference](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/534741/conversations/106481300030915)

MartinBriza:
### 📒 Description
So, after much painful debugging and weird backtraces, I found the source of this issue. The `reply` pointer in bugsnag-qt shouldn't have been deleted directly but after it has been processed completely, in the event loop. This PR fixes this issue and a crash that has been associated with it:

https://doc.qt.io/qt-5/qnetworkaccessmanager.html#finished

### 🕶️ Types of changes

### 🤯 List of changes
- Fix a crash happening with DNS problems

### 👫 Relationships
Closes #4253 

### 🔎 Review hints
Reproducer is described in #4253

### DIFF
--- a/third_party/bugsnag-qt/bugsnag.h
+++ b/third_party/bugsnag-qt/bugsnag.h
@@ -357,7 +357,7 @@ class BUGSNAGQTSHARED_EXPORT Bugsnag : public QObject {
                         << ", status code " << statusCode
                            << ", content " << QString(b);
         }
-        delete reply;
+        reply->deleteLater();
     }
 };
 


skel35 (approved):
LGTM

PR #4522, opened 2020-10-12
Draw line dislaying current time

### LINKED ISSUE #4270: Display current time on the Timeline UI (Windows)
### 💻 Environment
Platform: Windows

### 📒 Description

Display current time on the Windows Timeline UI like it's already done on macOS:

image
image

atsvirchkova:
### 📒 Description
Draw line displaying current time.

### 🕶️ Types of changes

### 👫 Relationships

 Closes #4270 

### 🔎 Review hints
Line position is updated every minute.


AndrewVebster:
Found two bugs:
1. When zooming in-out the line is disappearing. Seems like it is visible only for one zoom level
2. Sometimes the "current time" line is overlapped with TE even when TE end time is the same as current
Screenshot 2020-10-12 at 19 11 22

atsvirchkova:
Fixed issue with zooming.
As for that line is drawn above the description: this happens because description has partly transparent background (and this aligns with the designs https://app.zeplin.io/project/5d8d0415858722028d30888d/screen/5e17107b9f02f0833a28d65e). It's possible to draw line behind the description only if the later one is non-transparent (and this would align with macOS design). I made the background non-transparent in the last commit, this is really better for the cur time line, but not sure if it's better for everything else.

AndrewVebster (approved):
Bug with zooming is fixed. All seems to work 👍

PR #4528, opened 2020-10-13
Display pullUserPreferences error

### LINKED ISSUE #4487: App does not show message that version is unsupported/teapotted
### 💻 Environment
Platform: macOS (should also test Windows & Linux)
OS Version: macOS Any
Toggl Version: 7.5.286

### 🐞 Actual behavior
The app does not react to `HTTP 418 I'm a teapot` response from the server - the error message that the app version is not supported is not shown to the user.

### 💯 Expected behavior
The error message is shown to a user when the app version is unsupported: "This version of the app is not supported any more. Please visit Toggl website to download a supported app."

### 🔨 Steps to reproduce
1. Change the app version to unsupported. E.g. on macOS it can be any `7.4.*`
2. Run the app
3. Check if the response is `HTTP 418`
4. See that error message is not shown

### 📦 Additional info
With `git bisect` I narrowed it down to commit 736f251
If `pullUserPreferences()` returns an error (and it will because version is teapotted) we just call `break`. Assigning `state` to `LEGACY` or calling `displayError()` before `break`will fix the problem. But I'm not sure if this is the right way to fix this issue.

atsvirchkova:
### 📒 Description
Display error if pullUserPreferences returns with error.

### 🕶️ Types of changes

### 👫 Relationships

 Closes #4487 

### 🔎 Review hints
In debug mode change app version to one, that is not supported (e. g. 7.1.211 for Windows). Then compile and check if shows error message "This version is not supported any more."

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -5017,8 +5017,10 @@ void Context::syncerActivityWrapper() {
                 logger.log("Syncer bootup, will attempt to determine which protocol to use");
                 //Do it here to know the type of syncing before syncerActivityWrapper() pulls the preferences
                 auto error = pullUserPreferences();
-                if (error != noError)
+                if (error != noError) {
+                    displayError(error);
                     break;
+                }
                 state = user_->AlphaFeatureSettings->IsSyncEnabled() ? state = BATCHED : LEGACY;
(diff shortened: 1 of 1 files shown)


AndrewVebster (approved):
Tested, all works as expected 👍

MartinBriza (approved):
Good simple patch, this was definitely necessary.

PR #4532, opened 2020-10-13
Resizing TEs to change the start time or duration

### LINKED ISSUE #4434: Resizing TEs to change the start time or duration
### 💻 Environment
Platform: Windows

Depends on #4370

### 📒 Description
Introduce the possibility to resize the TEs (from the top or the bottom) to change its duration or duration+start time.

atsvirchkova:
### 📒 Description
Introduce the possibility to resize the TEs (from the top or the bottom) to change its duration or duration+start time.

### 🕶️ Types of changes

### 👫 Relationships

Closes #4434 

### 🔎 Review hints

There is a problem for very small TEs: it's not possible to click on them to open edit view, because the resize handles take all its size. On the other hand it would be annoying if edit view opened when user tries to resize a TE. On MacOS it's done vice versa, it's not possible to resize very small TEs (at least I couldn't do that). Not sure which solution is better.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
@@ -353,8 +353,10 @@ namespace TogglDesktop.ViewModels
 
     public class TimeEntryBlock : ReactiveObject
     {
+        [Reactive]
         public double VerticalOffset { get; set; }
         public double HorizontalOffset { get; set; }
+        [Reactive]
         public double Height { get; set; }
         public string Color { get; set; }
         public bool ShowDescription { get; set; }
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
@@ -171,7 +171,7 @@ namespace TogglDesktop.ViewModels
             {
                 var startTime = Toggl.DateTimeFromUnix(entry.Started);
                 var height = ConvertTimeIntervalToHeight(startTime, Toggl.DateTimeFromUnix(entry.Ended), selectedScaleMode);
-                var block = new TimeEntryBlock(entry.GUID, ScaleModes[SelectedScaleMode])
+                var block = new TimeEntryBlock(entry.GUID, ScaleModes[selectedScaleMode])
                 {
                     Height = height < 2 ? 2 : height,
                     VerticalOffset = ConvertTimeIntervalToHeight(new DateTime(startTime.Year, startTime.Month, startTime.Day), startTime, selectedScaleMode),
@@ -185,8 +185,9 @@ namespace TogglDesktop.ViewModels
                     HasTag = !entry.Tags.IsNullOrEmpty(),
                     IsBillable = entry.Billable,
                     Duration = entry.DateDuration,
-                    StartEndCaption = entry.StartTimeString + " - " + entry.EndTimeString
-                };
+                    StartEndCaption = entry.StartTimeString + " - " + entry.EndTimeString,
+                    IsResizable = height >= 14 //Resize handles take 5 px each (10 in total)
+            };
                 if (entry.Started != entry.Ended)
                 {
                     timeStampsList.Add((TimeStampType.Start, block));
@@ -267,7 +268,7 @@ namespace TogglDesktop.ViewModels
                 if (prevEnd != null && entry.Started > prevEnd.Value)
                 {
                     var start = Toggl.DateTimeFromUnix(prevEnd.Value+1);
-                    var block = new TimeEntryBlock(ScaleModes[SelectedScaleMode])
+                    var block = new TimeEntryBlock(ScaleModes[selectedScaleMode])
                     {
                         Height = ConvertTimeIntervalToHeight(start, Toggl.DateTimeFromUnix(entry.Started - 1), selectedScaleMode),
                         VerticalOffset =
@@ -375,6 +376,8 @@ namespace TogglDesktop.ViewModels
 
         [Reactive]
         public bool IsEditViewOpened { get; set; }
+
+        public bool IsResizable { get; set; }
 
         private readonly double _hourHeight;
 
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
@@ -7,6 +7,7 @@ using System.Reactive.Linq;
 using System.Threading.Tasks;
 using ReactiveUI;
 using ReactiveUI.Fody.Helpers;
+using TogglDesktop.Resources;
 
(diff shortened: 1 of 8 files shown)


src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
```diff
@@ -365,20 +367,42 @@ public class TimeEntryBlock : ReactiveObject
         [Reactive]
         public bool IsEditViewOpened { get; set; }
 
-        public TimeEntryBlock(string timeEntryId)
+        private double _hourHeight;
```
skel35, line 370: It is a mutable field, but its value is not changing

src/ui/windows/TogglDesktop/TogglDesktop/ui/controls/TimelineTimeEntryBlock.xaml.cs
```diff
@@ -7,9 +9,41 @@ namespace TogglDesktop
...
+
+        private void OnThumbBottomDragDelta(object sender, DragDeltaEventArgs e)
+        {
+            if (ViewModel.Height + e.VerticalChange > 0)
+                ViewModel.Height += e.VerticalChange;
+        }
+
+        private void OnThumbTopDragDelta(object sender, DragDeltaEventArgs e)
+        {
+            if (ViewModel.Height + e.VerticalChange > 0)
```
skel35, line 32: Something is wrong with this method - you're checking whether Height + VerticalChange > 0, but then you're subtracting the VerticalChange.

src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
skel35, line 370: Zooming support?

skel35 (changes requested):
There are some issues with the code, so I'm not testing it yet.
As for this
> There is a problem for very small TEs: it's not possible to click on them to open edit view, because the resize handles take all its size.

I think it's wise to disable all resizing for very small TEs because it's impossible for the user to distinguish resizing the top of the TE from the bottom of the TE. Moreover, Edit View provides the same functionality and more, so it's better to restrict resizing than Edit View.

src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
atsvirchkova: Zooming support is not necessary here, as all the TimeEntryBlok's are re-created when zooming changes and there will be a new TimeEntryBlock with new hourHeight.

src/ui/windows/TogglDesktop/TogglDesktop/ui/controls/TimelineTimeEntryBlock.xaml.cs
atsvirchkova: You're right, thank you.

atsvirchkova:
Rebased on master and fixed issue with small TEs edit view. Now ready for review.

src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
```diff
@@ -374,20 +377,43 @@ public class TimeEntryBlock : ReactiveObject
...
+        {
+            var hours = 1.0 * height / _hourHeight;
+            var dateTime = date.AddHours(hours);
+            return Toggl.UnixFromDateTime(dateTime);
+        }
+
+        public void ChangeStartTime()
+        {
+            Toggl.SetTimeEntryStartTimeStamp(TimeEntryId,
+                ConvertOffsetToTime(VerticalOffset, Toggl.DateTimeFromUnix(Started).Date));
```
skel35, line 410: Have you considered having timestamps / timespans / datetime on the viewmodel instead of px-based properties?

skel35 (changes requested):
I have a small critical comment. Otherwise, I'm ok with merging this PR.
The feature still needs a lot of improvements, but this first step is good enough. I'll create some tickets for improvements today.

src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
```diff
@@ -185,8 +185,9 @@ private static List<TimeEntryBlock> ConvertTimeEntriesToBlocks(List<Toggl.TogglT
                     HasTag = !entry.Tags.IsNullOrEmpty(),
                     IsBillable = entry.Billable,
                     Duration = entry.DateDuration,
-                    StartEndCaption = entry.StartTimeString + " - " + entry.EndTimeString
-                };
+                    StartEndCaption = entry.StartTimeString + " - " + entry.EndTimeString,
+                    IsResizable = height >= 14 //Resize handles take 5 px each (10 in total)
```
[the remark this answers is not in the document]
atsvirchkova: At first I tried to handle this in xaml only, binding visibility to height and using a coverter. But height changes on drag event, and when it becomes less than 14 while user is dragging, the handles would suddenly disappear. I wanted to avoid this behavior.

src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
atsvirchkova: Yes, and actually there are both start/end timestamp and verticalOffset/height in the view model. At first I thought it would be nice to avoid converters and work only with px-based properties, but it ended up that time-based properties became necessary at some point. I can try to refactor this and place all the px-based logic in the converters.

src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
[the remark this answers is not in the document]
skel35: You can add an immutable Height property instead (call it e.g. HeightBeforeResizing, or call it Height and the other one HeightMutable/HeightResizable). Then handling it in XAML is easy.
It might be slightly more code, but if we continue to hardcode px in the viewmodel, changing the UI may become really hard.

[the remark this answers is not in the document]
atsvirchkova: If you're concerned about that it's hardcoded in ViewModel, not in view, then I think the right way to go will be to create a *.cs file which would contain all the constants (and ScaleModes dictionary can also go there). And both views and viewModels would reference that constans.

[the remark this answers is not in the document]
skel35: Decoupling, yay 😄 Sounds like a good idea 👍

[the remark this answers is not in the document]
atsvirchkova: Can do it in this PR then 😄

skel35 (approved):
Looks good.
Issues for future improvements: #4562, #4563

PR #4533, opened 2020-10-13
Upload release asset before updating appcast xml file

### LINKED ISSUE #4505: Sparkle Appcast xml file is updated before the release is uploaded to the server
### 💻 Environment
Platform: macOS, GitHub Actions
Toggl Version: 7.5.298

### 🐞 Actual behavior
When the release script is running there's a moment when the Sparkle Appcast XML file is updated with the link to new release, but the link is still invalid as the file wasn't uploaded to the server yet.
Usually, it will be invalid for only a few seconds while the release is uploading, but it's possible that the uploading fails for some reason, in that case the users could be having an invalid update

### 💯 Expected behavior
When Sparkle Appcast XML file is updated with the link to new release, the link should already be valid.

### 🔨 Steps to reproduce
Release script is running automatically on each push to master

### 📦 Additional info
https://github.com/toggl-open-source/toggldesktop/blob/master/.github/workflows/main.yml#L103-L111

skel35:
### 📒 Description
Upload macOS release asset before updating the appcast xml file.
No actual issues were observed, but the fix seems to make the workflow safer.

### 🕶️ Types of changes

### 👫 Relationships
Closes #4505

### 🔎 Review hints
Just check to see if the code makes sense

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -100,15 +100,15 @@ jobs:
       with:
         name: TogglDesktop.dmg
         path: "${{ env.installer }}"
+    - name: Upload
+      if: github.event_name == 'release'
+      run: |
+        ./dist/upload-github-release-asset.sh github_api_token=${{ secrets.GITHUB_TOKEN }} tag="$TAG_NAME" filename="$installer" renameto="$installer_name"
     - name: Update Appcast
       if: github.event_name == 'release'
       run: |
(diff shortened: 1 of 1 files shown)


AndrewVebster (approved):
LGTM 👍

skel35:
reverted in 005dcf11c431568c2a395e0e472534d5754195cf

Issue #4529, opened 2020-10-13
Windows Timeline UI small UI/UX issues


skel35:
### 💻 Environment
Platform: Windows

### 📒 Description

- [x] The timeline should be automatically scrolled down to the first time entry of the day. The top of the time entry should be visible at the top of the day.
- [ ] Activity block tooltip - there should be a total time near each app name.
- [ ] Non-overlapping TEs look as if they are overlapped (the body of the block, not the left-side). On the screenshot both end time of the red TE and start time of the yellow TE are 14:06:43
![image](https://user-images.githubusercontent.com/16451391/95854169-41af6500-0d5f-11eb-866b-7b1de0078f55.png)
- [ ] Scrollbar should be moved a few px to the left (as in List view) because it overlaps with window resize when trying to drag it with the mouse.
- [ ] `+` inside the gap buttons should be centered. 
![image](https://user-images.githubusercontent.com/16451391/95854670-ecc01e80-0d5f-11eb-80c2-7420eb4de460.png)
- [ ] `No activity was recorded yet...` should be centered.
![image](https://user-images.githubusercontent.com/16451391/95854702-f6e21d00-0d5f-11eb-8842-3f9415e5fba7.png)

Minor
- [ ] Add some additional way to close the Record activity Info tooltip.
  Ideas: 
  - ~close after a period of time (this needs to be tested + we might not want to close the tooltip when the info icon is still being hovered over).~
  - ~add a close button to the tooltip.~
  - UPD: Info icon can become a button and clicking it would open/close the popup

PR #4537, opened 2020-10-14
Set scroll to first time entry

atsvirchkova:
### 📒 Description
Scroll in Timeline is automatically scrolled down to the first time entry of the day.

### 🕶️ Types of changes

### 👫 Relationships

Partly implements issue #4529

### 🔎 Review hints
What to test:
- Scroll is set to cur time if the date is today (centered)
- Scroll is set to the first TE of the day for all other dates
- Scroll doesn't change if there are no TEs for selected date
- Scaling works like it did before

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
@@ -4,6 +4,7 @@ using System.IO;
 using System.Linq;
 using System.Reactive;
 using System.Reactive.Linq;
+using System.Reactive.Subjects;
 using System.Threading.Tasks;
 using ReactiveUI;
 using ReactiveUI.Fody.Helpers;
@@ -13,9 +14,6 @@ namespace TogglDesktop.ViewModels
     public class TimelineViewModel : ReactiveObject
     {
         private DateTime _lastDateLoaded;
-
-        private List<Toggl.TimelineChunkView> _timelineChunks;
-        private List<Toggl.TogglTimeEntryView> _timeEntries;
 
         public TimelineViewModel()
         {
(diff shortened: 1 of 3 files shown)


skel35 (changes requested):
The date on the date picker doesn't change when clicking the left/right arrows.

Overall, the code seems way too messy and is mixing the reactive approach with TPL too much, which is not a good idea. Later today I can share some specific ways to refactor this, if you don't mind.

atsvirchkova:
I guess, what could lead to that dates stopped changing on left/right arrows. Will fix this.
About mixing the reactive approach with TPL too much: I agree with you, and really wanted to make it more reactive. But I spent quite a decent amount of time yesterday trying to invent how to do this and failed. The problem is the C++ event. Assume you need to subscribe on selection date change, but you need to wait until the time entry blocks are generated. And these 2 events - selecting date and timeline blocks generation - are connected only by C++ event (on date change we call C++ function which triggers an event on which all the UI is created). And you cannot relay on TE blocks change, because they can be changed from another source - time entry change. So, if you have some ideas how to overcome this, I would be really grateful to know it.
Thanks)

skel35:
Updated the branch with the refactorings of date selection logic and using BehaviorSubject instead of the library events.
@atsvirchkova Please review/test the changes and I'll approve the merge.

src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
```diff
@@ -14,52 +14,72 @@ public class TimelineViewModel : ReactiveObject
...
             OpenTogglHelpUri = ReactiveCommand.Create(() =>
                 Toggl.OpenInBrowser("https://support.toggl.com/en/articles/3836325-toggl-desktop-for-windows"));
-            this.WhenAnyValue(x => x.SelectedDate).Subscribe(HandleSelectedDateChanged);
+
+            Toggl.TimelineSelectedDate.Subscribe(date => SelectedDate = date);
+
+            this.WhenAnyValue(x => x.SelectedDate)
+                .Where(date => date != Toggl.TimelineSelectedDate.Value)
+                .ObserveOn(RxApp.TaskpoolScheduler)
+                .Subscribe(date => Toggl.SetViewTimelineDay(Toggl.UnixFromDateTime(date)));
```
atsvirchkova, line 29: Why do we need this condition `.Where(date => date != Toggl.TimelineSelectedDate.Value)`? Anyway we synchronize these 2 dates one line above. If date `Toggl.TimelineSelectedDate` would emit next value because of TE change, not because of date change, then `SelectedDate` would just stay the same and nothing would happen.

src/ui/windows/TogglDesktop/TogglDesktop/ui/views/Timeline.xaml.cs
```diff
@@ -35,15 +42,29 @@ protected override void OnPropertyChanged(DependencyPropertyChangedEventArgs e)
...
                     .DisposeWith(_disposable);
-                ViewModel?.WhenAnyValue(x => x.SelectedDate).Where(date => date.Date == DateTime.Today)
-                    .Subscribe(_ => SetScrollToCurrentTime())
+                ViewModel?.WhenValueChanged(x => x.IsTodaySelected)
+                    .Where(x => x)
+                    .ObserveOn(RxApp.MainThreadScheduler)
+                    .Subscribe(_ => MainViewScroll.ScrollToVerticalOffset(ViewModel.CurrentTimeOffset - MainViewScroll.ActualHeight / 2))
+                    .DisposeWith(_disposable);
+                ViewModel?.WhenAnyValue(x => x.FirstTimeEntryOffset)
+                    .Where(_ => !ViewModel.IsTodaySelected)
```
atsvirchkova, line 54: Would be nice to use `.ObserveOn(RxApp.MainThreadScheduler)` also here and get rid of overcomplicated ViewModel getter.

atsvirchkova (commented):
Tested, no issues found. Thank you for good job, there is only one question from me below.

src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
skel35: Because in case `Toggl.TimelineSelectedDate` emits a value, different from `SelectedDate`, we would do an unnecessary call of `Toggl.SetViewTimelineDay()`. I don't know if such a situation is possible, but I think it's bad to rely on the fact that it's impossible as it's the logic external to the `TimelineViewModel`. The `Where` clause is fairly simple and fast anyway.

src/ui/windows/TogglDesktop/TogglDesktop/ui/views/Timeline.xaml.cs
skel35: 53516b1 looks good 👍

skel35 (approved):
Tested a bit for the last time, approved from my side ✔️

Issue #4539, opened 2020-10-15
QNetworkConfigurationManager and QNetworkConfiguration are deprecated


MartinBriza:
These classes were deprecated in Qt 5.15, we should eventually port away from them, probably to QNetworkSession.

### 💻 Environment
Platform: Linux
Toggl Version: 7.5.324

### 📦 Additional info
Related to #4538 but a bit harder to fix

PR #4541, opened 2020-10-15
Remove most of C++ build warnings 

### LINKED ISSUE #4538: C++ Compile time warnings
There are some warnings coming from the library (and a few from the Linux app):
image
I'd like to change this because it's really hard to see actual helpful warnings when I'm working on new stuff in the library

MartinBriza:
### 📒 Description
With this PR, we get rid of most of compiler warnings with msvc, clang and g++. For a more comprehensive list of what has been changed, you can check the individual commits, I did a separate one for each type of change.

The most interesting ones are:
 * Passing `std::string` into `copy_string` istead of using  `c_str ` (avoids one unnecessary copy)
 * Deprecated Qt method changes (without change in behavior)
 * Passing const references instead of values for vectors and strings (because of possibly improved performance)
 * There was actually a small bug with autotracker rules when they were not stored as lowercase (and an unused variable was left behind) 

### 🕶️ Types of changes

### 👫 Relationships
Closes #4538 

### 🔎 Review hints
Everything should work the same as before, except:
 * No library C++ build warnings should occur on macOS
 * Only QNetworkConfiguration related ones on Linux (#4539)  
 * There should be less of them on Windows (I don't really understand why some of them happen in there)


atsvirchkova (changes requested):
There have become a lot less warning than before, thank you for this job. But according to review hints and build logs there is something left, which should not be left:
```
/home/runner/work/toggldesktop/toggldesktop/src/color_convert.cc:157:23: warning: ‘b’ may be used uninitialized in this function [-Wmaybe-uninitialized]
  157 |     return { r , g, b };
      |                       ^
/home/runner/work/toggldesktop/toggldesktop/src/color_convert.cc:157:23: warning: ‘g’ may be used uninitialized in this function [-Wmaybe-uninitialized]
/home/runner/work/toggldesktop/toggldesktop/src/color_convert.cc:157:23: warning: ‘r’ may be used uninitialized in this function [-Wmaybe-uninitialized]
```
from linux build,
get_focused_window_mac.cc contains some warnings, that something is deprecated (macOS build)
```
/Users/runner/work/toggldesktop/toggldesktop/src/toggl_api.h:352:40: warning: this function declaration is not a prototype [-Wstrict-prototypes]
    typedef void (*TogglContinueSignIn)(
```
also macOS build.

MartinBriza:
Thanks for the review, Anna, I addressed the remark about the colors being initialized, however I'm convinced it was not necessary because the `switch` below has covered all 6 possible cases. To be 100% sure, I added an explicit initialization to those values, though.

As for this:
```
/Users/runner/work/toggldesktop/toggldesktop/src/toggl_api.h:352:40: warning: this function declaration is not a prototype [-Wstrict-prototypes]
    typedef void (*TogglContinueSignIn)(
```
The prototype is now: 
```c++
    typedef void (*TogglContinueSignIn)();
```
And a proper solution to this is changing it to:
```c++
    typedef void (*TogglContinueSignIn)(void);
```
There's no real difference between those two but the compiler likes one of them more for some reason. I'm not changing it simply because I suspect it would break `make csapi` and I don't have Windows ready to fix that at this moment. I'm opening an issue for this as a possible future improvement: #4561

atsvirchkova (approved):
Good job 👍  Merging it now.

PR #4542, opened 2020-10-15
Automatically fill in info from the project that was created from Timer

### LINKED ISSUE #4334: Automatically fill in project that was created from project dropdown from Timer UI
### 💻 Environment
Platform: macOS

### 📒 Description
It is possible to add a new project from project dropdown on a Timer. But currently when you do so it is not automatically set for the time entry you're creating (or running TE) - dropdown is just closed.
As a user I expect new project to be automatically set for current TE.

AndrewVebster:
### 📒 Description
Currently, when no TE is running, after creating a new Project from Timer dropdown (or shortcut) project was created but not assigned to the Timer. So after clicking on Start the TE started without the project. 
This is not the case with the running TE - a new project is automatically set for the running TE.

This PR fixes this. Now it automatically fills in info from the project that was created from the Timer.

### 🕶️ Types of changes

### 🤯 List of changes
Turns out this can be done without changing the library 😄 

### 👫 Relationships
Closes #4334

### 🔎 Review hints

Just test if project selection and creation is not broken.

### DIFF
--- a/src/ui/osx/TogglDesktop/Features/Timer/TimerViewController/TimerViewModel.swift
+++ b/src/ui/osx/TogglDesktop/Features/Timer/TimerViewController/TimerViewModel.swift
@@ -231,6 +231,21 @@ final class TimerViewModel: NSObject {
         actionsUsedBeforeStart.insert(TimerEditActionTypeProject)
     }
 
+    func didCreateNewProject(_ project: Project) {
+        if !isRunning {
+            // when TE is running, the project is set on Library level
+            clearProject()
+        }
+
+        timeEntry.projectGUID = project.guid
+        timeEntry.projectColor = project.colorHex
+        timeEntry.projectLabel = project.name
+        timeEntry.taskLabel = project.taskName
+        timeEntry.clientLabel = project.clientName
+
+        onProjectUpdated?(project)
+    }
+
     // MARK: - Private
 
     private func startTimeEntry() {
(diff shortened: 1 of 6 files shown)


src/ui/osx/TogglDesktop/Features/Timer/TimerViewController/TimerViewModel.swift
```diff
@@ -231,6 +231,21 @@ final class TimerViewModel: NSObject {
...
+            clearProject()
+        }
+
+        timeEntry.projectGUID = project.guid
+        timeEntry.projectColor = project.colorHex
+        timeEntry.projectLabel = project.name
+        timeEntry.taskLabel = project.taskName
+        timeEntry.clientLabel = project.clientName
+
+        onProjectUpdated?(Project(timeEntry: timeEntry))
```
skel35, line 246: just wondering, what's the difference here between using `project` and `Project(timeEntry: timeEntry))`?

skel35 (approved):
Looks great, we're slowly approaching the perfect timer UX 😄 

I've discovered a different bug related to project creation #4546.

src/ui/osx/TogglDesktop/Features/Timer/TimerViewController/TimerViewModel.swift
AndrewVebster: no difference 🙈 Updated 👍

PR #4544, opened 2020-10-15
Remove BatchUpdate* methods (lib)

### LINKED ISSUE #4543: BatchUpdate* methods are unused
$subj

MartinBriza:
### 📒 Description
As mentioned in this PR and the original issue, there's not much else to it. The methods were not used anywhere except in tests so I'm just removing them.

### 🕶️ Types of changes
- ** Cleanup **

### 👫 Relationships
Closes #4543 

### 🔎 Review hints
Well, since the code was not used anywhere except in tests, there should be nothing to test. Maybe if we hit a compiler bug or somethiing.

### DIFF
--- a/src/model/base_model.cc
+++ b/src/model/base_model.cc
@@ -88,45 +88,6 @@ bool BaseModel::userCannotAccessWorkspace(const error &err) const {
         kCannotAccessWorkspaceError));
 }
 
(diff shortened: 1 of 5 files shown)


skel35 (approved):
Looks good

PR #4547, opened 2020-10-16
Fix bug with project creation popup closing when creating client (mac)

### LINKED ISSUE #4546: Creating a client after using project timer shortcut closes the project creation popup
### 💻 Environment
Platform: macOS
Toggl Version: 7.5.326

### 🐞 Actual behavior
Project creation popup is closed, the project was not created

### 💯 Expected behavior
Project creation popup is not closed and the user is allowed to finalize the project creation

### 🔨 Steps to reproduce
1. Type `@` and the name of a new project
2. Click "Create project"
3. In the client field, type the name of a new client
4. Click "Create client"

AndrewVebster:
### 📒 Description
Currently, when creating a client after using the project timer shortcut (or dropdown) the project creation popup
is closing automatically but should not.
This happens because we're closing the project window when it loses focus. And because Client creation is a different window, moving focus to the "Create" button results in a closed project window.
This PR adds an additional check - we don't close the window if it has child windows.

### 🕶️ Types of changes

### 👫 Relationships
Closes #4546

### DIFF
--- a/src/ui/osx/TogglDesktop/Features/Timer/TimerViewController/TimerViewController.swift
+++ b/src/ui/osx/TogglDesktop/Features/Timer/TimerViewController/TimerViewController.swift
@@ -512,7 +512,8 @@ class TimerViewController: NSViewController {
             object: projectAutoCompleteWindow,
             queue: .main
         ) { [unowned self] _ in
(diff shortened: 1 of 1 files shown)


skel35 (approved):
Nice and concise 👍

PR #4552, opened 2020-10-16
Fix bug when autocomplete mouse selection wasn't working on macOS

### LINKED ISSUE #4548: Cannot select item from autocomplete with a mouse (mac)
### 💻 Environment

Platform: macOS
OS Version: Any
Toggl Version: 7.5.324

### 🐞 Actual behavior
Cannot select an item from the description autocomplete with a mouse. Keyboard works as expected.

### 💯 Expected behavior
Should be able to select autocomplete item with a mouse.

### 🔨 Steps to reproduce
1. Type anything into the description field so autocomplete opens
2. Click on any item with mouse
3. Dropdown is close but item is not selected (or random item is selected)

AndrewVebster:
### 📒 Description
Fixes the bug when user could not select item from the description autocomplete with a mouse.

### 🕶️ Types of changes

### 👫 Relationships
Closes #4548 

### 🔎 Review hints
Test autocomplete in different use cases. This PR can break stuff

### DIFF
--- a/src/ui/osx/TogglDesktop/Features/AutoComplete/GenericAutoComplete/Storage/TagStorage.swift
+++ b/src/ui/osx/TogglDesktop/Features/AutoComplete/GenericAutoComplete/Storage/TagStorage.swift
@@ -19,7 +19,6 @@ extension Notification.Name {
     // MARK: Variables
 
     private(set) var tags: [Tag] = []
-    private var newTags: [Tag] = []
 
     // MARK: Public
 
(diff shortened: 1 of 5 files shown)


skel35 (changes requested):
The tags list is empty after starting the app before the Edit view is open.
I'm not sure if it's related to this PR, but I can't reproduce it on `master`.

AndrewVebster:
Updated, please review.
The last two commits are related to tags and also have extended commit message to explain what was done

skel35 (changes requested):
The description field gets reset in 1-2 secs after selecting an autocomplete item
desc

AndrewVebster:
The issue above is also present in a stable version. But I hope I fixed it.

skel35 (approved):
Tested a bit more, works as expected 👍

PR #4560, opened 2020-10-19
Improve mouse selection on autocomplete list (mac)

### LINKED ISSUE #4559: Mouse selection in Autocomplete sometimes does not work when TE is running
### 💻 Environment
Platform: macOS
OS Version: Any
Toggl Version: 7.5.333

### 🐞 Actual behavior
Sometimes when selecting an item from the description autocomplete the item is not selected but autocomplete is closed.
This is not 100% reproducible. Happens only when there's running TE.

Autocomplete dropdown is closing right after mouse click - not waiting for a mouse-up event. Because of that the selection background is also not shown to the user. E.g. it's not possible to select items with mouse-down -> drag -> up.

### 💯 Expected behavior
Mouse selection should always work.

### 🔨 Steps to reproduce
1. Run TE
1. Open autocomplete
1. Select an item with the mouse click
1. If the item was selected correctly (all data is pre-filled in TE) clear the description text field and repeat from step 2

AndrewVebster:
### 📒 Description

**Fixed issues:**
- sometimes item from the list was not selected after mouse click because of a race condition
- blue selection background was not visible to the user when selecting with a mouse - dropdown was closing too quickly

**Root cause:**
Before this change, `AutoCompleteTableView` could become a first responder. This resulted in a situation when `TimerDescriptionFieldHandler` was closing autocomplete because the text field was losing focus. Sometimes the list was closed quicker than it could process the selection event.

### 🕶️ Types of changes

### 👫 Relationships

Closes #4559

### DIFF
--- a/src/ui/osx/TogglDesktop/Features/AutoComplete/LegacyAutoComplete/AutoCompleteInput.m
+++ b/src/ui/osx/TogglDesktop/Features/AutoComplete/LegacyAutoComplete/AutoCompleteInput.m
@@ -88,6 +88,7 @@ static CGFloat dropdownHorizontalPadding = 11;
 							  forIdentifier :@"AutoCompleteTableCell"];
 
 	[self.autocompleteTableView setDelegate:self];
+	[self.autocompleteTableView setRefusesFirstResponder:YES];
 
 	[self.autocompleteTableContainer setDocumentView:self.autocompleteTableView];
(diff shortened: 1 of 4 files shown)


skel35 (approved):
I like that the keyboard focus now comes back into the text field instead of getting lost.
And it's always nice to remove unnecessary code.
👍

PR #4565, opened 2020-10-19
Fix loss of precision in color convert (lib)

MartinBriza:
### 📒 Description
This fixes the color conversion issue introduced in 48f10d9355ca7741a68c00315f2124b8cb40306e


AndrewVebster (approved):
Tested, colors are fixed on macOS 👍

PR #4569, opened 2020-10-20
Update unit tests for TimerDescriptionFieldHandler

AndrewVebster:
### 📒 Description
Updated and fixed unit tests related to the description shortcuts.

### DIFF
--- a/src/ui/osx/TogglDesktop/TogglDesktopTests/TimerDescriptionFieldHandlerTests.swift
+++ b/src/ui/osx/TogglDesktop/TogglDesktopTests/TimerDescriptionFieldHandlerTests.swift
@@ -76,27 +76,27 @@ class TimerDescriptionFieldHandlerTests: XCTestCase {
 
     // MARK: - Text processing
 
-    func testControlTextDidChangeDoNotCrashIfNoEditor() {
+    func testControlTextCursorDidChangeDoNotCrashIfNoEditor() {
         window.makeFirstResponder(nil)
     }
 
-    func testControlTextDidChangeProjectShortcut() {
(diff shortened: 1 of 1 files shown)


src/ui/osx/TogglDesktop/TogglDesktopTests/TimerDescriptionFieldHandlerTests.swift
```diff
@@ -200,6 +200,22 @@ class TimerDescriptionFieldHandlerTests: XCTestCase {
         XCTAssertEqual(handler.state, State.autocompleteFilter("some @proj"))
     }
 
+    func testNotRecalculatingStateOnSameCursorPosition() {
```
skel35, line 203: I'm not sure I understand the value of this test.
The way I see it is:
Scenario: text field string value is changing, although the cursor position doesn't change.
Expected: The autocomplete is not recalculated.

Is there a user scenario where this is supposed to happen?
To me, it looks like it's testing an implementation detail, but maybe I'm missing something.

AndrewVebster: There is one use case:
The user removes all text -> Autocomplete shows all items -> Esc to hide autocomplete -> Left/right-arrow key -> 
Autocomplete will show on every left/right-arrow key press if we're not using such condition.

AndrewVebster: Could not come up with a better solution. Removing this test case for now

skel35 (approved):
Looks good 👍

PR #4572, opened 2020-10-20
Fix Timeline breaking after drag-creating TE with open Edit (mac)

### LINKED ISSUE #4570: Timeline UI breaks when creating 2 overlapped TEs with click-drag
### 💻 Environment
Platform: macOS
Toggl Version: 7.5.340

### 🐞 Actual behavior
Timeline UI becomes empty. The only way to get it to work again is to restart an app.

### 💯 Expected behavior
Timeline UI should still function correctly.

### 🔨 Steps to reproduce
1. Create a TE 10:00 - 10:30.
1. Click on this time entry to open Edit popover
1. Without closing the Edit popover, click the Timeline view at 10:45 and drag towards 10:15.

AndrewVebster:
### 📒 Description
Fixes the issue when Timeline becomes empty after specific steps that are described in the related issue.

The bug became possible after we made Edit popover animated. So it basically closes after some delay, during the drag-to-create event. 

Narrowed down the fix to this small one line removal. Not sure why this line even was there. If you have any thoughts about this, please tell me. It calls `app(context)->OpenTimeEntryList();` that results in a chain of other calls. It seems like it reloads all time entries and because this happens during drag-to-create event the Timeline breaks. 

Another possible fix would be to make Edit popover not animated again. But this solution has two problems:
- When Edit popover is opened and the user clicks on Timeline, app will close popover but also will create TE with default duration.
- It seems like it'll just hides the real issue till it'll occur again next time.

### 🕶️ Types of changes

### 👫 Relationships
Closes #4570

### 🔎 Review hints
Please test if Timeline functions well. Especially unique use cases (if you know any).

### DIFF
--- a/src/ui/osx/TogglDesktop/Features/Timeline/TimelineDashboardViewController.swift
+++ b/src/ui/osx/TogglDesktop/Features/Timeline/TimelineDashboardViewController.swift
@@ -79,7 +79,7 @@ final class TimelineDashboardViewController: NSViewController {
     }()
     private lazy var editorPopover: EditorPopover = {
         let popover = EditorPopover()
-        popover.animates = true
+        popover.animates = false
         popover.prepareViewController()
         popover.delegate = self
         return popover


skel35 (changes requested):
Simply had a look at `blame` of this line and found https://github.com/toggl-open-source/toggldesktop/pull/3854, which is reproducible on this branch.

AndrewVebster:
Went with another fix - Edit popover is not animated now.
But I think we will need to do it better someday. E.g. fix #3854 does not work for running TE.

skel35 (approved):
👍

skel35:
>fix #3854 does not work for running TE.

@AndrewVebster can you please open an issue?

AndrewVebster:
Opened the Issue for the bug with running TE - #4582

PR #4573, opened 2020-10-20
Use SHA-256 for signing Windows binaries

### LINKED ISSUE #4571: Use more secure hash algorithm for signing Windows binaries
### 💻 Environment
Platform: Windows

### 📒 Description
We're still using the SHA-1 hash algorithm for code signing on Windows.
We should update to a more secure and standard nowadays SHA-256 algorithm as all of the modern apps do.
SHA-256 is fully supported by Windows XP SP3+, so we do not need to sign additionally with SHA-1 as we only support Windows 7+.

skel35:
### 📒 Description
Use SHA-256 for signing Windows binaries and installers.

### 🕶️ Types of changes
- **Enhancement** (non-breaking change which fixes an issue)

### 🤯 List of changes
- [x] Change hash algorithm to SHA-256 for Windows binaries, installers, and uninstallers
- [x] Change the order in CI to sign installers before uploading artifacts. These are mutually exclusive (release/non-release) under normal circumstances, so it doesn't affect anything.

### 👫 Relationships
Closes #4571

### 🔎 Review hints

1. Take the installers from the artifacts of the commit https://github.com/toggl-open-source/toggldesktop/commit/34db28c9d8fba9dbd4461a10fbf95d2a5208100c which is simulating the release build.
2. Check that the installer is signed with SHA-256 (Properties->Digital Signatures).
3. Install the app.
4. Check that the binaries are signed with SHA-256.
5. Check that the app can be successfully run.
6. Test the same on Windows 7.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -439,18 +439,18 @@ jobs:
     - name: Create x86 installer
       run: |
         cd dist/windows
-        makensis /DCERT_PASSWORD=${{ secrets.CERT_WINDOWS_PASSWORD }} TogglDesktopInstaller-x86.nsi
-    - name: Upload artifacts to GitHub
-      if: github.event_name != 'release'
-      uses: actions/upload-artifact@v2
-      with:
-        name: TogglDesktopInstaller-x86.exe
-        path: dist\windows\TogglDesktopInstaller.exe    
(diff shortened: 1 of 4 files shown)


atsvirchkova (approved):
All works fine 👍 Tested on Windows 7 and Windows 10.

PR #4575, opened 2020-10-20
Fix Tab not working on description field (mac)

### LINKED ISSUE #4574: Tab key don't work on timer description field on macOS
### 💻 Environment

Platform: macOS
OS Version: Any
Toggl Version: 7.5.340

### 🐞 Actual behavior
Tab key doesn't switch focus from the description field to the duration field in some cases.

### 💯 Expected behavior
Pressing on Tab key should switch to the duration text field if no dropdowns are visible.

### 🔨 Steps to reproduce
1. Type some text into Timer description field
1. Switch focus to Duration field
1. Shift-Tab to switch focus back to description field
1. Right-arrow key
1. Tab key -> don't work
1. Tab key -> works

AndrewVebster:
### 📒 Description
Now description field in `autocompleteFilter` state will process keyboard events the same as in `descriptionUpdate` state if dropdown is not visible.

### 🕶️ Types of changes

### 👫 Relationships

Closes #4574


skel35 (approved):
works fine

PR #4579, opened 2020-10-21
Use home/end keys in both dropdown and editfield for autocomplete

### LINKED ISSUE #4512: "Home" and "End" keys don't act in a standard way
### 💻 Environment

Platform: Linux
OS Version: Ubuntu 20.04
Toggl Version: 7.4.528

### 🐞 Actual behavior
"Home" and "End" keys do not work in a standard way when pressed with the cursor in the "What are you doing" and "Projects" fields. If there are entries in the drop-down, "Home" and "End" go to the first and last entries, respectively. If the drop-down is empty, the keys do nothing.

### 💯 Expected behavior
"Home" and "'End" should take the cursor to the beginning and end of a text field, respectively.

### 🔨 Steps to reproduce

1. In Toggl Desktop, click the `What are you doing?` field and enter text
2. Test "Home" and "End" key behaviour
3. Press Start, then Stop
4. Click the new log entry
5. With the cursor in the field containing the entry's name, test "Home" and "End" key behaviour
6. Repeat 5 in the Projects field below

### 📦 Additional info

This appears similar to the behaviour reported in https://github.com/toggl-open-source/toggldesktop/issues/3384, but doesn't seem to have been resolved.

MartinBriza:
### 📒 Description
Changes the behavior of home and end keys in the autocomplete dropdowns to modify the dropdown (when it's visible) and the edit field (when the dropdown is hidden).

### 🕶️ Types of changes

### 🤯 List of changes
- The Home and End keys now work with autocomplete dropdown as well.

### 👫 Relationships
Closes #4512

### DIFF
--- a/src/ui/linux/TogglDesktop/autocompletecombobox.cpp
+++ b/src/ui/linux/TogglDesktop/autocompletecombobox.cpp
@@ -77,6 +77,13 @@ bool AutocompleteComboBox::eventFilter(QObject *o, QEvent *e) {
             else
                 listView->keyPressEvent(ke);
             return true;
+        case Qt::Key_Home:
+        case Qt::Key_End:
+            if (!listView->isVisible())
+                lineEdit()->event(e);
+            else
+                listView->event(e);
+            return true;
         default:
             QComboBox::keyPressEvent(ke);
             //lineEdit()->keyPressEvent(ke);


skel35 (approved):
Home/End keys work as expected

PR #4583, opened 2020-10-22
Update start/end and duration on popup while resizing

### LINKED ISSUE #4562: Show the user the expected start time/end time/duration when resizing the TE
### 💻 Environment
Platform: Windows

### 📒 Description
In the Timeline view, the user needs to see more feedback while resizing the TE.

### Proposed solution
Update the duration, start time, and end time in the existing time entry tooltip.

### Note
The solution on macOS is to show a separate popup. Updating the tooltip seems like an easier solution and achieves the same result here.

atsvirchkova:
### 📒 Description
Update start/end and duration on popup while resizing

### 🕶️ Types of changes

### 👫 Relationships

Closes #4562 

### 🔎 Review hints
Check that start/end and duration are updated correctly on popup while resizing in different scale modes.


skel35 (changes requested):
Bug:
STR
1. Click a gap to create a gap time entry

AR
1. A newly created time entry has a date of 1st January 1970

atsvirchkova:
Fixed. Ready for review again.

skel35 (approved):
works fine

PR #4584, opened 2020-10-22
Fix switching focus from empty tags selection search field (mac)

### LINKED ISSUE #4506: Switching focus from empty tags selection text field doesn't work
### 💻 Environment
Platform: macOS
Toggl Version: 7.5.289

### 🐞 Actual behavior
Focus is cleared and it's impossible to return to any focused element with Tab or Esc

### 💯 Expected behavior
Focus is moved to the next focusable element

### 🔨 Steps to reproduce
1. Open the tags selection dropdown (via mouse or keyboard). Text field ("Find tags") should be focused.
2. Do not enter anything in the text field.
3. Press Tab.

AndrewVebster:
### 📒 Description

Fixes the bug by returning `true` - meaning we say to the OS that the keypress was handled by us. This way OS won't switch focus automatically.
Additionally, pressing on the Tab when the "Create" button is hidden will close the dropdown.

### 🕶️ Types of changes

### 👫 Relationships
Closes #4506


skel35 (approved):
Looks good

PR #4589, opened 2020-10-23
Fix DatePicker date format in the Timeline view

### LINKED ISSUE #4588: Windows Timeline UI - Date format "backwards" for me in USA. Should be 10/22/2020 but it is listed at 22/10/2020.
### 💻 Environment

Platform: Windows
OS Version: Windows 10
Toggl Version: 7.5.348

### 🐞 Actual behavior

Date is backwards for us in the USA. See screenshot.

### 💯 Expected behavior

Should be 10/22/2020 OR could do something like Oct. 22, 2020

### 📦 Additional info

image

skel35:
### 📒 Description
Use the Windows' "Long" date format which is not confusing for users with various locales.
This is the same date format that the Edit view's DatePicker is using.

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Use the "Long" date format for Timeline view's DatePicker

### 👫 Relationships
Closes #4588

### 🔎 Review hints
Test Timeline view's DatePicker with different Windows locale settings.
Settings -> Time & Language -> Regional Formats OR Change data formats


atsvirchkova (approved):
Works good 👍

PR #4592, opened 2020-10-23
When showing the app, focus the timer description (if available) (linux)

### LINKED ISSUE #2261: Always focus description field in the main view
The description field (in the [main view](https://support.toggl.com/wp-content/uploads/2012/03/Screenshot_2017-01-02_19-00-57.png)) is currently only focused once - when toggl is launched. After using the show/hide shortcut, the description field is not focused any more on Linux. Please have a look on that.

MartinBriza:
### 📒 Description
This PR makes the app focus the Description in the Timer when re-showing the app - if time entry list was visible before hiding it and the timer is not running.

### 🕶️ Types of changes

### 👫 Relationships
Closes #2261

### 🔎 Review hints
Other hide/show workflows should not break.


skel35 (approved):
works well

PR #4594, opened 2020-10-23
Make it possible to close the preferences with Escape

### LINKED ISSUE #4140: Allow for closing of Preferences windows with ESC key
### 💻 Environment

Platform: Linux, Fedora 32, Flatpak install, 7.4.528

### 📒 Description
When opening the settings window, dependening on the Window Manager, no close button is displayed. Asking for the ability to add an ESC key option (similar to the About dialog) which closes the settings window.

### ⭐️ Why do you want this?
Using a keyboard-only Window Manager (like i3wm, Material Shell) and allowing for usage with keyboard-only commands.

MartinBriza:
### 📒 Description
Now using escape actually closes the preferences dialog. As a side note: wth is the code above what I changed.

### 🕶️ Types of changes

### 🤯 List of changes
- Closing preferences with the Escape key is now possible

### 👫 Relationships
Closes #4140

### DIFF
--- a/src/ui/linux/TogglDesktop/preferencesdialog.cpp
+++ b/src/ui/linux/TogglDesktop/preferencesdialog.cpp
@@ -300,6 +300,11 @@ void PreferencesDialog::keyPressEvent(QKeyEvent *event) {
             ui->continueStopButton->setText(keySequence);
         }
     }
+    else {
+        if (event->key() == Qt::Key_Escape) {
+            close();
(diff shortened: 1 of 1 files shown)


skel35 (approved):
>wth is the code above what I changed.

yeah, I wouldn't expect that much spaghetti for something as simple as shortcuts handling :smile: 
anyway, works well :+1:

PR #4596, opened 2020-10-23
In the tarball, contain everyting in a root folder

### LINKED ISSUE #4490: Distributed tarball expands in same directory as tarball resides in
### 💻 Environment
Tarball: toggldesktop_linux_7.5.260_x86_64.tar.gz

Platform: Linux
OS Version: Ubuntu 18.04
Toggl Version: 7.5.260

### 🐞 Actual behavior
When the tarball is expanded, it creates its content in the same directory as the tarball resides. This is contrary to norms and expectations of users of tarballs.

### 💯 Expected behavior
The tarball should create its own subdirectory and expand its contents into that one. An appropriate directory name for a tarball named `toggldesktop_linux_7.5.260_x86_64.tar.gz` might be `toggldesktop_linux_7.5.260_x86_64`, that is to say, exactly the same except leaving out the file ending. `.tar.gz`.

### 🔨 Steps to reproduce
1. Download tarball `toggldesktop_linux_7.5.260_x86_64.tar.gz`.
2. Examine local directory before expanding, noting what content exists aside from tarball.
3. Run command `tar zxf toggldesktop_linux_7.5.260_x86_64.tar.gz` in a terminal where the tarball resides.
4. Examine local directory again and observe that multiple items have been expanded with non-obvious names such as `setup.sh`, `share` and `bin`.

MartinBriza:
### 📒 Description
With our tarball, some users were complaining they contain everything right in the root directory, that means, if you ran `tar xvf toggldesktop.tar.gz` in your `$HOME` directory, you would end up with having the `bin`, `lib` and `share` folders right in your `$HOME` which is a bit annoying. This PR adds a directory called `toggldesktop` to the root of the tarball so if you unpack it, it will be neatly contained in one folder.

### 🕶️ Types of changes

### 🤯 List of changes
- The tarball now contains everyting in a `toggldesktop` folder in its root

### 👫 Relationships
Closes #4490 

### 🔎 Review hints
Unpack an old archive, you'll end up with 3 folders and 1 file.
Unpack the new archive (hopefully it'll be in the artifacts) and you'll end up with just 1 folder (which will contain 3 folders and 1 file).

### DIFF
--- a/dist/linux/package.sh
+++ b/dist/linux/package.sh
@@ -160,7 +160,9 @@ function package() {
 
     echo "=========== Packaging" >&2
     mv "$BINDIR/TogglDesktop.sh" "$BINDIR"/..
-    tar cvfz $reporoot/toggldesktop_$(uname -m).tar.gz * >/dev/null
+    cd ..
+    mv "$PREFIX" toggldesktop
+    tar cvfz $reporoot/toggldesktop_$(uname -m).tar.gz toggldesktop >/dev/null
 
     popd >/dev/null
     echo "=========== Result is: $reporoot/toggldesktop_$(uname -m).tar.gz ==========="


skel35 (approved):
looks good

PR #4602, opened 2020-10-27
Fix Tab focus change from description field on Edit popup (mac)

### LINKED ISSUE #4601: Tab key does not switch focus from description field on Edit popup
### 💻 Environment

Platform: macOS
OS Version: Any
Toggl Version: 7.5.352

### 🐞 Actual behavior

Pressing TAB when focused on the Description field on the Edit popup does nothing.

### 💯 Expected behavior

Pressing TAB when focused on the Description field should switch focus to the Project field.

### 🔨 Steps to reproduce

1. Open Edit for any TE
2. Select description field
3. Press TAB

### 📦 Additional info

Caused by fix in PR #4584

AndrewVebster:
### 📒 Description

Returning true from keyboard handler to prevent switching focus was a bad idea because it disabled Tab key for description autocomplete on Edit popup. "Create" button is always hidden there.
This PR fixes this by allowing focus to flow but closing autocomplete when needed. Additionally, table view now cannot become first responder (the reason why previously focus was lost).

### 🕶️ Types of changes

### 👫 Relationships

Closes #4601 

### 🔎 Review hints

Test if Tab works as expected in different places where autocomplete is used.

### DIFF
--- a/src/ui/osx/TogglDesktop/Features/AutoComplete/GenericAutoComplete/AutoCompleteView.swift
+++ b/src/ui/osx/TogglDesktop/Features/AutoComplete/GenericAutoComplete/AutoCompleteView.swift
@@ -79,7 +79,12 @@ final class AutoCompleteView: NSView {
 
     // MARK: OUTLET
 
-    @IBOutlet weak var tableView: KeyboardTableView!
+    @IBOutlet weak var tableView: KeyboardTableView! {
+        didSet {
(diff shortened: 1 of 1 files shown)


skel35 (approved):
Tested tabbing and autocompleting in various places (Timer, Edit popup in List view and Timeline view, description, project, tag fields, project creation). Works well.

PR #4603, opened 2020-10-27
Count only time that fits the chunk for every event

### LINKED ISSUE #3099: Apps get tracked twice on Windows [Timeline]
### 💻 Environment
Platform: Windows
OS Version: Windows 10.0.17763
Toggl Version: 7.4.451

### 🐞 Actual behavior
Looking at my recorded timeline data, I can see that some 15-minute intervals contain more than 15 minutes of tracked time, approx. up to 30 min. When looking at entries in detail you can guess that there are duplicate entries, with a few seconds of difference in duration. 
image

### 💯 Expected behavior
Each 15-minute interval should not contain more than 15 minutes of tracked data. There should be no duplicates.

### 🔨 Steps to reproduce
Check "Record timeline".
Happens occasionally and, I think, on both x86 and x64 versions (though I might've gotten auto-updated from x64 to x86 without noticing).

I thought it might be happening due to launching the app both installed one and from IDE many times during the day, but I've noticed it happened even when I don't have any development environment open.

atsvirchkova:
### 📒 Description
Count only time that fits the chunk for every event.

### 🕶️ Types of changes

### 👫 Relationships

Closes #3099

### 🔎 Review hints
There should be no chunks whose activities sum up to over 15 mins.

### DIFF
--- a/src/gui.cc
+++ b/src/gui.cc
@@ -482,7 +482,7 @@ void GUI::DisplayTimeline(const bool open,
     }
 
     auto get_duration = [](time_t event_start, time_t event_end, time_t chunk_start, time_t chunk_end) {
-        return std::min(chunk_end, event_end) - std::max(chunk_start, event_start);
+        return (std::min)(chunk_end, event_end) - (std::max)(chunk_start, event_start);
     };
     // Get activity
     while (datetime.year() == TimelineDateAt().year()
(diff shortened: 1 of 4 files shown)


skel35 (approved):
tested by trying to track >30 mins long chunk, worked perfectly.
code looks reasonable.
great job!

Issue #3917, opened 2020-03-23
Show suggestion also when there is a running entry (Windows only)


skel35:
Enable only for Windows, but the implementation is in the library.

skel35:
Implemented in #3976

PR #4607, opened 2020-10-27
Start autotracker only if a window has been in focus for over 5 seconds

### LINKED ISSUE #4605: Autotracker prompt occurs too often
### 💻 Environment
Platform: Windows (Lib)

### 📒 Description
After #3917 the users of the Autotracker features have started seeing a lot more notifications.
E.g., if the user has set up `Notion` and `Slack` as autotracker terms for 2 different projects, then switching to Slack for a few seconds while doing Notion work always ends up in an autotracker notification, which can be annoying.
Some users have also complained about regular autotracker notifications happening too often.

Proposed solution: show autotracker notification only after a certain window has been in focus for 5+ seconds. The threshold of 5 seconds can be increased in the future depending on user feedback.

skel35:
### 📒 Description
Start autotracker only if a window has been in focus for over 5 seconds.

### 🕶️ Types of changes

### 🤯 List of changes
- [x] Start autotracker only if a window has been in focus for over 5 seconds
- [x] Fix idleness condition for autotracker - current value should be checked instead of the previous value

### 👫 Relationships
Closes #4605

### 🔎 Review hints
Set up a few keywords+projects, and try to use the autotracker.

The autotracker should pop up within 5 seconds of a window becoming active.
The autotracker should also work even if the window title changes.

- Test on Windows (as it has #3917 enabled, it covers more use cases).
- Test on one other platform

### DIFF
--- a/src/const.h
+++ b/src/const.h
@@ -20,7 +20,7 @@
 #define kTimerStartInterval 10
 #define kTimelineSecondsToKeep 604800
 #define kWindowFocusThresholdSeconds 10
-#define kAutotrackerThresholdSeconds 10
+#define kAutotrackerThresholdSeconds 5
 #define kBetaChannelPercentage 25
 #define kTimelineChunkSeconds 900
 #define kEnterpriseInstall false
--- a/src/window_change_recorder.cc
+++ b/src/window_change_recorder.cc
@@ -65,13 +65,19 @@ void WindowChangeRecorder::inspectFocusedWindow() {
         }
     }
 
+    idle = idle || getIsLocked() || getIsSleeping();
+
     time_t now;
     time(&now);
 
     time_t time_delta = now - last_event_started_at_;
 
+    if (last_filename_ != filename) {
+        current_app_started_at_ = now;
+    }
+
     if (last_event_started_at_ > 0) {
-        if (!last_idle_ && last_autotracker_title_ != title) {
+        if (!idle && last_autotracker_title_ != title && now - current_app_started_at_ >= kAutotrackerThresholdSeconds) {
             // Notify that the timeline event has started
             // we'll use this in auto tracking
             last_autotracker_title_ = title;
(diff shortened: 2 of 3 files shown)


robsonsobral:
May I say 5 seconds is too short? I was only reading a message!

(Anyway, I think is better to follow the same interval already in use)

skel35:
@robsonsobral 
we'll consider increasing the interval and/or making it customizable.
the same interval is in use for a related but different feature.
the main purpose of this change is to introduce the time interval delay for the entire autotracker feature.
thanks for your input.

AndrewVebster:
I don't know if this is expected so will leave it up to you.

OS: macOS

STR:
1. Open Safari with one tab set to https://github.com and others could be any
1. Select any tab except GitHub
1. Setup autotracker to search for "github"
1. Return back to Safari (not GitHub tab)
1. Wait >5 sec
1. Select to GitHub tab

AC:
Notification is shown right away and not after 5 sec.

This case is true for any app that utilizes tabs. Also tested on Finder.

AndrewVebster (approved):
Tested on Windows and macOS. Delay in 5 seconds works 👍

skel35:
> I don't know if this is expected so will leave it up to you.
> 
> OS: macOS
> 
> STR:
> 
> 1. Open Safari with one tab set to https://github.com and others could be any
> 2. Select any tab except GitHub
> 3. Setup autotracker to search for "github"
> 4. Return back to Safari (not GitHub tab)
> 5. Wait >5 sec
> 6. Select to GitHub tab
> 
> AC:
> Notification is shown right away and not after 5 sec.
> 
> This case is true for any app that utilizes tabs. Also tested on Finder.

The initial idea was to wait N seconds after a certain window becomes focused. I was thinking of a use case where the window title is changing frequently (e.g. contains a timer or some other dynamic info).
But now I think a more correct behavior would be to wait N seconds after a certain rule was matched.
I'll fix the PR.

skel35:
I've decided to merge this one first not to delay the release, as this PR ends up in users getting strictly fewer notifications than before. Implementing a more correct solution requires changing a lot of the logic and structure of the code.
I've created a separate issue to implement a proper solution #4624.

PR #4613, opened 2020-10-28
Add logging on/off parameter to TogglClient

### LINKED ISSUE #4578: Exclude analytics requests details from the log files
### 💻 Environment

Platform: All Platforms

### 📒 Description
All requests to Google analytics are now logged, which immediately fills in the log files. We don't want to have them in logs as it makes hard to extract any other valuable information from there.

atsvirchkova:
### 📒 Description
At the moment we log all http requests, and this logging is built in `makeHttpRequest` method. These changes add a boolean parameter which indicates if logging is needed. 

### 🕶️ Types of changes

### 🤯 List of changes
analytics.cc
http_client.h
http_client.cc

### 👫 Relationships

 Closes #4578 

### 🔎 Review hints
How to test: 
- Run the app in release mode.
- Open the log files.
- Check that there is not `google.analytics` requests there.

### DIFF
--- a/src/analytics.cc
+++ b/src/analytics.cc
@@ -158,7 +158,7 @@ void GoogleAnalyticsEvent::runTask() {
     req.host = "https://ssl.google-analytics.com";
     req.relative_url = relativeURL();
 
-    HTTPResponse resp = TogglClient::GetInstance().silentGet(req, false);
+    HTTPResponse resp = TogglClient::GetInstance().silentGet(req);
     if (resp.err != noError) {
         Logger("Analytics").error(resp.err);
         return;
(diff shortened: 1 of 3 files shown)


skel35 (changes requested):
This works fine, but let's improve the PR even more.
All the `silentGet` calls are the not-so-important calls (not related to syncing or the business logic), so we can remove logging from all of them. By doing that we would avoid an unnecessary boolean parameter in the `silentGet` method.

atsvirchkova:
@skel35 OK, that sounds reasonable. I'll remove the parameter from `silentGet`, but it will still remain in `request` and `makeHttpRequest` methods.

atsvirchkova:
Fixed, now ready for review again.

skel35 (approved):
logs are now cleaner for sure 👍 
looks good

PR #4618, opened 2020-10-29
Fix crash when onboarding is shown before window is loaded (mac)

### LINKED ISSUE #4617: Mac app crashes after launch when trying to present onboarding
### 💻 Environment

Platform: macOS
OS Version: Any
Toggl Version: 7.5.347

### 🐞 Actual behavior

App crashes when trying to show onboarding right after the app launch.

### 💯 Expected behavior

App should not crash

### 🔨 Steps to reproduce

This is hard to reproduce but the possible case is next:
1. App starts
1. Onboarding service starts
1. App creates the main window
1. Onboarding service wants to show the popover when the window is created but not yet visible

### 📦 Additional info

[Bugsnag](https://app.bugsnag.com/toggldesktop/toggldesktop/errors/5f99a02fffa2860017b055b9?filters[event.since][0][type]=eq&filters[event.since][0][value]=30d&filters[error.status][0][type]=eq&filters[error.status][0][value]=open&filters[search][0][type]=eq&filters[search][0][value]=onboarding)

AndrewVebster:
### 📒 Description
Setting `positioningRect` when the main `NSWindow` is not yet visible may crash the app with `NSInternalInconsistencyException` "window must exist".

To fix this I moved this line to the delegate method, so it'll be called only after the popover show method executes successfully. This means that onboarding won't show in such edge cases but also won't crash the app.

I think the actual root cause of this issue is that onboarding is shown after a constant delay (1 sec) after the app start. See `onboarding_service.cpp` line65:
```
// If it's the call from the launch
// Check condition for onboarding after 1 seconds, after the app is loading successfully
  if (isAtLaunchTime) {
      timer.add(seconds(1), [&](CppTime::timer_id id) {
        ...
```
But fixing this isn't trivial, so propose to merge this quick fix first.

### 🕶️ Types of changes

### 👫 Relationships

Closes #4617 

### 🔎 Review hints
The crash is very hard to reproduce but I was able to simulate it.
In `onboarding_service.cpp` comment lines 427 and 432 to force shortcuts onboarding to start every time on app launch.

In `OnboardingViewController` replace:
```
popover.show(relativeTo: presentView.bounds, of: presentView, preferredEdge: payload.preferEdges)
```
with:
```
let window = NSWindow()
let tmpView = NSView()
window.contentView = tmpView
popover.show(relativeTo: tmpView.bounds, of: tmpView, preferredEdge: payload.preferEdges)
```

### DIFF
--- a/src/ui/osx/TogglDesktop/Features/Onboarding/OnboardingViewController.swift
+++ b/src/ui/osx/TogglDesktop/Features/Onboarding/OnboardingViewController.swift
@@ -87,7 +87,6 @@ final class OnboardingViewController: NSViewController {
         preparePopoverContentView(with: payload)
 
         popover.show(relativeTo: presentView.bounds, of: presentView, preferredEdge: payload.preferEdges)
-        popover.positioningRect = payload.positioningRect(payload.hint, popover.positioningRect)
     }
 
     /// Dismiss all onboard view
@@ -160,6 +159,11 @@ extension OnboardingViewController: NSPopoverDelegate {
 
     func popoverWillClose(_ notification: Notification) {
         delegate?.onboardingViewControllerDidClose()
+    }
+
+    func popoverWillShow(_ notification: Notification) {
+        guard let payload = payload else { return }
+        popover.positioningRect = payload.positioningRect(payload.hint, popover.positioningRect)
     }
 }
 


skel35 (approved):
Doesn't crash anymore when simulating the crash situation, and works fine under regular conditions.
👍

PR #4637, opened 2020-11-02
Prevent possible crash when emptyActivityLbl is not yet loaded (mac)

### LINKED ISSUE #4636: Unknown crash on Timeline (Bugsnag)
### 💻 Environment

Platform: macOS
OS Version: Any
Toggl Version: 7.5.298+

### 🐞 Actual behavior

App crashed - [Bugsnag](https://app.bugsnag.com/toggldesktop/toggldesktop/errors/5f991d32e4ba5b0017e967d8?i=em&m=ws)

### 💯 Expected behavior

No crash

### 🔨 Steps to reproduce

Could not reproduce

AndrewVebster:
### 📒 Description
This PR possibly 🤞 fixes the `EXC_BAD_INSTRUCTION` crash reported on [Bugsnug](https://app.bugsnag.com/toggldesktop/toggldesktop/errors/5f991d32e4ba5b0017e967d8?i=em&m=ws&filters[event.since][0]=2020-09-03T16%3A43%3A38.658Z).
This crash seems rare and almost impossible to reproduce. So this PR is like a prevention of the possible issue:
We assume that the crash happened because `emptyActivityLbl` view was not yet loaded. The change that was done here is totally unharmful, so after the release, we can check Bugsnag again to see if the bug was really fixed or we need to investigate again.

### 🕶️ Types of changes

### 👫 Relationships

Closes #4636

### DIFF
--- a/src/ui/osx/TogglDesktop/Features/Timeline/TimelineDashboardViewController.swift
+++ b/src/ui/osx/TogglDesktop/Features/Timeline/TimelineDashboardViewController.swift
@@ -21,7 +21,11 @@ final class TimelineDashboardViewController: NSViewController {
     @IBOutlet weak var recordSwitcher: OGSwitch!
     @IBOutlet weak var collectionView: TimelineCollectionView!
     @IBOutlet weak var emptyLbl: NSTextField!
-    @IBOutlet weak var emptyActivityLbl: NSTextField!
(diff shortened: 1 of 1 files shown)


skel35 (approved):
looks good.
as you've said, we'll just have to see after a while if it still appears in Bugsnag

Issue #4580, opened 2020-10-21
Global Ctrl + F10 shortcut breaks Ctrl completely in X11


MartinBriza:
### 💻 Environment
Platform: Linux
Toggl Version: 7.5.324

### 🔨 Steps to reproduce
1. Set a global keyboard shortcut in TogglDesktop to Ctrl+F10
2. Try if it works when a different app is in focus
3. Try if Ctrl works in another app (like Ctrl+A to select text)

### 🐞 Actual behavior
Ctrl doesn't work in other apps

### 💯 Expected behavior
Ctrl should work same as before for other shortcuts.

### 📦 Additional info
Introduced with 453cdcb5e271ea7c411f2417f6f5ee318fda7f78 (which fixed other issues though)
https://app.intercom.com/a/apps/ayixs927/inbox/inbox/534741/conversations/106481304722762

PR #4643, opened 2020-11-05
Fix global keyboard shortcut issues

MartinBriza:
### 📒 Description
This PR is a combination of these sources of code:
 - https://github.com/qt/qtbase/tree/dev/src/plugins/platforms/xcb
 - https://github.com/hluk/qxtglobalshortcut/blob/master/src/qxtglobalshortcut_x11.cpp
 - Some experiments with the order of returning results based on our previous code

### 🕶️ Types of changes

### 🤯 List of changes
- Global keyboard shortcuts now should actually work as expected

### 👫 Relationships
#4580 

### 🔎 Review hints
Try setting some (especially "exotic") keyboard shortcuts and observe if they:
 * Work as expected
 * Don't break other global shortcuts
In case this PR breaks some more stuff, it may be worth reverting #4416

### DIFF
--- a/third_party/qxtglobalshortcut5/gui/qxtglobalshortcut_x11.cpp
+++ b/third_party/qxtglobalshortcut5/gui/qxtglobalshortcut_x11.cpp
@@ -29,7 +29,6 @@
 ** <http://libqxt.org>  <foundation@libqxt.org>
 *****************************************************************************/
 
-#include <QVector>
 #if QT_VERSION < QT_VERSION_CHECK(5,0,0)
 #   include <QX11Info>
 #else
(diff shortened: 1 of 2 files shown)


atsvirchkova (changes requested):
I don't see, that it fixes the issue. I tested on Ubuntu 20.04 GNOME.
- Set continue/stop shortcut in Toggl track to Ctrl+F10
- Check if Ctrl+A works in other apps. It doesn't work for me. After I close Toggl track it starts to work fine again.

skel35:
OK, thanks @atsvirchkova. Then let's just revert #4416 for now as the issue fixed there seems not so critical.

PR #4644, opened 2020-11-05
[DO NOT MERGE] ARM macOS Build

MartinBriza:
This is not really a pull request, more of an explanation on what I had to do to make the ARM macOS build work.
Please note this is just about ARM, not Universal binaries, however I expect building an Universal binary will be basically the same, with the difference of the target platform that gets passed to the compiler.

 - First, it was necessary to update OpenSSL (which is something that should be done much more often regardless). There is a new version that supports builds for ARM natively.

 - Second, POCO. POCO has to be rebuilt and the binary sofiles that we`re storing in the repo should be updated to support either ARM or the Universal binaries.

 - Cocoapods don`t (or didn`t, when I tried it) support ARM builds so I had to rebuild AppAuth and Sparkle from source code and inject them into the Pods folder. I suppose this will get sorted out soon though from the Cocoapods side. I included the binary folders inside the Library folder as AppAuth-actual and Sparkle-actual folders so you can try it yourselves if you want (and can, because ARM macs are not available yet)

 - And finally, the build process has to be adjusted slightly to change the compilation target platform.

All of these steps are also separated into commits in this PR, however this PR shoudln`t be used directly, take it more of a guideline to how to do it in the future. Also there`s probably no hurry because the x86 app still runs fine on the latest Big Sur beta on the DTK.


AndrewVebster:
Thanks a lot @MartinBriza ! That's very helpful 👍

PR #4649, opened 2020-11-09
Update keyboard navigation inside duration dropdown (mac)

### LINKED ISSUE #4610: Include Day Control into keyboard focus flow on Duration Dropdown
### 💻 Environment
Platform: macOS

### 📒 Description
Keyboard focus flow should also include Day Control. So the final flow looks like this:
-> Focus duration field 
-> Opens dropdown 
-> Tab -> move to Start 
-> Tab -> Day control left arrow 
-> Tab -> Day control middle/day label (optional) 
-> Tab -> Right arrow 
-> Tab -> Close dropdown

AndrewVebster:
### 📒 Description

This PR aims to improve keyboard navigation on duration dropdown.
We need to discuss what is the best way to do that.

Note that this PR works differently from what is described in Issue #4610.

On macOS there is an option in system Preferences:
Screenshot 2020-11-05 at 17 38 24

In the app we check if option is enabled:
- Yes (keyboard navigation is extended) - 
-> Opens dropdown
-> Tab -> move to Start
-> Tab -> Day control left arrow
-> Tab -> Date picker
-> Tab -> Right arrow
-> Tab -> Close dropdown
- No - 
-> Opens dropdown 
-> Tab -> move to Start 
-> Tab -> Close dropdown

### 👫 Relationships

Closes #4610

### 🔎 Review hints

Duration dropdown is hidden under the local feature flag. To enable it go to TimerDurationControl.swift line 28 and set isDurationDropdownEnabled to true.

Tell me what do you think.

### DIFF
--- a/src/ui/osx/TogglDesktop/Features/AutoComplete/GenericAutoComplete/AutoCompleteView.swift
+++ b/src/ui/osx/TogglDesktop/Features/AutoComplete/GenericAutoComplete/AutoCompleteView.swift
@@ -197,7 +197,7 @@ extension AutoCompleteView {
             self?.dataSource?.selectRow(at: clickedRow)
         }
 
-        createNewItemBtn.didPressKey = { key in
+        createNewItemBtn.didPressKey = { key, _ in
             if key == .tab {
                 self.delegate?.shouldClose()
             }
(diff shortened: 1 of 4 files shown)


skel35 (changes requested):
Works mostly fine.
It seems weird that Space simply moves the focus forward, instead of opening the calendar.
Can we use the same keyboard navigation as we have on a date picker inside the Edit view?

AndrewVebster:
I chose to leave the calendar out of keyboard navigation at all. Theoretically, we can do the same as on the Edit screen but because here the calendar is embedded (and it's not another popover) it will take more effort to do the navigation properly.
Maybe we can do this after we see that someone is using the date control at all?

skel35 (approved):
approved

PR #4654, opened 2020-11-10
Remove the ability to open Edit popover for running entry from Timer

### LINKED ISSUE #4619: Remove the ability to open Edit popover for running entry from Timer
### 💻 Environment

Platform: macOS

### 📒 Description
When duration dropdown is implemented there is no need in Edit popover for running TE. It should be removed, so when TE is running and the user clicks on Timer background it's not shown.

AndrewVebster:
### 📒 Description

We replace "clickable" box behind Timer UI with a non-clickable.

Because this PR closes the last Issue for duration dropdown, it also enables the local feature flag so it's available for everyone.

### 🕶️ Types of changes

### 👫 Relationships

Closes #4619


skel35 (commented):
why not removing `isDurationDropdownEnabled` variable altogether?

AndrewVebster:
Yep, no point in keeping `isDurationDropdownEnabled` - removed

skel35 (approved):
:shipit:

PR #4645, opened 2020-11-05
Click-drag to create time entry

atsvirchkova:
### 📒 Description
- Clicking the Timeline view (or dragging for less than 2 px) creates a time entry with a duration of 1hr.
- Clicking and dragging on Timeline view creates a TE, whose start and end corresponds to dragged distance and offset.

### 🕶️ Types of changes

 - **New feature** (non-breaking change which adds functionality)

### 👫 Relationships

Closes #4621 
Closes #4622

PR #4662, opened 2020-11-12
Fix resizing stops when TE update arrives

### LINKED ISSUE #4563: When the TE update arrives during resizing TE, the resizing stops
### 💻 Environment
Platform: Windows
Toggl Version: master after #4532

### 🐞 Actual behavior
The resizing is being stopped even though the user hasn't finished the mouse dragging (no MouseUp event).

### 💯 Expected behavior
The resizing is not stopped if the user hasn't finished the mouse dragging.

### 🔨 Steps to reproduce
1. Resize the TE.
2. Quickly start performing another resize
3. Wait 1-2 seconds while keeping the left mouse button pressed.

atsvirchkova:
### 📒 Description
Updating of time entry blocks is blocked if any of the items is being grabbed.

### 🕶️ Types of changes

### 👫 Relationships

 Closes #4563 

### 🔎 Review hints
To be reviewed after #4645 is merged.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/controls/TimelineRunningTimeEntryBlock.xaml.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/controls/TimelineRunningTimeEntryBlock.xaml.cs
@@ -31,6 +31,12 @@ namespace TogglDesktop
         private void OnThumbTopDragCompleted(object sender, DragCompletedEventArgs e)
         {
             ViewModel.ChangeStartTime();
+            ViewModel.IsDragged = false;
+        }
+
+        private void OnThumbDragStarted(object sender, DragStartedEventArgs e)
+        {
+            ViewModel.IsDragged = true;
         }
     }
 }
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/controls/TimelineRunningTimeEntryBlock.xaml.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/controls/TimelineRunningTimeEntryBlock.xaml.cs
@@ -1,5 +1,6 @@
 ﻿using System.Windows.Controls;
 using System.Windows.Controls.Primitives;
+using System.Windows.Input;
 using TogglDesktop.ViewModels;
 
 namespace TogglDesktop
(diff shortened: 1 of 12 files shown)


skel35 (changes requested):
Looks quite good, nice job!
One thing that's bothering me a bit - sometimes the tooltip does not update when resizing the running TE, especially when resizing twice in a row. I could not reproduce this issue when resizing the stopped TEs.
AOR3PPsfyp

atsvirchkova:
The issue with tooltip is finally fixed now. Found another bug meanwhile: duration on tooltip (and description) for running TE does not coincide with that on timer. I reported it separately #4676.

src/ui/windows/TogglDesktop/TogglDesktop/ui/controls/TimelineRunningTimeEntryBlock.xaml.cs
```diff
@@ -31,6 +32,26 @@ private void OnThumbTopDragDelta(object sender, DragDeltaEventArgs e)
...
+        {
+            ViewModel.IsDragged = true;
+        }
+
+        private TimelineTimeEntryBlockPopup _popup;
+        protected override void OnMouseEnter(MouseEventArgs e)
+        {
+            _popup ??= new TimelineTimeEntryBlockPopup();
+            _popup.DataContext = ViewModel;
+            _popup.Popup.PlacementTarget = this;
```
skel35, line 48: Why did you decide to change the logic of the popup positioning for the running time entry?
Old issues like this one are back now.
image

skel35 (changes requested):
In Timeline.xaml.cs there's already a logic that assigns DataContext, sets PlacementTarget and shows the popup. Partially duplicating this logic in another class makes little sense to me both from the UX and from the code perspective.

atsvirchkova:
This was the main pain point of yesterday's investigation. The reason for moving out the existing logic is that it simply didn't work the way it was, I mean the bug with not-updating tooltip stayed for running TE. I've just pushed the branch https://github.com/toggl-open-source/toggldesktop/tree/test/timeline_ui/resizing_bugfix where the separate logic for running TE is commented, and the existing in Timeline.xaml.cs is reused. You'll be able to reproduce the tooltip bug:
1. Grab running TE start and move it, then drop it but don't move the mouse (so that the tooltip stays).
2. Then grab it once more and move again, you'll see the duration and start time on tooltip won't change.

Though I still don't understand how that logic different from this one.

skel35:
@atsvirchkova https://github.com/toggl-open-source/toggldesktop/tree/test/timeline_ui/resizing_bugfix is exactly the same as the current one.

atsvirchkova:
Sorry, forgot to commit the changes. Now it's different.

skel35:
Maybe you misunderstood me, but I didn't mean the separate popup (although it would be very good to understand why it doesn't work without a separate popup). The issue in this PR is that the popup positioning is messed up (see screenshot above), and the positioning logic isn't reused for some reason.
I've just taken your branch and made a couple of small changes: https://github.com/toggl-open-source/toggldesktop/tree/fix/timeline_ui/resizing-v2. I can't reproduce the positioning issue anymore and the code is reused.
It's quick&dirty fix but I hope you get the point.

atsvirchkova:
Thank you. Yeh, I understand the issue with the positioning. It's not a big deal to reuse the positioning logic, I will move it to this branch. I just spent quite a lot of time trying to make it work with the same popup and gave up at the end.

atsvirchkova:
Popup logic is fixed now.

skel35 (approved):
Resizing works nice and smooth now. Good job!

PR #4668, opened 2020-11-13
Feature/duration dropdown improvements

### LINKED ISSUE #4667: Small duration dropdown improvements
### 💻 Environment
Platform: macOS

### 📒 Description
Small UX improvements for duration dropdown:
1. Field labels - the text is a bit too big currently, please change it to a smaller size
1. Date formatting - if possible to do it easy, please do it so that the day of the week is with 3 letters only so the text fits nicely there
1. Next date arrow should be disabled if the date is today
1. If I click on the project/tag icon on the timer, the duration dropdown should close. Now they both are visible at the same time
1. If I click on the header of the window close the duration dropdown (optional)
1. If I mess up the start time… e.g add a random text there and then press Tab key to close the dropdown then after when I open the dropdown it should update the start time again to correct time and not keep the unrecognizable weird number combination I added.

AndrewVebster:
### 📒 Description

Small UX fixes for the duration dropdown. 
The list of fixes is in the related issue (or see commit messages).

### 🕶️ Types of changes

### 👫 Relationships

Closes #4667

### 🔎 Review hints

Instead of reviewing all PR changes, it can be easier to review one commit at a time as one commit = one fix

### DIFF
--- a/src/ui/osx/TogglDesktop/Features/Timer/TimerEditViewController/TimerEditViewController.m
+++ b/src/ui/osx/TogglDesktop/Features/Timer/TimerEditViewController/TimerEditViewController.m
@@ -193,4 +193,9 @@ extern void *ctx;
     }
 }
 
+-(void)mouseDown:(NSEvent *)event
+{
+    [self.timerViewController mouseDown:event];
+}
+
 @end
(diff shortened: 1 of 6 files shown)


skel35 (approved):
The issues mentioned in #4667 are fixed. No remarks from my side.
LGTM

PR #4679, opened 2020-11-17
Fix "date change" analytics sending at app launch (mac)

### LINKED ISSUE #4677: Duration dropdown "date-change" event is reported at app launch on macOS
### 💻 Environment

Platform: macOS
OS Version: Any
Toggl Version: 7.5.383

### 🐞 Actual behavior

Analytics event `timer-duration-dropdown` - `date-change` is reported at every app launch.

### 💯 Expected behavior

Event should be reported only if user tries to change start date.

### 🔨 Steps to reproduce

1. Set breakpoint to `TimeEditView.swift` line 42
1. Launch app in debug mode

AndrewVebster:
### 📒 Description
Assigning `dateValue` from `awakeFromNib()` would call `didSet` block of this property and as a result make a call to the analytics service.
Setting initial `dateValue` in `init` fixes this problem because `didSet` won't be called then.

### 🕶️ Types of changes

### 👫 Relationships

Closes #4677

### 🔎 Review hints

To test you can set breakpoint to `TimeEditView.swift` line 42 and observe when it's called.

### DIFF
--- a/src/ui/osx/TogglDesktop/Features/Timer/TimerViewController/Views/TimeEditView.swift
+++ b/src/ui/osx/TogglDesktop/Features/Timer/TimerViewController/Views/TimeEditView.swift
@@ -81,9 +81,15 @@ class TimeEditView: NSView {
 
     // MARK: - Overrides
 
+    required init?(coder: NSCoder) {
+        dateValue = Date()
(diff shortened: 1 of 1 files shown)


skel35 (approved):
looks good

PR #4680, opened 2020-11-17
Include duration dropdown date controls into key loop for everyone (mac)

### LINKED ISSUE #4678: Include day control into key view loop on duration dropdown for all macOS users
### 💻 Environment

Platform: macOS

### 📒 Description

Currently, when the user tabs through controls on Duration Dropdown it will focus on "day control" only if the user has enabled "Use keyboard navigation" in System Preferences.
We should enable this behavior for all users.

AndrewVebster:
### 📒 Description

On the Duration Dropdown the keyboard navigation now will work the same, regardless of "Use keyboard navigation" option in System Preferences.

### 🕶️ Types of changes

### 👫 Relationships

Closes #4678


skel35 (approved):
#4678 is addressed, keyboard navigation in duration dropdown works the same way regardless of the system setting.

PR #4681, opened 2020-11-17
Use environment files instead of set-env (actions)

### LINKED ISSUE #4516: Upgrade from `set-env` to environment files
### 💻 Environment
Platform: GitHub Actions

### 📒 Description
Upgrade from `set-env` to environment files in our GitHub Actions scripts.

>The `set-env` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

skel35:
### 📒 Description
Use environment files instead of the deprecated `set-env` feature (actions).
See https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files

### 🕶️ Types of changes

### 👫 Relationships
Closes #4516 

### 🔎 Review hints
Check the code and check that CI does not fail.
Environment variables are used in the macOS scripts, so check that the environment that is printed before each GitHub Actions step looks the same as in the previous successful runs. 
For example: image

Docs: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,12 +28,12 @@ jobs:
         echo $installer_name
 
         # Export env to all future steps
-        echo "::set-env name=TAG_NAME::$TAG_NAME"
-        echo "::set-env name=version::$version"
-        echo "::set-env name=timestamp::$timestamp"
-        echo "::set-env name=escaped_version::$escaped_version"
-        echo "::set-env name=installer::$installer"
-        echo "::set-env name=installer_name::$installer_name"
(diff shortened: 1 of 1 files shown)


AndrewVebster (approved):
LGTM, thanks for quick fix!

PR #4692, opened 2020-11-18
Fix incorrect gap blocks generation with selected date other than today

### LINKED ISSUE #4691: Extra gap time entry button if there is a running TE and today is not selected
Gap time entry button is generated starting from the last TE (of the selected date) if selected date is not today and there is a running TE.

### 💻 Environment

Platform: Windows
OS Version: all Windows
Toggl Version: 7.5.389

### 🐞 Actual behavior

image

### 💯 Expected behavior
No extra buttons.

### 🔨 Steps to reproduce

1. Start a time entry.
2. Go to Timeline UI and select some day (not today) where there is at least on TE.
3. See a gap TE button lasting until the end of day.

P. S. This button disappears as soon as the TE is stopped.

### 📦 Additional info
For the fix: make sure to pass the Running TE for gaps generation function only if selected date  is today.

atsvirchkova:
### 📒 Description
Remove running TE from calculating gap entries if selected date is not today.

### 🕶️ Types of changes

### 👫 Relationships

 Closes #4691 

### 🔎 Review hints
Testing:
Start a TE, then go to another day where you have some other TEs. Check there is no gap TE button after the last TE.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
@@ -69,8 +69,8 @@ namespace TogglDesktop.ViewModels
                         .ToDictionary(pair => pair.Key, pair => pair.Value))
                 .ToPropertyEx(this, x => x.TimeEntryBlocks);
             blocksObservable.Select(blocks => GenerateGapTimeEntryBlocks(blocks.Values
-                .Where(v => Toggl.DateTimeFromUnix(v.Started).Date <= SelectedDate &&
-                            Toggl.DateTimeFromUnix(v.Ended).Date >= SelectedDate).ToList(), SelectedScaleMode)).ToPropertyEx(this, x => x.GapTimeEntryBlocks);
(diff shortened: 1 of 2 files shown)


skel35 (commented):
1. How about changing the condition to: "TE started on SelectedDate OR TE ended on SelectedDate"?
2. I've noticed we're using `Toggl.DateTimeFromUnix(...)` a lot, especially on `TimeEntryBlock`. How about adding some extension methods, e.g. `DateTime StartedDate(this TimeEntryBlock);`?

atsvirchkova:
1. "TE started on SelectedDate OR TE ended on SelectedDate" doesn't take into account TEs which last for over 24 hours (e. g. TE started the day before yesterday and finished today). I know this might not be the use case, but formally it's possible.
2. Yes, sounds like a good idea. It's not needed even to be an extension method, it can be added as class member.

skel35:
1. That makes sense 👍 
2. In general I prefer extension methods over instance methods, for a better separation of concerns and SRP application. Ties better with FP too. If it's used a lot in the same way, an instance method is probably OK too.

Maybe also change the variable name `v` to something more readable, like `timeEntry` or `timeEntryBlock`.

atsvirchkova:
Introduced some extension methods for `TimeEntryBlock` , `Toggl.TogglTimeEntryView` and `Toggl.TimelineChunkView`. Ready for review again.

skel35 (approved):
Nice, looks good 👍

PR #4694, opened 2020-11-19
Fix gap TE overlap & Running TE visibility

atsvirchkova:
### 📒 Description
Fix gap TE overlap & Running TE visibility

### 🕶️ Types of changes

### 🔎 Review hints

During quick fix of #4691 a new bug was introduced: a gap TE now always overlaps with previous TE. It's not visible while it's a gap, but a new TE, which would be created from it would go to the second layer of TEs.
Also noticed a strange issue with running TE description being displayed while there is no running TE. Fixed visibility.


atsvirchkova:
Steps to reproduce for visibility issue (tested on 7.5.392):
1. Go to Timeline UI and choose today.
2. There will be some strange description field on top of the day.

skel35 (approved):
Confirmed this fixes both of the mentioned issues. Looks good.

PR #4331, opened 2020-08-17
Kill the app process when the updater is launched

skel35:
### 📒 Description
Kill the app process when the updater is launched.
Apparently, `Environment.Exit(0)` makes the application hang on Windows 7 under certain conditions.

### 🕶️ Types of changes
 - **Bug fix** (non-breaking change which fixes an issue)

### 🤯 List of changes
- [x] Kill the app process instead of `Environment.Exit(0)` when the updater is launched

### 👫 Relationships
Closes #4247
Closes #4317

### 🔎 Review hints
Test the auto-update on Windows 7 and Windows 10.
It should be fine to use only the build from this branch because the bug occurs in the application *from* which the update is performed (as opposed to the app *to* which the update is performed).

PR #4703, opened 2020-11-20
Fix quitting the app doesn't end the process on Windows 7

### LINKED ISSUE #4682: Application doesn't launch on Windows 7
### 💻 Environment

Platform: Windows
OS Version: Windows 7
Toggl Version: 7.5.383

### 🐞 Actual behavior
Toggl process stays in Task Manager after app quit.

### 💯 Expected behavior
No Toggl processed must stay alive after app quit.

### 🔨 Steps to reproduce

1. Start Windows 7 and run Toggl Track.
2. Quit Toggl Track and see that its process stays in Task manager.
3. It's impossible to start Toggl Track again until the process is killed in Task manager.

### 📦 Additional info
It's reproducible ONLY after Windows 7 restart, so to reproduce it once more it's necessary to reboot Windows.

atsvirchkova:
### 📒 Description
Fix quitting the app doesn't end the process on Windows 7

### 🕶️ Types of changes

### 👫 Relationships

Closes #4682 

### 🔎 Review hints
Like in #4331 `Process.GetCurrentProcess().Kill()` is used here. I inserted a check for Windows version to do that only for Windows 7, but for Windows 10 it also works OK, just feels like incorrect process ending.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/Program.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/Program.cs
@@ -45,7 +45,13 @@ static class Program
     public static void Shutdown(int exitCode)
     {
         Toggl.Clear();
-        Environment.Exit(exitCode);
+        //For mysterious reason on Windows 7 Exit() doesn't always unload the process of TogglDesktop.exe from memory,
+        //and it makes impossible to start the app again next time.
+        //See https://github.com/toggl-open-source/toggldesktop/issues/4682 for more details.
+        if (Utils.IsWindows7())
+            Process.GetCurrentProcess().Kill();
+        else
+            Environment.Exit(exitCode);
     }
 
     public static string Version()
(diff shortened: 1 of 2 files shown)


src/ui/windows/TogglDesktop/TogglDesktop/Program.cs
```diff
@@ -45,7 +45,10 @@ private static void OnBeforeStartup()
     public static void Shutdown(int exitCode)
     {
         Toggl.Clear();
-        Environment.Exit(exitCode);
+        if (Environment.OSVersion.Version.Major == 6 && Environment.OSVersion.Version.Minor == 1)
```
skel35, line 48: It is unclear what exactly this means by just reading it (even though I understand this and could look up to which Windows version this corresponds to).
I'd extract an extension or utility method (smth like `IsWindows7()`) + maybe add a small comment about this issue.

skel35 (changes requested):

src/ui/windows/TogglDesktop/TogglDesktop/Program.cs
atsvirchkova: Good point, thanks.
Fixed now.

skel35 (approved):
Looks good

PR #4707, opened 2020-11-23
Fix TEs are displayed overlapped when they are actually not

### LINKED ISSUE #4673: TEs are displayed as overlapped when they are actually not
### 💻 Environment
Platform: Windows
Toggl Version: 7.5.382

### 🐞 Actual behavior
TEs are shown as overlapped.

### 💯 Expected behavior
TEs should not be shown as overlapped

### 🔨 Steps to reproduce
1. TE #1: 10:28:18 -> 10:30:28
2. TE #2: 10:30:30 -> 10:43:44
3. Zoom out to the last or second last zoom level.

### 📦 Additional info
image

macOS:
image

Note: It's better to show overlaps correctly instead of making sure the TE is fully visible, so it's fine if the TE is rendered partially on top of the previous one.

atsvirchkova:
### 📒 Description
Fix TEs are displayed overlapped when they are actually not.

### 🕶️ Types of changes

### 👫 Relationships

Closes #4673 

### 🔎 Review hints
Some refactoring is done is scope if this bugfix:
1. Started and Ended time stamps are removed from `TimeEntryBlock` class, as all necessary info can be got from `VerticalOffset` and `Height` properties.
2. Height and offset properties now correspond to time stamps strictly, the case of zero time entry length is handled in xaml code by setting `MinHeight`.
3. There is now  a precision constant introduced for doubles comparison, so everything that has an overlap less than 10^-5 is not considered to be an actual overlap.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/Resources/TimelineConstants.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/Resources/TimelineConstants.cs
@@ -12,6 +12,7 @@ namespace TogglDesktop.Resources
         public const double DescriptionOpacity = 0.2;
         public const double TimeEntryBlockWidth = 20;
         public const double GapBetweenOverlappingTEs = 5;
+        public const double AcceptableBlocksOverlap = 1e-5;
 
         public static IReadOnlyDictionary<int, int> ScaleModes { get; } = new Dictionary<int, int>()
         {
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineBlockViewModel.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineBlockViewModel.cs
@@ -15,6 +15,8 @@ namespace TogglDesktop.ViewModels
         public double Height { get; set; }
         [Reactive]
         public double VerticalOffset { get; set; }
+
+        public double Bottom => VerticalOffset + Height;
     }
 
     public class GapTimeEntryBlock : TimelineBlockViewModel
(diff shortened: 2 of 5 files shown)


skel35 (changes requested):
Bug discovered.

STR
1. Click the running time entry to open the Edit view
2. Close the Edit view
3. Switch days

AR
Edit view sometimes opens up by itself

Note: when TE is stopped the issue is still happening

atsvirchkova:
@skel35 I cannot reproduce the bug. There is logic in the code to close Edit view if date has changed (by closing I mean calling `Toggl.Edit` which may also open Edit view if it's closed) and I think this may cause the issue. But then it should be reproducible in master also, there was no change to this logic in this PR.

skel35:
You're right that it is reproducible on `master` as well.
It's weird that you can't reproduce it, it's straightforward. I'll open an issue.

skel35:
OK, now I can't reproduce it myself 😄 
I will try to figure this out tomorrow, otherwise, we'll need to look into the code to find out how this state can be achieved.

atsvirchkova:
I suppose, you have to be faster than your PC to reproduce this bug. I think it happens when `SelectedForEditTEId` is not nulled yet after edit view close, and you change the date at that time.

atsvirchkova:
Fixed the two mentioned issues. As for making a `BehaviorSubject<(Started, Ended)>` and using it for `ChangeStartTime` and `ChangeEndTime` - this will result into calling this two functions every time a new value arrives which is not desirable as they call lib functions that trigger saving to database.

skel35 (approved):
Looks good!
And the bug with running time entry displayed on the future days is also fixed 👍

PR #4714, opened 2020-11-25
Activate start button on Enter key press (mac)

### LINKED ISSUE #4669: Activate Start/Stop button on Enter key press when focused
### 💻 Environment

Platform: macOS

### 📒 Description

This is needed because not all users are aware that on macOS by default the `Space` key activates focused controls and not `Enter` (like on Windows).
Having both keys to perform the same action, in this case, should improve the UX.

AndrewVebster:
### 📒 Description

Start button on the timer now activates on Enter and Space key press.

### 🕶️ Types of changes

### 👫 Relationships

Closes #4669

### 🔎 Review hints

Refactored buttons + rewrote `NSHoverButton` from Objective C to Swift.
Because of that, it would be nice to check if buttons in other places behave as before.


skel35 (approved):
The start button is getting clicked on Enter key press as expected.
Tested other buttons throughout the application, didn't find any differences between this and `master`.

PR #4716, opened 2020-11-27
Fix non-accurate duration for running TE

### LINKED ISSUE #4676: Duration for running TE does not coincide with that in timer
### 💻 Environment

Platform: Windows
OS Version: all Windows
Toggl Version: 7.5.383

### 🐞 Actual behavior
image

### 💯 Expected behavior
Minutes on timer and on description for running TE should not differ for over one minute.

### 🔨 Steps to reproduce

1. Open Timeline UI and create running TE.
2. Decrease the scale, than this bug will be more obvious.

### 📦 Additional info
This happens because of substracting one from running TE height to not overlap with current time line. End time of time entry block depends on height, and duration depends on end time.

atsvirchkova:
### 📒 Description
This PR fixes duration on the tooltip for running TE, which was a bit shorter than expected. 

### 🕶️ Types of changes

### 👫 Relationships

 Closes #4676 

### 🔎 Review hints
Shorter duration was caused by shorter height for running TE block (it was needed to not obscure the current time line). Now the height for running TE is normal, and order of canvases for running TE and current line is changed (so that the line is always over the running entry). Also the click and drag feature was placed in a separate canvas to avoid creating a TE when clicking on running TE.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
@@ -90,7 +90,7 @@ namespace TogglDesktop.ViewModels
                 .Subscribe(h => CurrentTimeOffset = h);
             this.WhenAnyValue(x => x.CurrentTimeOffset).Where(_ => RunningTimeEntryBlock != null)
                 .Select(off => Math.Max(TimelineConstants.MinTimeEntryBlockHeight,
-                    CurrentTimeOffset - RunningTimeEntryBlock.VerticalOffset - 1))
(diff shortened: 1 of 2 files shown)


skel35 (changes requested):
It's better now, i.e., the difference is smaller, but the issue is still not fixed:
image

atsvirchkova:
Yes, I know this. But this won't be that easy to fix, because duration in tooltip is updated once a minute, and this update is linked to current time line, not to the point of start of the timer. I'll see if there is an option to reset the current time line timer, when there is a new running TE.

skel35:
OK, then let's consider this issue the same as #4722 and close this one, as the problem is not so critical anymore and this PR looks good.

PR #4719, opened 2020-11-27
Timeline UI - fix overlapping for running TE

### LINKED ISSUE #4717: Timeline UI - overlapping running TE is not considered as overlap when it's too small
### 💻 Environment

Platform: Windows
OS Version: all Windows
Toggl Version: 7.5.401

### 🐞 Actual behavior
image

### 💯 Expected behavior
Must hide the description of the first TE, and show running TE in the second layer like this
image

### 🔨 Steps to reproduce

1. Create a TE that crosses current time line.
2. Start a new TE.

### 📦 Additional info
This is not present in 7.4.394 release, it was introduced lately by PR #4707.

atsvirchkova:
### 📒 Description
Fixes that running TE is not considered as overlap when it's too small.

### 🕶️ Types of changes

### 👫 Relationships

Closes #4717 

### 🔎 Review hints
For running TE end time stamp was less than start time stamp, this caused wrong sorting in the layout algorithm.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
@@ -284,8 +284,6 @@ namespace TogglDesktop.ViewModels
 
         public static double ConvertTimeIntervalToHeight(DateTime start, DateTime end, int scaleMode)
         {
-            if (end <= start) return 0;
-            
             var timeInterval = (end - start).TotalMinutes;
             return timeInterval * TimelineConstants.ScaleModes[scaleMode] / 60;
         }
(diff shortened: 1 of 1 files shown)


src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
```diff
@@ -284,6 +284,8 @@ private enum TimeStampType
 
         public static double ConvertTimeIntervalToHeight(DateTime start, DateTime end, int scaleMode)
         {
+            if (end <= start) return 0;
```
skel35, line 287: as discussed in Slack, this line doesn't fix the given bug, and having `end < start` potentially uncovers other issues, so it might be better to address those separately

atsvirchkova: OK, agree with you. Already removed this from fix.

skel35 (approved):
Overlapping now works as expected 👍

PR #4720, opened 2020-11-30
Context menu: continue, start from this entry, delete

### LINKED ISSUE #4684: Context menu for TEs in Timeline view: continue, start from this entry, delete
### 💻 Environment

Platform: Windows

### 📒 Description
Introduce a context menu for time entries in the Timeline view. This is part of #4670 
Context menu items:

- "Continue this entry"
- "Start entry from the end of this entry" if the selected day is today, or "Create entry from the end of this entry" if the selected day is different (will create a 1h time entry and open the Edit view).
- "Delete"

### ⭐️ Why do you want this?
This is part of Timeline UI functionality.

atsvirchkova:
### 📒 Description
Implementation of context menu for time entry block. List of menu items:
 - Continue
- "Start entry from the end of this entry" if the selected day is today, or "Create entry from the end of this entry" if the selected day is different (will create a 1h time entry and open the Edit view).
- Delete

### 🕶️ Types of changes

### 👫 Relationships

 Closes #4684 

### 🔎 Review hints
Test:
Right-click on any TE, there is supposed to be 3 menu-items.
For running TE 2 of 3 items are disabled and only "Delete" is accessible.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/Resources/TimelineConstants.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/Resources/TimelineConstants.cs
@@ -14,6 +14,8 @@ namespace TogglDesktop.Resources
         public const double GapBetweenOverlappingTEs = 5;
         public const double AcceptableBlocksOverlap = 1e-5;
         public const double MinGapTimeEntryHeight = 10;
+        public const int DefaultTimeEntryLengthInSeconds= 3600;
+        public const double DefaultTimeEntryLengthInHours = 1;
 
         public static IReadOnlyDictionary<int, int> ScaleModes { get; } = new Dictionary<int, int>()
         {
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/Resources/TimelineConstants.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/Resources/TimelineConstants.cs
@@ -14,7 +14,7 @@ namespace TogglDesktop.Resources
         public const double GapBetweenOverlappingTEs = 5;
         public const double AcceptableBlocksOverlap = 1e-5;
         public const double MinGapTimeEntryHeight = 10;
-        public const int DefaultTimeEntryLengthInSeconds= 3600;
+        public const int DefaultTimeEntryLengthInSeconds= (int) (DefaultTimeEntryLengthInHours * 3600);
         public const double DefaultTimeEntryLengthInHours = 1;
 
         public static IReadOnlyDictionary<int, int> ScaleModes { get; } = new Dictionary<int, int>()
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineBlockViewModel.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineBlockViewModel.cs
@@ -46,13 +46,14 @@ namespace TogglDesktop.ViewModels
         public string TaskName => _timeEntry.TaskLabel;
         public bool HasTag => !_timeEntry.Tags.IsNullOrEmpty();
         public bool IsBillable => _timeEntry.Billable;
+        public long DurationInSeconds => _timeEntry.DurationInSeconds;
         public string Duration { [ObservableAsProperty]get; }
         public string StartEndCaption { [ObservableAsProperty]get; }
         public ReactiveCommand<Unit, Unit> OpenEditView { get; }
         public ReactiveCommand<Unit, bool> ContinueEntry { get; }
         public ReactiveCommand<Unit, Unit> CreateFromEnd { get; }
         public ReactiveCommand<Unit, Unit> StartFromEnd { get; }
-        public ReactiveCommand<Unit, bool> Delete { get; }
+        public ReactiveCommand<Unit, Unit> Delete { get; }
         public string TimeEntryId => _timeEntry.GUID;
 
         [Reactive]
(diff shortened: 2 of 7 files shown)


src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineBlockViewModel.cs
```diff
@@ -72,6 +77,11 @@ public TimeEntryBlock(Toggl.TogglTimeEntryView te, int hourHeight, DateTime date
             DateCreated = date;
             _timeEntry = te;
             OpenEditView = ReactiveCommand.Create(() => Toggl.Edit(TimeEntryId, false, Toggl.Description));
+            var isNotRunningObservable = Observable.Return(_timeEntry.DurationInSeconds >= 0);
+            ContinueEntry = ReactiveCommand.Create(() => Toggl.Continue(TimeEntryId), isNotRunningObservable);
+            CreateFromEnd = ReactiveCommand.Create(() => TimelineUtils.CreateAndEditTimeEntry(Ended, Ended + 3600), isNotRunningObservable);
+            StartFromEnd = ReactiveCommand.Create(() => TimelineUtils.CreateAndEditRunningTimeEntryFrom(Ended), isNotRunningObservable);
+            Delete = ReactiveCommand.Create(() => Toggl.DeleteTimeEntry(TimeEntryId));
```
skel35, line 84: There should be a confirmation message when deleting the time entry - see `ModelExtensions.DeleteTimeEntry()`

src/ui/windows/TogglDesktop/TogglDesktop/ui/views/Timeline.xaml
```diff
@@ -205,6 +217,7 @@
                         <ItemsControl.ItemTemplate>
                             <DataTemplate>
                                 <togglDesktop:TimelineTimeEntryBlock Panel.ZIndex="1" DataContext="{Binding Path=Value}"
+                                                                     MouseDown="OnTimeEntryBlockMouseDown"
```
skel35, line 220: MouseDown - this looks hacky. All the single-clicking, dragging, and resizing can be potentially captured.
ContextMenuOpening seems a better option here.

It'd also be great to make it more modular, e.g. by extracting it to Behavior with an attached property and OneWayToSource binding to ActiveTimeEntryBlock, not sure if it's possible here.

skel35 (changes requested):

src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineBlockViewModel.cs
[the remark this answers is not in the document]
atsvirchkova: I can do it, of course, though I like more the current solution, because it doesn't require to handle the case, when `ActiveTimeEntryBlock` is null.

src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineBlockViewModel.cs
```diff
@@ -72,6 +77,11 @@ public TimeEntryBlock(Toggl.TogglTimeEntryView te, int hourHeight, DateTime date
             DateCreated = date;
             _timeEntry = te;
             OpenEditView = ReactiveCommand.Create(() => Toggl.Edit(TimeEntryId, false, Toggl.Description));
+            var isNotRunningObservable = Observable.Return(_timeEntry.DurationInSeconds >= 0);
+            ContinueEntry = ReactiveCommand.Create(() => Toggl.Continue(TimeEntryId), isNotRunningObservable);
+            CreateFromEnd = ReactiveCommand.Create(() => TimelineUtils.CreateAndEditTimeEntry(Ended, Ended + 3600), isNotRunningObservable);
```
skel35, line 82: One more issue - hardcoded number 3600.
We're also creating 1h-long time entries in click-to-create, so the same constant should be used in both places.

src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineBlockViewModel.cs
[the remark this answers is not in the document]
skel35: I give it a pass because the potential perf benefit seems relatively small here, and because we don't care about 100+ TEs a day cases.
I don't necessarily agree with the principle - null handling is quite trivial. And currently, such potential null errors are just swallowed by XAML - it's not very obvious edge cases handling either.

atsvirchkova:
Fixed the mentioned issues. I've also tried to move context menu opening assignment to Behavior. The problem is that this behavior would be attached to `TimelineTimeEntryBlock` with an attached property `ActiveTimeEntryBlockProperty` (and this property would be set on context menu open). This would require a "multiple to one" binding - we have many time entry blocks and just one `ActiveTimeEntryBlock` in the view model. I tried this out, and it didn't work properly.
Ready for review again.

src/ui/windows/TogglDesktop/TogglDesktop/ui/Resources/TimelineConstants.cs
```diff
@@ -14,6 +14,8 @@ public static class TimelineConstants
         public const double GapBetweenOverlappingTEs = 5;
         public const double AcceptableBlocksOverlap = 1e-5;
         public const double MinGapTimeEntryHeight = 10;
+        public const int DefaultTimeEntryLengthInSeconds= 3600;
```
skel35, line 17: `DefaultTimeEntryLengthInSeconds` should depend on `DefaultTimeEntryLengthInHours`

atsvirchkova: Fixed

skel35 (approved):
Looks good!

PR #4724, opened 2020-12-02
Fix ConvertTimeIntervalToHeight can return negative height

### LINKED ISSUE #4722: Running time entry can initially appear below the line of current time
### 💻 Environment
Platform: Windows
Toggl Version: 7.5.401

### 🐞 Actual behavior
Running time entry is displayed beneath the line representing the current time.

### 💯 Expected behavior
Running time entry should be very close to or touching the line representing the current time, but should not be beneath this line.

### 🔨 Steps to reproduce
1. Make sure you're zoomed in as much as possible.
2. Start a time entry

### 📦 Additional info
image

### LINKED ISSUE #4723: ConvertTimeIntervalToHeight can return negative height
### 💻 Environment
Platform: Windows
Toggl Version: 7.5.401

### 🐞 Actual behavior
`ConvertTimeIntervalToHeight` method can return negative height. Operating with negative heights is not something a developer would expect, so it can lead to unpredictable issues.

### 💯 Expected behavior
`ConvertTimeIntervalToHeight` method does not return negative heights. 

### 🔨 Steps to reproduce
1. Set a conditional breakpoint or an `if` statement + regular breakpoint inside or after the `ConvertTimeIntervalToHeight` method.
2. Start a running time entry.

atsvirchkova:
### 📒 Description
Removed dependency between current time line and running TE height. Now there is an observable timer with an interval of 1 sec; and both TE height and cur line update on that timer. Frequent current time line updates remove the problem with that it could be above the running entry.

### 🕶️ Types of changes

### 👫 Relationships

Closes #4722 
Closes #4723

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
@@ -56,7 +56,7 @@ namespace TogglDesktop.ViewModels
                 .ToPropertyEx(this, x => x.ActivityBlocks);
             var blocksObservable = Toggl.TimelineTimeEntries
                 .CombineLatest(Toggl.RunningTimeEntry, scaleModeObservable,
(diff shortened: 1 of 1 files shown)


skel35 (approved):
looks good

PR #4730, opened 2020-12-04
Context menu: overlap resolving

### LINKED ISSUE #4685: Context menu for TEs in Timeline view: change first entry stop time, change last entry start time
### 💻 Environment

Platform: Windows

### 📒 Description
Add menu items, which would do the following:

- "Change first entry stop time" (enabled only if a time entry is intersecting with another time entry)
- "Change last entry start time" (enabled only if a time entry is intersecting with another time entry)

### ⭐️ Why do you want this?
This is part of Timeline UI functionality #4670

atsvirchkova:
### 📒 Description
Add menu items, which would do the following:
- "Change first entry stop time" (enabled only if a time entry is intersecting with another time entry)
- "Change last entry start time" (enabled only if a time entry is intersecting with another time entry)

### 🕶️ Types of changes

### 👫 Relationships

Closes #4685 

### 🔎 Review hints
What to test:
image
If there are 2 overlaps for current TE (at the start and at the end), then the first one (sorting by start time) will be resolved.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
@@ -14,6 +14,7 @@ namespace TogglDesktop.ViewModels
     public class TimelineViewModel : ReactiveObject
     {
         private DateTime _lastDateLoaded;
+        private List<TimeEntryBlock> SortedTimeEntryBlocks { [ObservableAsProperty] get; }
 
         public TimelineViewModel()
         {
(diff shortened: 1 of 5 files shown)


src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
```diff
@@ -366,14 +366,14 @@ public static void ChangeFirstEntryStop(TimeEntryBlock item, List<TimeEntryBlock
         {
             var (first, last) = GetOverlappingPair(item, blocks);
             if (last == null) return;
-            Toggl.SetTimeEntryEndTimeStamp(first.TimeEntryId, (long)last.Started - 1);
```
skel35: @atsvirchkova this commit is "dont allow posibility of negative duration".
when can the negative duration be achieved? since Started is a UTC timestamp, it will be zero only on Jan 1st 1970, so I think we're safe here. Plus we're using -1 and +1 in the macOS app: https://github.com/toggl-open-source/toggldesktop/blob/505af8744c6dcd9736e13b7c4f9ca39529d64166/src/ui/osx/TogglDesktop/Features/Timeline/Models/TimelineData.swift#L203-L228

skel35: sorry, I misread "negative duration" for "negative timestamp".
perhaps we can just not show these items in case the duration is zero 🤔 ?

atsvirchkova:
Here the start time of two TEs are completely equal:
image
An if you try to resolve the first-layer one by clicking on "Change first entry stop time", it'll try to reduce the second-layer TE stop time to the start time of the first one (and its own start time also). And if we substract one from it, we'll be in a situation where end < start.

skel35:
good point, nice catch.
I'll update the macOS algorithm too, so that it handles this edge case as well 👍

skel35 (approved):
Code and behavior are looking good. Good job!

skel35:
(I'll merge it once the CI build is done)

skel35:
The CI error is a timeout when uploading artifacts - looks random, so I'll disregard that.

PR #4731, opened 2020-12-04
Unit tests for ConvertTimeEntriesToBlocks and GenerateGapTimeEntryBlocks 

### LINKED ISSUE #4696: Write unit tests for ConvertTimeEntriesToBlocks and GenerateGapTimeEntryBlocks
### 💻 Environment
Platform: Windows

### 📒 Description
Write unit tests for TimelineViewModel's ConvertTimeEntriesToBlocks and GenerateGapTimeEntryBlocks

atsvirchkova:
### 🕶️ Types of changes

 - **Unit tests**

### 👫 Relationships

 Closes #4696 

### 🔎 Review hints
Introduced `now` parameter in `ConvertTimeEntriesToBlocks` method to test that `runningEntry` is filtered correctly.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop.Tests/TimelineConvertTimeEntriesToBlocksTests.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop.Tests/TimelineConvertTimeEntriesToBlocksTests.cs
@@ -17,18 +17,19 @@ namespace TogglDesktop.Tests
         [MemberData(nameof(GetData))]
         public void TestConvertTimeEntriesToBlocks_CorrectOffsets(List<Toggl.TogglTimeEntryView> timeEntries,
             Toggl.TogglTimeEntryView? runningEntry,
-            int selectedScaleMode, List<(string Guid, double VerticalOffset, double Height, double HorizontalOffset)> expectedValues)
+            int selectedScaleMode, List<(string Guid, double VerticalOffset, double Height, double HorizontalOffset, bool IsOverlapping)> expectedValues)
         {
             var blocks = TimelineViewModel.ConvertTimeEntriesToBlocks(timeEntries,
                 runningEntry, selectedScaleMode, _selectedDate);
 
             Assert.Equal(expectedValues.Count, blocks.Count);
-            foreach (var (guid, verticalOffset, height, horizontalOffset) in expectedValues)
+            foreach (var (guid, verticalOffset, height, horizontalOffset, isOverlapping) in expectedValues)
             {
                 Assert.True(blocks.ContainsKey(guid));
                 Assert.Equal(verticalOffset, blocks[guid].VerticalOffset, 5);
                 Assert.Equal(height, blocks[guid].Height, 5);
                 Assert.Equal(horizontalOffset, blocks[guid].HorizontalOffset, 5);
+                Assert.Equal(isOverlapping, blocks[guid].IsOverlapping);
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/ui/windows/TogglDesktop/TogglDesktop.Tests/TimelineConvertTimeEntriesToBlocksTests.cs
```diff
@@ -0,0 +1,170 @@
...
+            var scaleMode = 1;
+            return new object[]
+            {
+                timeEntryList,
+                null,
+                scaleMode,
+                _selectedDate,
+                new List<(string Guid, double VerticalOffset, double Height, double HorizontalOffset, bool IsOverlapping)>()
+                {
+                    (timeEntryList[0].GUID, TimeToHeight(timeEntryList[0].Started - _selectedDateUnix, scaleMode),
```
skel35, line 107: This part of the tests is quite hard to read.
My suggestion to improve the readability:
1. Extract constants, like T1 = 0, T2 = 50, T3 = 100, T4 = 150, T5 = 200.
2. Use theses constants in your expected blocks list, e.g. `TimeToHeight(T1, scaleMode)` is easier to read than `TimeToHeight(timeEntryList[0].Started - _selectedDateUnix, scaleMode)`.

skel35 (changes requested):

atsvirchkova:
Is this situation possible?
- what if TE list has a TE with the same GUID as the running TE has
Currently in our code, TEs with the same GUID are considered equal, and such TEs will appear in the final list only once. What should be the expected behavior?

skel35:
> Is this situation possible?
> 
> * what if TE list has a TE with the same GUID as the running TE has
>   Currently in our code, TEs with the same GUID are considered equal, and such TEs will appear in the final list only once. What should be the expected behavior?

1. We don't guarantee that it's impossible.
2. Why do we need this line then? https://github.com/toggl-open-source/toggldesktop/blob/f97b9f9cede72e78a1d1a42886473edea858be58/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs#L223

atsvirchkova:
> > Is this situation possible?
> > 
> > * what if TE list has a TE with the same GUID as the running TE has
> >   Currently in our code, TEs with the same GUID are considered equal, and such TEs will appear in the final list only once. What should be the expected behavior?
> 
> 1. We don't guarantee that it's impossible.
> 2. Why do we need this line then? https://github.com/toggl-open-source/toggldesktop/blob/f97b9f9cede72e78a1d1a42886473edea858be58/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs#L223

I checked and yes, it may happen at the moment when the timer is being stopped. We receive a new list with already stopped TE, but `runningTE` is still not nulled. Currently we include the first met one, and omit the second. Not sure what the correct behavior should be.

atsvirchkova:
Fixed all the issues mentioned above. Also did a test for the situation with running TE have same guid as the one from the list. It reproduces the behavior of real application: `now` time is the same as `Ended` for this TE.

src/ui/windows/TogglDesktop/TogglDesktop.Tests/TimelineConvertTimeEntriesToBlocksTests.cs
```diff
@@ -0,0 +1,268 @@
...
+namespace TogglDesktop.Tests
+{
+    public class TimelineConvertTimeEntriesToBlocksTests
+    {
+        private static readonly DateTime _selectedDate = new DateTime(2020, 12, 3);
+        private static readonly ulong _selectedDateUnix = (ulong)Toggl.UnixFromDateTime(_selectedDate);
+        private const double _horizontalTimeEntryOffset =
+            TimelineConstants.GapBetweenOverlappingTEs + TimelineConstants.TimeEntryBlockWidth;
+
+        private static readonly ulong T1 = 0;
```
skel35, line 16: why not `const`? 😄

skel35 (approved):
Looks good

src/ui/windows/TogglDesktop/TogglDesktop.Tests/TimelineConvertTimeEntriesToBlocksTests.cs
atsvirchkova: Fixed

PR #4736, opened 2020-12-07
Fix press enter twice to start timer

### LINKED ISSUE #4158: Windows desktop - press enter twice to start timer
### 💻 Environment

Platform: Windows
OS Version: Windows 10
Toggl Version: 7.5.162

### 🐞 Actual behavior
From Diane "Looks like Windows Desktop users now have to press enter twice if they are starting a time entry using the keyboard. The first enter closes the drop down suggestion menu and the second starts the time entry. Mac seems to only have to be pressed once."

### 💯 Expected behavior
Pressing enter once should achieve the same outcome

### 🔨 Steps to reproduce
from user

1. For some reason since the last update the first enter hit often doesn't work and I need to hit enter twice. 
2. I'm so used to just typing and hitting enter though I frequently go back and find the timer not running. So frustrating

### 📦 Additional info
[Example case](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/conversation/27435393952)

atsvirchkova:
### 📒 Description
Fixes bug: when Autocompletion popup is open, user has to press Enter twice; first one to close the popup, and second one to actually start the timer.

### 🕶️ Types of changes

### 🤯 List of changes
Don't set PreviewKeyDown event to handled in AutoCompletionPopup.xaml.cs

### 👫 Relationships

 Closes #4158 

### 🔎 Review hints
Test that the following keys work as expected when focused and not on timer: key down, key up, esc, enter.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/controls/AutoCompletionPopup.xaml.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/controls/AutoCompletionPopup.xaml.cs
@@ -356,7 +356,7 @@ namespace TogglDesktop
         {
             var item = this.controller.SelectedItem;
             this.select(item, withKeyboard);
-            return true;
+            return item != null;
         }
 
         private void select(IAutoCompleteItem item, bool withKeyboard)


skel35 (changes requested):
Nice and neat change which corresponds to the issue spec, but I think such a fix is too radical. Many users are too used to being able to edit the time entry before it's being started.

I've checked the behavior of the old UI, and I believe we should stick to it.
Basically:
1. If no item from autocomplete is selected -> pressing Enter should start the time entry. This fixes the given bug.
2. If an item from autocomplete is selected -> pressing Enter should behave just like now, description/project/etc is filled, but the time entry is not started yet, and the user can edit the description/project before starting it.

atsvirchkova:
Fixed, now it starts only if nothing is selected in autocomplete.

skel35 (approved):
Tested various Autocomplete popups inside the app, works well 👍

PR #4740, opened 2020-12-08
Don't set stop time if it's not valid (lib)

### LINKED ISSUE #3193: "Stop time year must be between 2006 and 2030" Error
### 💻 Environment

Platform: macOS
OS Version: Current
Toggl Version: 7.4.484 

### 🐞 Actual behavior
When starting an entry by clicking on an older entry's play button, an error message appears at the bottom of the app screen stating "Stop time year must be between 2006 and 2030" 

User experiencing this; I cannot replicate

### 📦 Additional info
[Link](https://toggl.slack.com/archives/C029WCX0T/p1565194235087500) of Slack convo
[Link](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/2137598/conversations/23148809600) of Intercom convo

AndrewVebster:
### 📒 Description

Instead of setting `stop time` string right away, we should first check if this date is valid and > than `start time`.

This will probably fix the "Stop time year must be between 2006 and 2030" error.

From the logs, we can tell that this error is triggered by the Edit popover (because of the `DisplayTimeEntryEditor` event).
Platform code on Edit popover is very straightforward, it just passes text from the text field to the library. So I suppose the error is not there.

So the root cause of this issue is still unknown to me... From what I see this would happen only if when setting stop time string and `te->StartTime()` returns 0 (context.cc, line 3684). But it seems strange.

### 🕶️ Types of changes

### 👫 Relationships

Closes #3193

### 🔎 Review hints

Please tell me if this solution makes sense.

### DIFF
--- a/src/model/time_entry.cc
+++ b/src/model/time_entry.cc
@@ -223,10 +223,6 @@ void TimeEntry::SetDescription(const std::string &value, bool userModified) {
         SetDirty();
 }
 
-void TimeEntry::SetStopString(const std::string &value, bool userModified) {
-    SetStopTime(Formatter::Parse8601(value), userModified);
-}
-
 void TimeEntry::SetCreatedWith(const std::string &value) {
     if (CreatedWith.Set(value, false))
         SetDirty();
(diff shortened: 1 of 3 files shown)


skel35:
@AndrewVebster, I think this solution makes sense. We'd have to go deep to find the root cause - it looks like some sort of race condition, can't tell if it's on the platform or the library side. Somehow `SetStopUserInput()` is being called on the running time entry without a proper `SetTimeEntryStop()` call, while the user sets the stop time of the stopped time entry.

Your fix does not fully resolve the issue, but it seems like the fastest fix - the time entry editing would silently fail once but allow the user to edit the time entry without errors, so not a big deal.

src/model/time_entry.cc
```diff
@@ -223,10 +223,6 @@ void TimeEntry::SetDescription(const std::string &value, bool userModified) {
         SetDirty();
 }
 
-void TimeEntry::SetStopString(const std::string &value, bool userModified) {
```
skel35, line 226: you forgot to remove it from the `.h` file

skel35 (changes requested):
Fix the Linux build and `time_entry.h` and it's good to go

src/model/time_entry.cc
AndrewVebster: Done 👍

skel35 (approved):
✔️

PR #4746, opened 2020-12-10
Show Toggl Track menu item

### LINKED ISSUE #4174: Add "Open TogglDesktop" in the contextual menu of "mini timer".
### 💻 Environment
Platform: Windows 10 - 1909

### 📒 Description
In the previous versions, we could open TogglDesktop from the contextual menu of "mini timer".
But in the current version, this item is removed from contextual menu.

### ⭐️ Why do you want this?
It's more user-friendly, we can show quickly the main window of TogglDesktop from the "mini timer".

atsvirchkova:
### 📒 Description
Add "Show Toggl Track" menu item in mini Timer.

### 🕶️ Types of changes

### 👫 Relationships

Closes #4174 

### 🔎 Review hints
Test that it shows and activates main windows if it's minimized, hidden in tray or not active.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/Resources/DesignUpdate/ContextMenu.xaml
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/Resources/DesignUpdate/ContextMenu.xaml
@@ -41,7 +41,7 @@
         <Setter Property="Padding" Value="0 4" />
     </Style>
 
-    <Style TargetType="MenuItem" x:Key="ContextMenu.Item" BasedOn="{StaticResource MahApps.Styles.MenuItem}">
+    <Style TargetType="MenuItem" x:Key="Toggl.ContextMenu.Item" BasedOn="{StaticResource MahApps.Styles.MenuItem}">
         <Setter Property="Background" Value="{DynamicResource Toggl.CardBackground}" />
         <Setter Property="Height" Value="32" />
         <Style.Triggers>
@@ -51,9 +51,9 @@
         </Style.Triggers>
     </Style>
 
-    <Style TargetType="{x:Type MenuItem}" BasedOn="{StaticResource ContextMenu.Item}"/>
+    <Style TargetType="{x:Type MenuItem}" BasedOn="{StaticResource Toggl.ContextMenu.Item}"/>
 
-    <Style TargetType="{x:Type MenuItem}" x:Key="ContextMenu.Item.ShowIfEnabledStyle" BasedOn="{StaticResource ContextMenu.Item}">
+    <Style TargetType="{x:Type MenuItem}" x:Key="Toggl.ContextMenu.Item.ShowIfEnabledStyle" BasedOn="{StaticResource Toggl.ContextMenu.Item}">
         <Setter Property="Visibility" Value="Collapsed"/>
         <Style.Triggers>
             <Trigger Property="IsEnabled" Value="True">
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/Resources/DesignUpdate/ContextMenu.xaml
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/Resources/DesignUpdate/ContextMenu.xaml
@@ -41,12 +41,23 @@
         <Setter Property="Padding" Value="0 4" />
     </Style>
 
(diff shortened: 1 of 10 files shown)


src/ui/windows/TogglDesktop/TogglDesktop/ui/Resources/DesignUpdate/ContextMenu.xaml
```diff
@@ -51,6 +51,17 @@
         </Style.Triggers>
     </Style>
 
+    <Style TargetType="{x:Type MenuItem}" BasedOn="{StaticResource ContextMenu.Item}"/>
+
+    <Style TargetType="{x:Type MenuItem}" x:Key="ContextMenu.Item.ShowIfEnabledStyle" BasedOn="{StaticResource ContextMenu.Item}">
```
skel35, line 56: all app-level custom style keys here start with "Toggl."

src/ui/windows/TogglDesktop/TogglDesktop/ui/windows/MainWindow.xaml.cs
```diff
@@ -121,6 +126,9 @@ private void initializeContextMenu()
                     asMenuItem.CommandTarget = this;
                 }
             }
+
+            ShowMainWindowCommand = ReactiveCommand.Create(this.ShowOnTop, mainContextMenu.WhenAnyValue(x => x.PlacementTarget)
```
skel35, line 130: Discovered a bug: this item also appears on the tray icon menu, if the context menu was previously open on the MiniTimer.
Proposing an, admittedly hacky, fix: also check if contextMenu.Placement != Placement.AbsolutePoint, because this is what tray icon context menu is using (probably a hack in itself from `Hardcodet.Wpf.TaskbarNotification` library, but at least we can say it's not applicable for other appearances of this context menu).

skel35 (changes requested):

src/ui/windows/TogglDesktop/TogglDesktop/ui/windows/MainWindow.xaml.cs
atsvirchkova: In this case I'd rather vote for different xaml for mini timer. It shouldn't be that much work. I'll try how it'd look and work.

atsvirchkova:
I've created separate menus for main window and mini timer window. Tried to use the existing "Show" command in mini timer, but found a bug with it: open menu in main window, than right-click on mini timer to open context menu there. "Show Toggl Track" will be disabled there; that is because of `canExecute` condition on it. So I created separate command for it.
Ready for review.

src/ui/windows/TogglDesktop/TogglDesktop/ui/windows/MiniTimerWindow.xaml.cs
```diff
@@ -43,12 +42,9 @@ static void OnClosing(object sender, CancelEventArgs e)
 
         protected override void OnMouseRightButtonDown(MouseButtonEventArgs e)
```
skel35, line 43: I don't see a point in this method anymore. It tweaks the placement slightly, but
- it can be tweaked in XAML anyway
- I don't see a benefit in doing so, the default way works fine

I think this is just a leftover from some old logic and it can be safely removed.

skel35 (approved):
Looks good, I have just 1 small comment

src/ui/windows/TogglDesktop/TogglDesktop/ui/windows/MiniTimerWindow.xaml.cs
atsvirchkova: The default way would be different from the current: it would place the menu on mouse point. Is this OK for us?

skel35: I think it's totally fine. Do you see any downside with this?

atsvirchkova: I think it would be even better. There is a strange bug in placement now:
image
And if we go for mouse point placement, it will be fixed automatically.

atsvirchkova:
Removed `OnMouseRightButtonDown` method.

skel35:
Looks good 👍

PR #4769, opened 2020-12-17
Fix automatic scrolling to first entry

### LINKED ISSUE #4760: Scrolling to first time entry does not happen when switching back from current day
### 💻 Environment
Platform: Windows
Toggl Version: 7.5.418

### 🐞 Actual behavior
Automatic scrolling in the Timeline view does not happen.

### 💯 Expected behavior
Timeline View is automatically scrolled to the first time entry of the day

### 🔨 Steps to reproduce
1. Make sure you have time entries tracked for yesterday and no time entries tracked for today
2. Open Today
3. Open the day before.

atsvirchkova:
### 📒 Description
Fixes scrolling for two scenarios:
1. You have no entries tracked today, then you go to some day with some TEs, then go back to today and then back to that day. No scrolling happened previously.
2. You so to some day with no entries (not today). Create a TE. Scrolling happened here, despite it shouldn't.

### 🕶️ Types of changes

### 🤯 List of changes

- TimelineViewModel.FirstTimeEntryOffset was made a nullable double
- Timeline.xaml.cs condition for calling `SetMainViewScrollOffset`

### 👫 Relationships

Closes #4760 

### 🔎 Review hints
Test the scenarios described in Description section.

### DIFF
--- a/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
+++ b/src/ui/windows/TogglDesktop/TogglDesktop/ui/ViewModels/TimelineViewModel.cs
@@ -94,11 +94,10 @@ namespace TogglDesktop.ViewModels
                     GenerateRunningGapBlock(tuple.TimeEntries.Values, tuple.Running, CurrentTimeOffset, SelectedDate))
                 .ToPropertyEx(this, x => x.RunningGapTimeEntryBlock);
 
-            this.WhenAnyValue(x => x.TimeEntryBlocks)
+            FistTimeEntryOffsetForSelectedDate = this.WhenAnyValue(x => x.TimeEntryBlocks)
                 .Select(_ => SelectedDate).Buffer(2, 1)
                 .Where(b => b[1] != b[0]) // if there is new first TE offset for the same date, skip
-                .Select(blocks => TimeEntryBlocks.MinOrDefaultTimeEntryBlock())
-                .ToPropertyEx(this, x => x.FirstTimeEntryOffset);
+                .Select(blocks => TimeEntryBlocks.FirstTimeEntryVerticalOffset());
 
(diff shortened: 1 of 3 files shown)


skel35 (changes requested):
Works fine.
This logic is a bit tricky. I've left a couple small comments, but otherwise looks good.

atsvirchkova:
Moved transformations done on `FirstTimeEntryOffset` to the view model. Not sure if it's better there: conditions when we want and don't want to scroll to the first offset is more about view part, putting them into view model makes it not clear enough right there why we want `FirstTimeEntryOffset` to be calculated only sometimes.
Ready for review.

skel35:
>Moved transformations done on FirstTimeEntryOffset to the view model. Not sure if it's better there: conditions when we want and don't want to scroll to the first offset is more about view part, putting them into view model makes it not clear enough right there why we want FirstTimeEntryOffset to be calculated only sometimes.

Good point about FirstTimeEntryOffset property being more general than just scrolling position. This can be overcome by renaming it to something like ScrollToFirstTimeEntryOffset.

When I mentioned MinOrDefault() I meant a more universal method, sorry for not being clear.
This does look like a good method though, but the name doesn't fit well because someone reading the name won't understand what it means. I would actually call this method something like FirstTimeEntryOffset() 😄

atsvirchkova:
`ScrollToFirstTimeEntryOffset` sounds more like an action than a property 😃 I though about renaming it, but was unable to come up with something reasonable and left it as it was in the end. I think this would better look as an event with parameter `offset`, if we want it to be in view model.
Will rename `MinOrDefaultTimeEntryBlock` to `FirstTimeEntryOffset`. I'd also go for more universal method, but the first null check ruins the ability to transform dictionary to something more general.

atsvirchkova:
Fixed, ready for review again.

skel35 (approved):

PR #4776, opened 2020-12-22
Fix Windows build pipeline

### LINKED ISSUE #4774: GH Actions release fails for Windows targets
### 💻 Environment
Platform: Windows, GH Actions

### 🐞 Actual behavior
Starting from 7.5.430, releasing fails https://github.com/toggl-open-source/toggldesktop/actions/runs/437457740

### 💯 Expected behavior
Release script succeeds

### 🔨 Steps to reproduce
1. Merge something to master, see Actions

### 📦 Additional info
This issue may possibly have some info to help
https://github.com/actions/virtual-environments/issues/1090

atsvirchkova:
### 📒 Description
Add nuget clean step to Windows build pipeline.

### 🕶️ Types of changes

### 🤯 List of changes
main.yml

### 👫 Relationships

Closes #4774 

### 🔎 Review hints
Testing: check if both Windows builds are green.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -427,6 +427,9 @@ jobs:
       with:
         path: third_party/poco/lib
         key: ${{ runner.os }}-poco-lib-1.9.0
+    - name: Clean
+      shell: cmd
+      run: dotnet nuget locals all --clear
     - name: Build on Windows x86 without POCO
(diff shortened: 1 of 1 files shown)


skel35 (approved):
A bit weird to use `dotnet` command which is .NET Core even though we don't have any dependency on it in our solution. But this should still work as NuGet cache is global (per OS user) so looks good 👍

PR #4777, opened 2020-12-23
Clean git repo before proceeding with release script

skel35:
### 📒 Description
GH Actions workflow for releasing macOS app breaks with an error:
```
error: Your local changes to the following files would be overwritten by checkout:
2457
	src/ui/osx/Gemfile.lock
```

Cleaning the git repo before proceeding with the release script should be totally fine and it should clean up any local changes that have appeared before.

### 🕶️ Types of changes
- **Bug fix** (non-breaking change which fixes an issue)

### 🤯 List of changes
- Clean git repo before proceeding with release script

### 👫 Relationships
Closes #4767 (hopefully)

### 🔎 Review hints
Merge and see if it builds.
There should be other safer ways to test, but they are much slower.

PR #4778, opened 2020-12-23
Trying to fix macOS CI build again

AndrewVebster:
### 📒 Description

GH Actions workflow for releasing macOS app still breaks with an error:
```
error: Your local changes to the following files would be overwritten by checkout:
2457
	src/ui/osx/Gemfile.lock
```

Force checkout should remove this error.

### 👫 Relationships

Closes ##4767 
Related to PR #4777


skel35 (approved):
yeah, I forgot `git clean` doesn't remove staged files...

forcing the checkout should help

PR #4783, opened 2020-12-29
Analytics opt out setting

nikitalita:
### 📒 Description

Added a setting in preferences to opt out of analytics. Checking the option prevents analytics from being run. It does this by checking to see if the opt-out is set before sending an analytics event (similar to how it was checking if the environment was `"production"` before).

### 🕶️ Types of changes

Analytics opt-out setting in Preferences.

### 🤯 List of changes

Added Analytics opt-out setting in preferences.
### 👫 Relationships

Closes https://github.com/toggl-open-source/toggldesktop/issues/4782

### 🔎 Review hints

Make sure it doesn't break old databases; I think I've implemented the Migration feature properly, but I'm not certain.


nikitalita:
I had run `make fmt`, but it changed *a lot* of files as a result, including ones that I hadn't touched, so I didn't run it for this PR.

nikitalita:
Ok, I think that I've gotten this working properly on both windows and linux. I attempted to add this to the macOS version, but I am unfamiliar with how the Interface Builder works and all my attempts ended up screwing up the layout, so I will leave that to you.

I have added tests, but I am not sure how to run these? `make test` fails on both macOS and linux, and seems to also be hardcoded to a specific developers home directory.

skel35:
Thanks for the PR @nikitalita!
We'll try to review it and get back to you by the end of the week!

skel35:
Unfortunately didn't have time for it this week, we'll try to review it next week.

skel35:
@nikitalita
Unfortunately, we'd like to decline your pull request politely.
The app sends only _anonymized_ usage data to Google Analytics.  The data we send is about which buttons get clicked, which settings are turned on inside the app, or which OS version the user has. No personally identifiable information gets sent to Google servers.
As a side note, we have concrete plans to get rid of Google Analytics data collection in the future.
Sorry we couldn't merge it or notify you earlier about the impossibility of this, but thanks for the contribution anyway. Have a good day!

nikitalita:
> @nikitalita
> Unfortunately, we'd like to politely decline your pull request.
> The app sends only _anonymized_ usage data to Google Analytics. The data we send is about which buttons get clicked, which settings are turned on inside the app, or which OS version the user has. No personally identifiable information gets sent to Google servers.

With respect, this is not true. A user necessarily sends their IP when analytics information is sent. Combined with the fact that OS info and other personalized data is sent when first installed, this means that any data sent is easily correlated with a person. Correlating these things is exactly what Google is good at.

> As a side note, we have concrete plans to get rid of Google Analytics data collection in the future.

Great. I would still like to have the option to opt-out of analytics. I do leave analytics on for certain software products where I think the developers would be good stewards of my data, but I want to ability to opt out if that changes.

> Sorry we couldn't merge it or notify you earlier about the impossibility of this, but thanks for the contribution anyway. Have a good day!

I find it very concerning that you consider it "impossible" to not respect the privacy of your users, especially when the feature is trivial to implement. If this is your view, I don't feel comfortable with entrusting you with any data of mine.

PR #4789, opened 2021-01-07
Fix time format in Timeline view

### LINKED ISSUE #4785: Time format in Timeline view issue
### 💻 Environment

Platform: Windows
OS Version: Windows 10.1
Toggl Version: 7.5.418

### 🐞 Actual behavior
Your selected time format (profile settings page via the web app) is not reflected in the Timeline view, if you choose the 24-hour format the timeline view will remain in the AM/PM format. There's also an error for the stop time, as you can see in the following example, it shows an AM/PM start time and a 24h format stop time with PM added (15:17 PM): 

image

### 🔨 Steps to reproduce

1. Open the app
2. Go to Timeline view
3. Hover your mouse over a time entry and check the time format

[Intercom link](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/2097020/conversations/106481306148188)

atsvirchkova:
### 📒 Description
This PR utilizes time of day format from user profile settings. I introduced `toggl_time_of_day_format` function in c++ api, to make it accessible in c# code. Also added `time_of_day_format` field to settings structure. The format provided is not aligned with existing formats for `DateTime` in c#, so I introduced a function `ToDateTimeFormat` for format converting. This function is an analogue to c++ function `togglTimeOfDayToPocoFormat` in `formatter.cc`.

### 🕶️ Types of changes

### 👫 Relationships

Closes #4785 

### 🔎 Review hints
Testing: 
- Go to https://track.toggl.com (or https://track.toggl.space if tested on stage server) and see the time format.
- Run desktop app, open timeline view and check that time format is correct in TE tooltip.
- Change format on the web. Check that the format has changed in in the tooltip (and it's correct).

### DIFF
--- a/src/toggl_api.cc
+++ b/src/toggl_api.cc
@@ -1017,6 +1017,11 @@ bool_t toggl_timeline_is_recording_enabled(
     return app(context)->IsTimelineRecordingEnabled();
 }
 
+char_t* toggl_time_of_day_format(
+    void* context) {
+    return copy_string(app(context)->GetTimeOfDayFormat());
+}
+
 bool_t toggl_can_see_billable(
     void *context,
     const int64_t workspaceID) {
--- a/src/toggl_api.cc
+++ b/src/toggl_api.cc
@@ -1017,8 +1017,8 @@ bool_t toggl_timeline_is_recording_enabled(
     return app(context)->IsTimelineRecordingEnabled();
 }
 
-char_t* toggl_time_of_day_format(
-    void* context) {
+char_t *toggl_time_of_day_format(
+    void *context) {
     return copy_string(app(context)->GetTimeOfDayFormat());
 }
 
--- a/src/context.h
+++ b/src/context.h
@@ -469,6 +469,10 @@ class TOGGL_INTERNAL_EXPORT Context : public TimelineDatasource {
         return user_ && user_->RecordTimeline();
     }
 
+    std::string GetTimeOfDayFormat() const {
+        return user_ ? user_->TimeOfDayFormat() : "";
+    }
+
     error SetDefaultProject(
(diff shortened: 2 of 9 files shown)


skel35:
It shows the time in a 12h format when it should show it in a 24h format. The Edit view is showing the format correctly.
image
image

For example, 01:22 (AM) and 13:22 is displayed as the same, we should not be the case neither in 12h nor in 24h format.
image

atsvirchkova:
Thank you for testing, now the problem is fixed.

src/toggl_api.cc
```diff
@@ -1017,6 +1017,11 @@ bool_t toggl_timeline_is_recording_enabled(
     return app(context)->IsTimelineRecordingEnabled();
 }
 
+char_t* toggl_time_of_day_format(
+    void* context) {
```
skel35, line 1021: The `*` should be to the left of the member name as everywhere, not to the right of the type name.
One issue is that our API generation will fail if the formatting is incorrect.

`make format` will fix this but it will probably change too many things so this better be fixed manually.

skel35 (changes requested):
Please also use the variable format on the hours' scale on the left side, as it's now easy to do so.

atsvirchkova:
Fixed the time format on the hours scale in Timeline view. One thing: hours list is currently updated only when scale is changed; it's not updated when hour format setting is changed in web. For this to work it would be needed to create a callback in the api to notify about changed hour format, and the benefit is not worth the complication. For now, hours on the left side will be updated after user restarts the app or changes scale.

skel35 (approved):
Looks good

skel35:
one moment, found another minor formatting issue, I'll just push the fix right away

PR #4876, opened 2021-12-06
Reset validation error when setting TE duration - that will trigger TE to be synced when stopped

### LINKED ISSUE #4868: Issues due to Required Fields
### 💻 Environment

Have reproduced the issue on Mac (7.5.441), customer is using latest Win. 
Platform: 
OS Version: 
Toggl Version: 

### 🐞 Actual behavior

If a team is using the Required Fields feature, desktop app is not handling the entries in a user friendly way. Current state:
If a user saves an entry that doesn't have all required fields selected; any following TEs that do have all required elements are not saved either. This can result in extreme data losses. Plus it is not easy for the customer to fix this.
In addition the error message that is displayed - "unmet constraint" - doesn't tell the user what exactly is wrong and how they should act to save their data.

### 💯 Expected behavior

The TE with unmet constrains should be clearly marked and the error should clearly state what the user should do in order to sync this particular TE. All other TEs that are added after the first entry with unmet constrains, should not be affected!! Currently we are not syncing these TEs hence user is running the risk of losing data. Ideally the entry which doesn't meet required fields criteria shouldn't be saved at the first place - like the webapp is currently handling this. 

### 🔨 Steps to reproduce

1. 
2. 
3. 

### 📦 Additional info

Discussion [here](https://toggl.slack.com/archives/C029WCX0T/p1632221260022800)

atsvirchkova:
### 📒 Description

This PR fixes the following problem:

1. Start a TE, that violates some constraints.
2. Stop this TE.
3. Create a TE, that meets all necessary constraints. See the sync error displayed (because of the previous violating TE).
4. Delete the TE, that violates the constraints.

AR: The TE meeting all constraints won't be ever synced.

ER: TE that meets all the constraints will be synced when stopped.

### 🕶️ Types of changes

### 🤯 List of changes

This PR simply clears validation error in `StopTracking()` (this method is called when TE is stopped).
Note that the same thing is done in `SetStopUserInput` - I'm not sure that this method is used anywhere (there is method in `context` that calls it, but that method is most probably never called through `TogglApi`).

### 👫 Relationships

 Closes #4868

### 🔎 Review hints

Try to reproduce the bug:

1. Start a TE, that violates some constraints.
2. Stop this TE.
3. Start another TE and fill it in to meet all necessary constraints.
4. Stop this TE.

There should be no sync error for the second TE (it should be synced after stopping). The error for the first TE should be preserved until user fixes the error.

### DIFF
--- a/src/model/time_entry.cc
+++ b/src/model/time_entry.cc
@@ -144,7 +144,10 @@ void TimeEntry::DiscardAt(const Poco::Int64 at) {
 
     SetDurationInSeconds(duration, true);
     SetStopTime(at, true);
-    SetUIModified();
+    if (Dirty()) {
+        ClearValidationError();
+        SetUIModified();
+    }
 }
 
 void TimeEntry::StopTracking() {


AndrewVebster (approved):
LGTM 👍

